PR #1, opened 2022-06-21
Initial dqlite cluster implementation

masnax:
This is the internal cluster implementation for `microcluster`, following the overall structure of `lxd-cloud`. 

The daemon is started with a listen address and a state directory, holding two listeners:
* A unix socket for local control
* HTTPS socket for the main REST API

The daemon can then be bootstrapped as a new node or join an existing cluster with a join token.

Join tokens are issued over the `/secrets` endpoint, with the server certificate of the joining node. A token key is generated using LXD's `shared.RandomCryptoString`. This key along with the address and cluster certificate of the existing cluster member is base64 encoded and returned as a string. 

Initializing a daemon with this join token will provide the cluster public and private keys, along with the local list of cluster members to the joining node.


internal/logger/logger.go
```diff
@@ -0,0 +1,116 @@
+package logger
```
stgraber, line 1: Any reason we can't just import `github.com/lxc/lxd/shared/logging` for this?

internal/db/cluster/schema.go
```diff
@@ -0,0 +1,61 @@
+package cluster
+
+import "time"
+
+// CreateSchema is the default schema applied when bootstrapping the database.
+const CreateSchema = `
```
stgraber, line 6: Don't we want the schema to be assembled through the patches similar to what we're doing in LXD?
Or is that one of the subsequent PRs?

Also, I thought we said we'd prefix the internal table names so that we reduce the likelihood of name conflicts with the consumer?

internal/db/cluster/schema.go
```diff
@@ -0,0 +1,61 @@
...
+CREATE TABLE certificates (
+  id           INTEGER        PRIMARY  KEY    AUTOINCREMENT  NOT  NULL,
+  fingerprint  TEXT           NOT      NULL,
+  type         INTEGER        NOT      NULL,
+  name         TEXT           NOT      NULL,
+  certificate  text           NOT      NULL,
+  UNIQUE       (fingerprint)
+);
+
+CREATE TABLE secrets (
```
stgraber, line 23: Maybe call this `tokens` instead of `secrets`?

To me `tokens` assume something ephemeral whereas `secrets` are more permanent.

stgraber:
@masnax I wonder if we can't also somewhat namespace the internal API endpoints in the same way we'll be doing the DB.
Basically move everything under `/1.0/cluster/XYZ` leaving the rest of `/1.0` free for the consumer.

This would make documentation easier by saying that the `cluster` namespace is owned by `microcluster` and cannot be used in DB table names or API paths, the rest is fair game.

stgraber:
Or maybe for the API we should actually use `/cluster/1.0/XYZ`, leaving the rest open for the consumer, so they can implement a `/1.0` like we do in LXD or can implement `/v1` or something if they prefer, they'd just need to stay away from `/cluster`.

masnax:
> @masnax I wonder if we can't also somewhat namespace the internal API endpoints in the same way we'll be doing the DB. Basically move everything under `/1.0/cluster/XYZ` leaving the rest of `/1.0` free for the consumer.
> 
> This would make documentation easier by saying that the `cluster` namespace is owned by `microcluster` and cannot be used in DB table names or API paths, the rest is fair game.

What I've done instead is have a separate `/internal/XYZ` and `/public/XYZ` for internally included and externally added APIs, respectively. I can just rename `/public` to `/1.0` if you would prefer :)

masnax:
@stgraber Since I see there's a bit of confusion about what's implemented in which PR, do you think it would help if I just throw the whole thing together into 1 big PR?

internal/db/cluster/schema.go
masnax: I was hesitant to call it `tokens` because it's not actually the tokens returned from the API that is being stored in the database, rather just a unique record of who we should issue a token for. I suppose I could make it `token_records`?

stgraber:
> @stgraber Since I see there's a bit of confusion about what's implemented in which PR, do you think it would help if I just throw the whole thing together into 1 big PR?

Nah, that PR made it easier to review.

> What I've done instead is have a separate `/internal/XYZ` and `/public/XYZ` for internally included and externally added APIs, respectively. I can just rename `/public` to `/1.0` if you would prefer :)

So I think we have 3 categories of API:
 - /cluster/1.0/XYZ => "public" stable APIs to manage microcluster
 - /cluster/internal/XYZ => internal API for microcluster
 - /XYZ => whatever the consumer wants,

masnax:
I see, so then my implementation currently lumps `/cluster/1.0/XYZ` and `/cluster/internal/XYZ` together. Other than heartbeats and dqlite, is there anything else we want on `/cluster/internal`?

stgraber:
> @stgraber I see, so then my implementation currently lumps `/cluster/1.0/XYZ` and `/cluster/internal/XYZ` together. Other than heartbeats and dqlite, is there anything else we want on `/cluster/internal`?

Nope, I think that's the two we usually have on /internal, well, the SQL endpoint is also usually on there but may be part of your dqlite endpoint.

masnax:
@stgraber PRs should be good to go now, I've added an `internal_` prefix to all databases, and split up the endpoints as such:

`/1.0/XYZ` is available to the consumer for any custom APIs. By default there's nothing on it.
`/cluster/1.0` is for default public APIs.
`/cluster/control` is for default unix socket APIs
`/cluster/internal` is for restricted APIs

PR #4, opened 2022-07-11
External packages and example CLI

masnax:
Depends on #3 

This mainly refactors the project to expose some components so that they can be externally called from other packages. 
Basically, this moves stuff out of the `internal` package and adds top-level packages for:
* `client` - Adds a `/1.0` endpoint helper and helpers for querying the entire cluster at once. Creating a client is still internal, except through a few helpers in the `microcluster` package.
* `cluster` - The generated components of the database need to be external because external calls to the generator will rely on them.
* `rest` - Exposes types for API endpoint creation. The actual handling is still internal.
* `state` - This one's a bit hacky, basically we need the external APIs to have access to the daemon's state to be able to actually do things, so this just exposes a type definition.

Additionally there is now a `microcluster` package with an `App` command that takes some state args and interfaces with the `internal` package to create some useful helpers for interacting with the daemon externally. This will allow passing in our own set of `[]rest.Endpoint` and `map[int]schema.Update` that will be appended to the `/1.0` endpoints and the default schema, respectively.

Lastly there is an `example` directory with a `microd/microctl` project inside. This uses `microcluster.App` along with the  packages above to create a basic cluster. [Here](https://github.com/canonical/microcluster/blob/ab8ada7fec4d269d1953a9ffff3ca49bd99feee8/example/README.md) is the readme.


cluster/cluster_members.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		return err
+	}
+
+	n, err := result.RowsAffected()
+	if err != nil {
+		return err
+	}
+	if n != 1 {
+		return fmt.Errorf("updated %d rows instead of 1", n)
```
stgraber, line 94: Nit, should be `Updated`

example/Makefile
```diff
@@ -0,0 +1,51 @@
+SPHINXENV=doc/.sphinx/venv/bin/activate
```
stgraber, line 1: Not needed

PR #5, opened 2022-07-13
Various fixes.

masnax:
Depends on #4 

This adds several bug fixes that I noticed upon more nuanced testing:

* Most importantly, this moves initiating the heartbeat out of the dqlite dialFunc. Calls to the dqlite client methods (to check the leader and cluster status during a heartbeat) will initiate further calls to the dialFunc, which calls more heartbeats, and on and on. I noticed my CPU spiking up to 600% from getting slammed with heartbeat requests. The corresponding commit here adds a helper that calls the heartbeat function every second, and is now set up to run in a goroutine upon opening the database.

* Auto-schema-update logic was missing from the heartbeat sequence. Now there is an environment variable: `SCHEMA_UPDATE` that will be checked for an auto-update executable.
* Adds a 5s context timeout for calls to the dqlite client since they can be blocking in some cases. I'm unable to replicate this frequently enough (It seems to happen randomly as far as I can tell) to be able to test how exactly this works, but I noticed that very occasionally, attempting to check if we are the dqlite leader froze the daemon. If this happens on the leader then no heartbeats will be sent until the leader is reloaded.
* Ensures all API responses are fully read and closed so we can reuse the connection. 
* Prevents erroring out during reload if other cluster members are still updating
* The heartbeat timeout logic was incorrect, and is now correctly sleeping a maximum of half the client context timeout for heartbeats (`HeartbeatTimeout`), and a minimum of 2 seconds, as the other cluster members will report their heartbeat times a bit ahead of the leader).
* Adds some debug logs for the heartbeat.

### DIFF
--- a/example/cmd/microctl/main.go
+++ b/example/cmd/microctl/main.go
@@ -46,11 +46,8 @@ func main() {
 	var cmdPeers = cmdClusterMembers{common: &commonCmd}
 	app.AddCommand(cmdPeers.Command())
 
-	//var cmdReload = cmdReload{common: &commonCmd}
-	//app.AddCommand(cmdReload.Command())
-
-	//var cmdShutdown = cmdShutdown{common: &commonCmd}
-	//app.AddCommand(cmdShutdown.Command())
+	var cmdShutdown = cmdShutdown{common: &commonCmd}
+	app.AddCommand(cmdShutdown.Command())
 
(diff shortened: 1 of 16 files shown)


masnax:
This is a bit longer than 8 commits now as I've added the shutdown command as well.

PR #6, opened 2022-08-12
Use "name" for join token creation

masnax:
This replaces using the joining cluster member's server certificate with a name string.

The side effect of this is that the name string now needs to be passed alongside the token when joining a cluster.
```bash
microctl init <name> --token <token>
```
In terms of the CLI, requesting a join token has been changed from 
```bash
microctl secrets <server.crt>
```
to 
```bash
microctl tokens add <name>
```

There is now also a `microctl tokens list` for listing tokens, and `microctl tokens revoke <name>` for deleting a token.

I didn't actually bump the schema version, so if that's something we want to start doing, I'll make that change as well.


stgraber:
@masnax why do we need to specify the name at join time?

Surely, the secret provided in the token will only match up with a single token in the DB and we therefore already know what the name needs to be.

masnax:
> @masnax why do we need to specify the name at join time?
> 
> Surely, the secret provided in the token will only match up with a single token in the DB and we therefore already know what the name needs to be.

This was just a side-effect of how I'm implementing the API. I would think doing a POST on `/tokens/<token_string>` might be troublesome with the content of the string, so I wanted to do `/tokens/name` instead. But I've simplified it now to just include the server name in the token.

stgraber:
Hmm, I guess I would have expected the token creation to be done through `POST /cluster/1.0/tokens`, this would effectively just take in a name for the token. Then on the consumer side, they'd hit `POST /cluster/1.0/cluster` to join the cluster, providing the token as part of that.

This means the token itself doesn't need to contain the name nor does the joiner need to know what name it's been assigned.

stgraber:
@masnax am I missing something?

masnax:
@stgraber To hit `POST /cluster/1.0/cluster` you would need to know the address of who you're posting to, which is just swapping `name` for an `address` identifier, no?

masnax:
@stgraber Alright, so I've got it working like so now:

* `JoinToken` is now a field on the `POST` struct for adding a cluster member.
* Rather than handling the token on its own API, the token is now handled when we call `AddClusterMember`, and cluster info is returned. So `POST /tokens/NAME` is gone.
* The token now stores the addresses of all existing cluster members. The joiner will iterate on these addresses and complete after the first successful run of `AddClusterMember`, or error out if none work.

masnax:
Just fixed the order of handling the token and creating a new cluster member. Previously it would create the cluster member and then check the validity of the token, now it's the other way around.

PR #9, opened 2022-08-19
Remove a cluster member

masnax:
Adds the ability to remove a cluster member via `DELETE` to `/cluster/NAME` on both the control and public endpoints.

When a cluster member is removed, its state directory is wiped via a `PUT` to`/cluster/NAME` that is initiated by the `DELETE`. Its database is stopped and its daemon is then reset, so it can be re-initialized with an `init --bootstrap` or `init --token`. 

The CLI example has also been slightly edited:
`microctl cluster list` now prints the cluster information.
`microctl cluster remove` removes a cluster member.

### DIFF
--- a/internal/rest/resources/cluster.go
+++ b/internal/rest/resources/cluster.go
@@ -6,6 +6,9 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"os"
+	"strings"
+	"sync"
 	"time"
 
 	dqliteClient "github.com/canonical/go-dqlite/client"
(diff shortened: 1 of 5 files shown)


internal/rest/resources/cluster.go
```diff
@@ -192,3 +206,251 @@ func clusterGet(state *state.State, r *http.Request) response.Response {
...
+		clusterDisableMu.Lock()
+		defer clusterDisableMu.Unlock()
+		execPath, err := os.Readlink("/proc/self/exe")
+		if err != nil {
+			execPath = "bad-exec-path"
+		}
+
+		// The execPath from /proc/self/exe can end with " (deleted)" if the lxd binary has been removed/changed
+		// since the lxd process was started, strip this so that we only return a valid path.
+			logger.Info("Restarting daemon following removal from cluster")
```
tomponline, line 241: Strange tabbing here.

tomponline: Do you run your code through `go fmt` on save?

masnax: Vim plugins were probably acting up. Sorry about that.

PR #18, opened 2022-09-06
Fix internal_cluster_members primary key for lxd-generate and update invocations.

masnax:
This correctly applies the `primary=yes` lxd-generate tag to the `InternalClusterMembers` struct on the `Name` field, which corresponds to the unique `name` column on the `internal_cluster_members` table.

Additionally, this updates the `--database` (or `-d` shorthand) flag to expect the full import path for the cluster package, rather than just the package name and hoping `goimports` finds the right one.

### DIFF
--- a/cluster/cluster_members.go
+++ b/cluster/cluster_members.go
@@ -38,8 +38,8 @@ const Pending Role = "PENDING"
 // InternalClusterMember represents the global database entry for a dqlite cluster member.
 type InternalClusterMember struct {
 	ID          int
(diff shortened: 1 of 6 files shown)


masnax:
@stgraber Once this and https://github.com/lxc/lxd/pull/10881 are merged, you should update the `lxd-generate` invocations in `microceph` to also use `-d github.com/canonical/microcluster/cluster` instead of `-d cluster`.

PR #20, opened 2022-09-20
Supply daemon address/name on bootstrap/join

masnax:
Previously, the daemon would require a listen address to be supplied in order to start, and the value would not be stored anywhere. A restart would require specifying this address again.

Additionally, the cluster member name for each daemon was generated by its state directory name (for testing, mainly). This was problematic for scripting as each daemon would need a unique state directory name.

To solve both of these problems, this updates the MicroCluster daemon to accept a `name` and `address` argument upon bootstrapping or joining a cluster with a token. These values will then be saved in a `daemon.yaml` file in the state directory. 

* Updates go mod and lxd-generate code to the most recent updates.
* Supplies a context to helpers from LXD that now require it.
* As we want to update the `name` and `address` of the daemon, this adds a subtype to `Remote` called `Location` (similar to `LXD Cloud`) that holds location information about the node we are dealing with. This does not include `Certificate` so that we don't face issues unmarshalling an empty value.
* Adds a `setDaemonConfig` helper, on the daemon which will either insert a new `daemon.yaml` if a `Location` is supplied, or read the current config from the existing yaml.

### DIFF
--- a/example/README.md
+++ b/example/README.md
@@ -12,11 +12,11 @@ The daemon can be started with `microd` and is controlled by `microctl`.
 `make`
 
 ## Running
-This starts up three daemons at the given addresses.
+This starts up three daemons.
 ```bash
-microd --state-dir /path/to/state/dir1 --address 127.0.0.1:9001 &
-microd --state-dir /path/to/state/dir2 --address 127.0.0.1:9002 &
-microd --state-dir /path/to/state/dir3 --address 127.0.0.1:9003 &
+microd --state-dir /path/to/state/dir1 &
+microd --state-dir /path/to/state/dir2 &
+microd --state-dir /path/to/state/dir3 &
 ```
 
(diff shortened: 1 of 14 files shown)


example/micro2
```diff
@@ -0,0 +1,43 @@
+#!/bin/zsh
```
stgraber, line 1: What's this script?

example/test.go
```diff
@@ -0,0 +1,45 @@
+package main
```
stgraber, line 1: And that file?

masnax: Oh wow, oops... Must've done a `git add .` at some point without realizing it.

example/micro2
masnax: Just a script for testing, must've done a `git add .` or `git add <directory>` by accident. Both files should be gone now.

PR #24, opened 2022-09-22
Hooks interface

masnax:
Adds a `Hooks` interface under a `config` package. There are 5 hooks currently defined:
* `OnBootstrapHook` which runs after a bootstrap
* `OnStartHook` which runs after the daemon starts, and before `waitready` returns
* `OnJoinHook` which runs after joining an existing cluster
* `OnRemoveHook` which runs on the leader after a cluster member is removed
* `OnHeartbeatHook` which runs on the leader after a successful heartbeat

These can be implemented and the parent struct can be supplied to `microcluster.Start`.

The `State` has two additional functions for `OnRemoveHook` and `OnHeartbeatHook` as these are actions that take place over the API. Alternatively I can stick them in the endpoint handler and have them run after the specific endpoints are hit. This would avoid having to add the functions to the `State`, but I feel that would be hiding them a bit too much.

### DIFF
--- a/config/hooks.go
+++ b/config/hooks.go
@@ -1,0 +1,24 @@
+package config
+
+import (
+	"github.com/canonical/microcluster/internal/state"
+)
+
+// Hooks is an interface representing customizable functions that can be called at varying points by the daemon to
+// integrate with other tools.
+type Hooks interface {
+	// OnBootstrapHook is run after the daemon is initialized and bootstrapped.
+	OnBootstrapHook(s *state.State) error
+
+	// OnStartHook is run after the daemon is started.
+	OnStartHook(s *state.State) error
+
+	// OnJoinHook is run after the daemon is initialized and joins a cluster.
+	OnJoinHook(s *state.State) error
+
+	// OnRemoveHook is run after the daemon is removed from a cluster.
+	OnRemoveHook(s *state.State) error
+
+	// OnHeartbeatHook is run after a successful heartbeat round.
+	OnHeartbeatHook(s *state.State) error
+}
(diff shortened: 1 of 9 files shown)


stgraber:
@masnax for the functions on the state struct, do they need to be public?

stgraber:
It doesn't look like we intend for a user to directly set them or access them, so just making them a private property of the state struct should do the trick?

masnax:
@stgraber They need to be public because they're called in the API handlers. I've removed them from the `state` struct and instead just made them function variables defined in the internal `state` package. They're assigned to the daemon's equivalent functions when we get a new state. That should prevent any external API handlers from being able to call them, as they will use the public `state` package instead of the internal one.

tomponline:
@masnax why did you use an interface rather than a struct with individual hook fields as func types? This way one could just override selective hooks without needing to implement every one (by just populating the fields needed). An interface of funcs looks a little odd.

masnax:
> @masnax why did you use an interface rather than a struct with individual hook fields as func types? This way one could just override selective hooks without needing to implement every one (by just populating the fields needed). An interface of funcs looks a little odd.

My thought process was that there’s probably some sort of common information that will be changed/notified of changes in these hooks, so an interface would allow for an implementation that would share that information as other struct fields and methods. It definitely looks weird when the implementation is an empty struct though.

tomponline:
Yeah I meant just a struct with fields that are func type for each hook, then you can pass an empty struct (defaults the fields to nil) and then Start can check for any non nil entries and wire them up.

PR #29, opened 2022-09-28
Hooks

masnax:


config/hooks.go
```diff
@@ -4,21 +4,21 @@ import (
...
 )
 
-// Hooks is an interface representing customizable functions that can be called at varying points by the daemon to
+// Hooks holds customizable functions that can be called at varying points by the daemon to.
 // integrate with other tools.
-type Hooks interface {
+type Hooks struct {
 	// OnBootstrapHook is run after the daemon is initialized and bootstrapped.
-	OnBootstrapHook(s *state.State) error
+	OnBootstrapHook func(s *state.State) error
```
tomponline, line 11: The "Hook" at the end of each field name seems redundant seeing as they belong to the Hooks struct.

internal/daemon/daemon.go
```diff
@@ -154,6 +149,37 @@ func (d *Daemon) init(extendedEndpoints []rest.Endpoint, schemaExtensions map[in
...
+	if d.hooks.OnStartHook == nil {
+		d.hooks.OnStartHook = defaultHook
+	}
+
+	if d.hooks.OnHeartbeatHook == nil {
+		d.hooks.OnHeartbeatHook = defaultHook
+	}
+
+	if d.hooks.OnRemoveHook == nil {
+		d.hooks.OnRemoveHook = defaultHook
```
tomponline, line 179: What does the defaultHook do? Is it required ?

masnax: It's just an empty function, so we don't have to constantly check if a hook is nil before trying to run it.

tomponline: Ah ok maybe call it noopHook

PR #35, opened 2022-10-06
Forward request to target if `?target=name` is specified

### LINKED ISSUE #31: Implement `UseTarget` in client and add easy forwarding logic on server side
With MicroCeph, I find myself needing a way to easily throw a particular request at a specific server.
I could have the client tool handle this by hitting the cluster API to fetch the list of servers and then connect to the specific remote, but that's a bit annoying to do.

Instead, I think we should facilitate the same logic we have in LXD. Add a `UseTarget(name)` function to the client which returns an alternative version of the client that marks all HTTP requests with `?target=name`. On the server side, we should then have an option on the endpoints to proxy to the requested target.

So something like `ProxyTarget: true` would then have the endpoint handler on the server, look for a `target` URL attribute and if found, will connect to the remote and send the request its way rather than processing it locally.

masnax:
Closes #31 

Adds a `ProxyTarget` flag to each request method handler. If this is true, and the request URL has `?target=name` set, we will attempt to send the request over to the cluster member with the given `name`. 

A client method `UseTarget` has been added, which will return a new `client` with the `?target=name` query set.

### DIFF
--- a/internal/rest/rest.go
+++ b/internal/rest/rest.go
@@ -3,17 +3,22 @@ package rest
 import (
 	"context"
 	"crypto/x509"
+	"database/sql"
 	"fmt"
 	"net/http"
+	"net/url"
 	"path/filepath"
 
 	"github.com/gorilla/mux"
 	"github.com/lxc/lxd/lxd/request"
 	"github.com/lxc/lxd/lxd/response"
 	"github.com/lxc/lxd/lxd/util"
+	"github.com/lxc/lxd/shared/api"
 	"github.com/lxc/lxd/shared/logger"
 
+	"github.com/canonical/microcluster/cluster"
 	"github.com/canonical/microcluster/internal/rest/access"
+	"github.com/canonical/microcluster/internal/rest/client"
 	internalState "github.com/canonical/microcluster/internal/state"
 	"github.com/canonical/microcluster/rest"
 )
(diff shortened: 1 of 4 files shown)


internal/rest/rest.go
```diff
@@ -52,9 +57,76 @@ func handleAPIRequest(action rest.EndpointAction, state *internalState.State, w
...
+		return response.Unavailable(fmt.Errorf("Cannot send request to target. Daemon not yet initialized"))
+	}
+
+	if r.URL == nil {
+		return action.Handler(s, r)
+	}
+
+	values, err := url.ParseQuery(r.URL.RawQuery)
+	if err != nil {
+		logger.Warnf("Failed to parse query string %q: %v", r.URL.RawQuery, err)
```
masnax, line 78: I noticed in LXD, we don't error out if we fail to parse a query, instead just returning a `Warn` log and ignoring the query. I kept the same behaviour, but I'm not sure why this is the case. Thoughts @stgraber?

stgraber: Not exactly sure why we have that in LXD.

tomponline: In LXD that logic is inside the `queryParam(request *http.Request, key string) string` helper which doesn't return an error for convenience (of not having to check it). The query param is either there or not, and if the query string is invalid, its not there effectively.

In this case I think its probably more correct to return an error if an invalid query string is passed.

PR #55, opened 2023-03-24
Wrap MicroCluster in an interface

kyzrfranz:
Hey guys, when I wrote those tests for microceph I realised it might be a good idea to have this here, so that we can mock things better.
I'd be happy to add some more tests & align the structure throughout the micro-tools for lxd if you guys feel like that's a good idea.


stgraber:
Commits must be signed off before we can review them.

masnax:
The interface here should also have a function to grab the filesystem so that projects using microcluster can directly interact with the state directory. 

There's a `FileSystem` field on the internal `microCluster` struct, and this should be returned by a function that's included in the interface.

PR #56, opened 2023-03-30
Add facility to configure socket's group ownership

gboutry:
Hi, I need to configure the socket's group and thought it was best implemented here.

### DIFF
--- a/example/cmd/microd/main.go
+++ b/example/cmd/microd/main.go
@@ -44,7 +44,8 @@ func (c *cmdGlobal) Run(cmd *cobra.Command, args []string) error {
 type cmdDaemon struct {
 	global *cmdGlobal
 
-	flagStateDir string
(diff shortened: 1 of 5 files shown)


stgraber:
@gboutry looks good to me, I'll happily merge it, though I'm wondering what your use case is for this.
What project is that for and where is it that we want to open up access to the unix socket?

I'm mostly wondering from a security standpoint as all downstream projects (MicroCloud, MicroCeph and MicroOVN) have a lot of very scary things you can do through their API should a non-root user be allowed to access it.

gboutry:
@stgraber my usecase is for snap-openstack, to be able perform administration commands on the cluster (it's using microcluster as its state backend)

stgraber:
@gboutry okay, cool, great to hear we have another downstream project!

PR #62, opened 2023-05-24
Fix various truststore races

masnax:
This fixes various race conditions which resulted in the truststore being temporarily wiped. This caused a lot of weird issues in `microcloud` and the other `micro` services. Some of these are recorded [here](https://github.com/canonical/microcloud/issues).

The changes here effectively ensure that
* We won't send a heartbeat based on an empty cluster record.
* Attempting to clear the local remotes will fail with an error, so the behaviour won't be non-deterministic.


internal/rest/resources/tokens.go
```diff
@@ -58,6 +59,14 @@ func tokensPost(state *state.State, r *http.Request) response.Response {
 		joinAddresses = append(joinAddresses, addr)
 	}
 
+	if len(joinAddresses) == 0 {
```
tomponline, line 62: Thanks for this fix.

Do we need to have this "workaround" though? Can we not wait for the  trust store to become ready?

internal/trust/remotes.go
```diff
@@ -68,9 +68,15 @@ func (r *Remotes) Load(dir string) error {
...
 
-		r.data[remote.Name] = *remote
+		remoteData[remote.Name] = *remote
 	}
 
+	if len(remoteData) == 0 {
+		return fmt.Errorf("Failed to parse new remotes from truststore")
+	}
+
+	r.data = remoteData
```
tomponline, line 78: Do we need a mutex around this?

internal/trust/remotes.go
```diff
@@ -260,11 +260,17 @@ func (r *Remotes) Count() int {
 	return len(r.data)
 }
 
+// RemotesByName returns a copy of the list of peers, keyed by each system's name.
 func (r *Remotes) RemotesByName() map[string]Remote {
 	r.updateMu.RLock()
 	defer r.updateMu.RUnlock()
 
-	return r.data
+	remoteData := make(map[string]Remote, len(r.data))
```
tomponline, line 268: Feels like we should have a mutex protecting access to r.data

masnax: There's a mutex in all of these, it's just deferred from the top of the function. I figured it's better to block for the whole duration of the function so we don't try to read in the middle of the operation.

internal/rest/resources/tokens.go
masnax: I think this is still useful, as we might not be able to reach the other nodes of the truststore for whatever reason. This at least guarantees that since we got the token request, we can answer it and validate it.

internal/trust/remotes.go
masnax: See [below](https://github.com/canonical/microcluster/pull/62#discussion_r1204566152)

PR #65, opened 2023-05-25
cluster/stmt: Split registered statements by caller package

masnax:
If a microcluster project indirectly imports microcluster from another project (i.e. microcloud using microceph, which both use microcluster), then the map of registered statements may end up polluted with entries from that project.

To avoid this, a call to microcluster.Start will record the name of the caller package (whether from snap build, GOPATH, or go module), and record it in the daemon.

As well, on start, when calls to `RegisterStmt` are executed, those packages will be recorded, and statements will be registered by their package.

Then, when preparing statements, only statements belonging to the daemon package, or to 'microcluster' (as that is ubiquitous) will be prepared.

### DIFF
--- a/cluster/stmt.go
+++ b/cluster/stmt.go
@@ -3,10 +3,16 @@ package cluster
 import (
 	"database/sql"
 	"fmt"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+
+	"github.com/lxc/lxd/shared/logger"
 )
 
-var stmts = map[int]string{}            // Statement code to statement SQL text
-var preparedStmts = map[int]*sql.Stmt{} // Statement code to SQL statement.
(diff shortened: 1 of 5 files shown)


stgraber:
I'm confused as to why we need to specify the project in NewDaemon.

Given we have the logic to extract the project from the file path at the time of adding the statements, we should be able to similarly extract the caller's project at the time of preparing them, no?

masnax:
> I'm confused as to why we need to specify the project in NewDaemon.
> 
> Given we have the logic to extract the project from the file path at the time of adding the statements, we should be able to similarly extract the caller's project at the time of preparing them, no?

The package found by `PrepareStmts` will always be the go module path for `microcluster` because it's called internally in `microcluster`. The earliest point where we call into `microcluster` from `microcloud` is in `microcluster.Start`.

stgraber:
Ah, right, that makes sense.

PR #69, opened 2023-07-05
Renames LXD module for new project location

markylaing:


markylaing:
I'm just going to double check this works with a local replace to microceph/microcloud and then unmark as draft

tomponline (approved):

PR #70, opened 2023-07-05
internal/trust/remotes: Fix deadlock when fetching remotes

masnax:
Fixes an (eventual) deadlock where `CertificatesNative` takes a read lock, and then calls `Certificates`, which takes another.


masnax:
Merging immediately as it's rather critical.

PR #72, opened 2023-11-22
Allow passing additional configuration during bootstrap

markylaing:
This PR adds an argument to `(*Microcluster).NewCluster` of the form `map[string]string` which is passed into the `OnBootstrap` hook. This allows clients to pass additional configuration that might be required at bootstrap.

### DIFF
--- a/config/hooks.go
+++ b/config/hooks.go
@@ -8,18 +8,18 @@ import (
 // integrate with other tools.
 type Hooks struct {
 	// OnBootstrap is run after the daemon is initialized and bootstrapped.
-	OnBootstrap func(s *state.State) error
+	OnBootstrap func(s *state.State, initConfig map[string]string) error
 
 	// OnStart is run after the daemon is started.
 	OnStart func(s *state.State) error
 
 	// PostJoin is run after the daemon is initialized, joined the cluster and existing members triggered
 	// their 'OnNewMember' hooks.
-	PostJoin func(s *state.State) error
+	PostJoin func(s *state.State, initConfig map[string]string) error
 
 	// PreJoin is run after the daemon is initialized and joined the cluster but before existing members triggered
 	// their 'OnNewMember' hooks.
-	PreJoin func(s *state.State) error
+	PreJoin func(s *state.State, initConfig map[string]string) error
 
 	// PreRemove is run on a cluster member just before it is removed from the cluster.
 	PreRemove func(s *state.State) error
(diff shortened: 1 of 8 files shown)


masnax:
Since we're taking this approach, I suppose we can rip the bandaid off and add a `force` flag to the remove hooks as well.

masnax (changes requested):
Since not every service using microcluster is going to have MicroCloud's neat and tidy automatic join process, we should probably make this config a more general `InitConfig` and pass it to all of `OnBootstrap`, `PreJoin`, and `PostJoin` so that when going through the join setup, user specified config can get propagated there as well.

markylaing:
> Since we're taking this approach, I suppose we can rip the bandaid off and add a `force` flag to the remove hooks as well.
> 
> Not necessarily in this PR

We don't necessarily need to take this approach. It's still up for discussion :) That being said, I do think this is better than the alternative, which (I think) seems to require the client to make two api calls, one to bootstrap the microcluster, and the other to initialise ceph... which kind of defeats the purpose of the hooks.

markylaing:
> Since not every service using microcluster is going to have MicroCloud's neat and tidy automatic join process, we should probably make this config a more general `InitConfig` and pass it to all of `OnBootstrap`, `PreJoin`, and `PostJoin` so that when going through the join setup, user specified config can get propagated there as well.

Yes. Another thought I had was that in it's current form, this PR does the microcluster bootstrap part before the hook is called. This means that any validation of the config that happens server side will be happening too late. Maybe a `PreBootstrap` hook is necessary now?

masnax:
> > Since not every service using microcluster is going to have MicroCloud's neat and tidy automatic join process, we should probably make this config a more general `InitConfig` and pass it to all of `OnBootstrap`, `PreJoin`, and `PostJoin` so that when going through the join setup, user specified config can get propagated there as well.
> 
> Yes. Another thought I had was that in it's current form, this PR does the microcluster bootstrap part before the hook is called. This means that any validation of the config that happens server side will be happening too late. Maybe a `PreBootstrap` hook is necessary now?

Unlike with `PreJoin` where we actually might have to reconcile information with the existing cluster, there wouldn't be any pre-setup that would have to happen exclusively on the daemon prior to bootstrapping since we won't even have a database or any internal setup, so that would make a `PreBootstrap` hook effectively just a `validateMyConfig` hook, and even then the validation would be limited to sanity checks because we won't have any reference point for what is or isn't a valid config. So I think it's fair to expect the caller to validate the config before sending it through.

UtkarshBhatthere (approved):
LGTM!

markylaing:
@masnax I think I've made the changes you've requested.

masnax (approved):
Looks good, thanks!

PR #73, opened 2023-11-23
Pass `force` to the Pre/PostRemove hooks

masnax:
The deletion API does actually support a `force` flag already, but I noticed that the `PreRemove` hook was actually explicitly blocked when `force=1` was set as a query parameter. This should definitely not be the behaviour, so I went ahead and just propagated the `force` param to each of `PreRemove` (which runs on the to-be-removed node) and `PostRemove`  (which runs on every other node after the node is removed). The `PreRemove` hook will now always fire, even if `force=1`.


masnax:
Need to update MicroOVN to use the latest microcluster which will need some touch-ups to functions so I'm going to merge this one too.

PR #74, opened 2023-11-23
Cancel the listener context before closing it

masnax:
Fixes a misleading error log because we were closing the listeners but not closing their contexts.

### DIFF
--- a/internal/endpoints/network.go
+++ b/internal/endpoints/network.go
@@ -23,11 +23,14 @@ type Network struct {
 	listener net.Listener
 	server   *http.Server
 
-	ctx context.Context
+	ctx    context.Context
+	cancel context.CancelFunc
 }
 
 // NewNetwork assigns an address, certificate, and server to the Network.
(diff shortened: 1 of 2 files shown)


roosterfish (approved):
Looks good to me. 

I was wondering why MicroCloud starts up the servers again after the init process went through? Essentially I don't get who is calling `Serve()` again. Can you please elaborate on this?

masnax:
> Looks good to me.
> 
> I was wondering why MicroCloud starts up the servers again after the init process went through? Essentially I don't get who is calling `Serve()` again. Can you please elaborate on this?

It's an accommodation for MicroCloud, mostly. In order to fetch the disks/interfaces on each system, and also instruct the systems to actually join the cluster, we need to be able to talk to them remotely. So there is an option in microcluster to specify a port which will be used to open a listener on all addresses. Then when MicroCloud has determined which subnet it wants to use for the cluster, it tears down the wildcard listener on each system and replaces it with a proper address:port combo.

roosterfish:
Ah ok, and that is the reason why there is the `err="accept tcp [::]:9443:` wildcard in the error message which indicates the first listener that isn't used anymore. Thanks!

PR #75, opened 2023-11-29
Wait to join cluster before executing new-member hook

masnax:
When there are concurrent cluster joins happening, the `NewMember` hook might fire on nodes that appear as cluster members, but haven't finished fully setting up yet.

### DIFF
--- a/internal/rest/resources/cluster.go
+++ b/internal/rest/resources/cluster.go
@@ -51,6 +51,18 @@ func clusterPost(s *state.State, r *http.Request) response.Response {
 	// If we received a forwarded request, assume the new member was successfully added on the leader,
 	// and execute the new member hook.
 	if client.IsForwardedRequest(r) {
+		ctx, cancel := context.WithTimeout(s.Context, 30*time.Second)
+		defer cancel()
+
+		// Wait for the database to be set up in case we received this request at the same time as joining ourselves.
+		for !s.Database.IsOpen() {
+			select {
+			case <-ctx.Done():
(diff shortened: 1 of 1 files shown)


internal/rest/resources/cluster.go
```diff
@@ -51,6 +51,18 @@ func clusterPost(s *state.State, r *http.Request) response.Response {
 	// If we received a forwarded request, assume the new member was successfully added on the leader,
 	// and execute the new member hook.
 	if client.IsForwardedRequest(r) {
+		ctx, cancel := context.WithTimeout(s.Context, 30*time.Second)
```
tomponline, line 54: Why was 30s picked here?

masnax: It's the same as the query timeout in the corresponding client function.

tomponline: Is that sufficient for slower machines?

masnax: I haven't personally been able to test mich on slower machines, but I wouldn't think this would be a problem. Dqlite itself starts up quickly, it's the ceph and ovn setup that takes a while in MicroCloud at least. 

I stumbled upon this in MicroOVN where it causes a panic in the `OnNewMember` hook trying to use the database. But given that I haven't seen this error reported anywhere and clusters haven't been failing to form at this point, I'd say it's probably sufficient.

tomponline: OK thanks for confirming

PR #77, opened 2024-01-10
Add support for non-cluster members

masnax:
This PR allows for adding trusted systems to a microcluster without having them open a database or join dqlite. These nodes will only have access to the `/cluster/1.0` (public, included) and `/1.0` (public, external) API endpoints, and not the `/cluster/internal` endpoints. 

The role for a cluster member can be one of either `cluster` or `non-cluster`. This is first specified when creating a join token, and is then set in the daemon config at `daemon.yaml` and in the truststore entries with the key `role`. 

All nodes will have truststore entries for all other nodes with the appropriate cluster roles set so that they can maintain trust, but only `cluster` role nodes will be part of the dqlite database. On a join, truststore remotes are immediately propagated to all systems, and on a node removal, the remote is removed from all systems. 

`non-cluster` nodes have the ability to upgrade into `cluster` nodes. Any node can request another node to be upgraded. The node to-be-removed then instructs a random dqlite member to find the dqlite leader, which will update the database and re-assign the role for the node in the truststore.

Joining and removal will re-use the same hooks, except that hooks that run on the whole cluster will only run on dqlite members. For upgrading, there is a new set of 3 hooks which works similarly to the join hooks.


markylaing:
@masnax can you link the JIRA ticket or give some background info on the use case for this please? I'm a little confused as to what this will be used for.

masnax:
@markylaing @tomponline This was requested by MicroCeph to add the capability of large-scale ceph deployment with microcluster without the overhead of dqlite. The goal is to effectively have ~500 systems set up with microcluster that run ceph, with a smaller subset of dqlite cluster members that would be controllers.

markylaing:
> @markylaing @tomponline This was requested by MicroCeph to add the capability of large-scale ceph deployment with microcluster without the overhead of dqlite. The goal is to effectively have ~500 systems set up with microcluster that run ceph, with a smaller subset of dqlite cluster members that would be controllers.
> 
> The external APIs that are specified to microcluster would still be accessible by these nodes, but any sort of database access would have to go through a controller.

Ah that makes sense, thanks.

markylaing:
@masnax I'm just now getting around to this (sorry!). This PR updates both the database schema and the API, and given that more and more canonical applications are using it I think it would be worth writing a spec with the changes. Additionally, it would be good to see an open API specification before and after.

So we would have:
1. A separate PR that lands some commits that are small bugfixes/improvements and are not absolutely required for this feature.
2. A separate PR for an OpenAPI spec for the existing API.
3. A specification outlining the changes you're making.
4. A final PR to land the commits specific to this feature.

If there is significant time pressure on this no worries, just let me know and I'll come back to it as soon as I can.

UtkarshBhatthere:
Thanks (a ton) for this change @masnax.
@markylaing it was a feature requested by MicroCeph to help navigate scalability constructs of dqlite. However, It is not time critical at this moment.

masnax:
@markylaing @roosterfish @tomponline @UtkarshBhatthere a specification for this feature can be found here: 
https://discourse.ubuntu.com/t/add-support-for-trusted-qnon-cluster-members/43182

markylaing:
@masnax shall we close or set as draft this for now while we're still discussing the spec and potential changes to DQLite?

tomponline:
closing for now.

PR #80, opened 2024-01-19
internal/rest/resources: Refresh cluster info when assigning a role to another node

### LINKED ISSUE #79: Fail to scale down from 2 to 1 when removed node is voter
Microcluster fails to scale down when there's two members, and the one leaving is the voter.

Error: `Found no voters to transfer leadership`

When the one leaving is the spare member, scale down works.

gboutry:
FIX #79 

In the case the cluster is composed of 2 members, and you try to remove the voter. The code successfully transfers leadership to the spare member, but since cluster info is not updated, subsequent checks fail with "Found no voters to transfer leadership to".

### DIFF
--- a/internal/rest/resources/cluster.go
+++ b/internal/rest/resources/cluster.go
@@ -461,13 +461,14 @@ func clusterMemberDelete(s *state.State, r *http.Request) response.Response {
 				if err != nil {
 					return response.SmartError(err)
 				}
-				// Roles for second member has been updated, so re-query the cluster.
-				info, err = leader.Cluster(s.Context)
-				if err != nil {
-					return response.SmartError(err)
-				}
 			}
 		}
+	}
+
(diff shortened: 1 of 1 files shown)


masnax (changes requested):
Good catch, thanks for this!

Could you please move the new call to `leader.Cluster` to just outside the for-loop? The current roles where you are seeing the error are checked in the following block anyway, so we shouldn't need to fetch the whole list in each iteration. It might also cause some problems if a node managed to join at that exact moment.

gboutry:
I can, but in this exact scenario, the call is performed only once. Since the for loop is always just 2 turns, and only applied when successfully assigning role to the spare member. I thought it was a good place to set it.

But indeed, other nodes might have joined in the interim. I will move it after the loop.

masnax (approved):
Thanks!

PR #81, opened 2024-01-22
internal/db/update: Drop implicit global state

neoaggelos:
### Summary

Refactor the `internal/db/update` package such that it does not carry any implicit global state for the database migrations.

### Description

When trying to instantiate more than one microcluster instance, the global `updates` state would perform duplicate migrations on instances other than the first and thus fail.

### Proposed change

- Create a `update.SchemaUpdateManager` type with the `Schema()`, `AppendSchema()`, `SchemaDotGo()` methods.
- Create a `globalSchemaUpdateManager` object
- Preserve the old `update.Schema()`, `update.AppendSchema()`, `update.SchemaDotGo()` functions, so that we don't break API. These use the global object, being backwards compatible.
- Update the microcluster instance creation to not use the global schema, but create its own schema instance and use that for initialization.

### Background

In https://github.com/canonical/k8s-snap/pull/69, we introduce tests that initialize separate microcluster instances and validate our database logic. These currently fail because creating a second microcluster instance attempts to apply the migrations multiple times.

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -24,7 +24,6 @@ import (
 	"github.com/canonical/microcluster/cluster"
 	"github.com/canonical/microcluster/config"
 	"github.com/canonical/microcluster/internal/db"
-	"github.com/canonical/microcluster/internal/db/update"
 	"github.com/canonical/microcluster/internal/endpoints"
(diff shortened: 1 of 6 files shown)


masnax (approved):
Thanks for this!

PR #82, opened 2024-01-23
Allow replacing cluster certificate

masnax:
Adds a new internal endpoint to update the cluster certificate used with the listeners across the cluster. It sends a request to all cluster members and instructs them to write the certs into the state directory, and load them onto the daemon.

To facilitate this by default before the endpoints start, a new `PreBootstrap` hook is introduced which runs just after the daemon registers its name & address. The default cluster cert can be overwritten in the state directory at this point and get picked up during the bootstrap process. `OnBootstrap` has been renamed to `PostBootstrap` to keep things consistent.

Additionally, there's a client function `UpdateClusterCertificate` which will update the cluster cert after the daemon has started.

cc @gboutry


example/cmd/microd/main.go
```diff
@@ -69,7 +69,7 @@ func (c *cmdDaemon) Run(cmd *cobra.Command, args []string) error {
 	// exampleHooks are some example post-action hooks that can be run by MicroCluster.
 	exampleHooks := &config.Hooks{
 		// OnBootstrap is run after the daemon is initialized and bootstrapped.
```
roosterfish, line 71: The comment also needs an update to `PostBootstrap`.

example/cmd/microd/main.go
```diff
@@ -81,6 +81,18 @@ func (c *cmdDaemon) Run(cmd *cobra.Command, args []string) error {
 			return nil
 		},
 
+		PreBootstrap: func(s *state.State, initConfig map[string]string) error {
```
roosterfish, line 84: Just for my understanding. I guess when you import `MicroCluster` you can overwrite those hooks with your own ones?

internal/daemon/daemon.go
```diff
@@ -380,7 +391,7 @@ func (d *Daemon) StartAPI(bootstrap bool, initConfig map[string]string, newConfi
 			return err
 		}
 
-		return d.hooks.OnBootstrap(d.State(), initConfig)
+		return d.hooks.PostBootstrap(d.State(), initConfig)
```
roosterfish, line 397: I am wondering if we should handle the `error` returned from `PostBootstrap` explicitly meaning adding the description as for `PreBootstrap` above. Something like `Failed to run post-bootstrap hook before starting the API: %w`

config/hooks.go
```diff
@@ -7,8 +7,11 @@ import (
 // Hooks holds customizable functions that can be called at varying points by the daemon to.
 // integrate with other tools.
 type Hooks struct {
-	// OnBootstrap is run after the daemon is initialized and bootstrapped.
-	OnBootstrap func(s *state.State, initConfig map[string]string) error
+	// PreBootstrap is run before the daemon is initialized and bootstrapped.
```
roosterfish, line 10: Does this break any compatibility with libraries that import `MicroCluster`? I guess the change would land right away in `main`.

internal/daemon/daemon.go
```diff
@@ -492,6 +495,9 @@ func (d *Daemon) StartAPI(bootstrap bool, initConfig map[string]string, newConfi
 
 // ClusterCert ensures both the daemon and state have the same cluster cert.
 func (d *Daemon) ClusterCert() *shared.CertInfo {
+	d.clusterMu.RLock()
```
roosterfish, line 498: Since the pointer to `d.clusterCert` is returned, wouldn't this break the logic of the read lock if the caller reads from the pointers address afterwards?

internal/endpoints/network.go
```diff
@@ -71,6 +71,12 @@ func (n *Network) Listen() error {
 	return nil
 }
 
+// UpdateTLS updates the TLS configuration of the network listener.
+func (n *Network) UpdateTLS(cert *shared.CertInfo) {
+	n.cert = cert
+	n.listener.(*listeners.FancyTLSListener).Config(cert)
```
roosterfish, line 77: To prevent running into a panic why not switching to:

```go
c, ok := n.listener.(*listeners.FancyTLSListener)
if ok {
    c.Config(cert)
}
```

Maybe also log an error in case this didn't work as expected?

internal/daemon/daemon.go
```diff
@@ -501,6 +501,23 @@ func (d *Daemon) ClusterCert() *shared.CertInfo {
...
+	defer d.clusterMu.Unlock()
+
+	clusterCert, err := util.LoadClusterCert(d.os.StateDir)
+	if err != nil {
+		return err
+	}
+
+	d.clusterCert = clusterCert
+	d.endpoints.UpdateTLS(clusterCert)
+	d.db.ClusterCert = clusterCert
```
roosterfish, line 516: Maybe instead of exporting `db.ClusterCert` we can have a function `func (db *DB) UpdateClusterCert()` to keep things separated?

roosterfish (changes requested):
Just some smaller comments :)

example/cmd/microd/main.go
masnax: Yes, you can specify your own set of hooks. For example with MicroCloud we use `PostBootstrap` and `PostJoin` to stop the mDNS broadcast of our availability, since we just finished bootstrapping or joining a cluster:

https://github.com/canonical/microcloud/blob/82613aa8398252af24f27eb26109b185531cefaa/microcloud/service/microcloud.go#L57-L63

config/hooks.go
masnax: Yeah, they would all have to rename `OnBootstrap` to `PostBootstrap`. I suppose we should begin tagging versions soon, given both this change and the non-cluster member PR will introduce some changes.

internal/daemon/daemon.go
masnax: Good catch, thanks

config/hooks.go
roosterfish: Ok thanks, I didn't check if there are any releases yet.

internal/endpoints/endpoints.go
```diff
@@ -38,6 +39,19 @@ func (e *Endpoints) Up() error {
...
+// UpdateTLS updates the TLS configuration of the network listeners.
+func (e *Endpoints) UpdateTLS(cert *shared.CertInfo) {
+	e.mu.Lock()
+	defer e.mu.Unlock()
+
+	for _, l := range e.listeners {
+		n, ok := l.(*Network)
+		if ok {
+			n.UpdateTLS(cert)
+		}
```
roosterfish, line 51: Is there any way to log an error if the casting didn't work? Not sure if applicable :)

internal/db/dqlite.go
```diff
@@ -191,6 +191,11 @@ func (db *DB) NotifyUpgraded() {
 	}
 }
 
+// SetClusterCert updates the cluster cert.
+func (db *DB) SetClusterCert(clusterCert *shared.CertInfo) {
+	db.clusterCert = clusterCert
```
roosterfish, line 196: Does this need some locking in case some other routine is reading the old certificate?

roosterfish (commented):
Sorry for the late reply, looking good, just two more comments I am not sure about.

internal/endpoints/endpoints.go
masnax: So not every `Endpoint` is a `Network` type, because there's also unix socket listeners which we want to quietly skip in this loop. Additionally, if the daemon isn't initialized yet, there may be no Network listener at all yet. It would still be valid to change the cluster cert at this point, so we shouldn't necessarily error out if we can't find the Endpoint.

roosterfish: Ok, thanks!

internal/db/dqlite.go
masnax: Instead of going with this approach, I just decided to do what we already do with the `State` struct and pass the daemon's already thread-safe clusterCert function to the `DB` struct. So now whenever the daemon's cluster cert is updated, the db's should be as well as it's calling the same function.

e8b509879cf6d38b6e0b4cd0bc9030e22bbdf815

roosterfish (approved):
LGTM!

PR #83, opened 2024-01-31
internal/rest/client: Use TLSClientConfig with the client dial func

masnax:
The tlsConfig was previously hard-coded to what the internal client set, and was unaffected by changes to TLSClientConfig. This updates the client to instead use the same TLSClientConfig for the dialer as well.

This is particularly useful for MicroCloud where we want to set `InsecureSkipVerify` as we are overriding TLS authentication in favour of the auth secret obtained over mDNS until the system is set up.

### DIFF
--- a/internal/rest/client/client.go
+++ b/internal/rest/client/client.go
@@ -118,48 +118,53 @@ func tlsHTTPClient(clientCert *shared.CertInfo, remoteCert *x509.Certificate, pr
 
 	}
 
-	tlsDialContext := func(ctx context.Context, network string, addr string) (net.Conn, error) {
-		host, port, err := net.SplitHostPort(addr)
-		if err != nil {
-			return nil, err
-		}
-
-		addrs, err := net.LookupHost(host)
-		if err != nil {
-			return nil, err
-		}
-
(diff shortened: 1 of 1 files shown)


markylaing (approved):
LGTM

tomponline:
> This is particularly useful for MicroCloud where we want to set `InsecureSkipVerify` as we are overriding TLS authentication in favour of the auth secret obtained over mDNS until the system is set up.

When you come to do that, can you point it out to me, ta.

tomponline (approved):

masnax:
> > This is particularly useful for MicroCloud where we want to set `InsecureSkipVerify` as we are overriding TLS authentication in favour of the auth secret obtained over mDNS until the system is set up.
> 
> When you come to do that, can you point it out to me, ta.

We kind of already are, but in a poor way. We're actually ripping out the entire dial func by replacing the whole `http.Transport`. This lets us at least preserve the existing `http.Transport`.

https://github.com/canonical/microcloud/blob/63736afc5080f03398b1beb991d4ed4df34ae57f/microcloud/service/microcloud.go#L113-L121

PR #84, opened 2024-02-02
Add minimal CI tests

simondeziel:
@masnax I made the needed changes to please golangci-lint but revive was too much so I made the `check-static` warn instead of hard failing for now.

I looked at building it but it wants `lxd-generate` and I couldn't quickly find where to pull it from. Something for a future PR :)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    labels: []
+    schedule:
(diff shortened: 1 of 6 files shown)


simondeziel:
@masnax OK, thanks I've added the corresponding `make -C example` calls to the tests job.

masnax (approved):
LGTM, though did we always have those github actions errors about package comments? I don't believe any of our projects have strictly required package comments.

simondeziel:
AFAIK, revive isn't used much elsewhere.

PR #86, opened 2024-02-12
Add a new 'runtime extension' system to MicroCluster

gabrielmougard:
Introducing an 'extensions' endpoint in MicroCluster would help in scenarios where service A (backed by MicroCluster) needs to check that service B (also backed by MicroCluster) has a certain feature (e.g, does the installed version of MicroOVN supports custom encapsulation ip that has been entered by a user in a MicroCloud setup)

Required by https://github.com/canonical/microovn/pull/121 and https://github.com/canonical/microovn/pull/113 and https://github.com/canonical/microcloud/pull/245

### DIFF
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -15,12 +15,13 @@ import (
 	"github.com/canonical/lxd/shared/logger"
 
 	"github.com/canonical/microcluster/cluster"
+	"github.com/canonical/microcluster/internal/extensions"
 	"github.com/canonical/microcluster/internal/sys"
 )
 
 // Open opens the dqlite database and loads the schema.
 // Returns true if we need to wait for other nodes to catch up to our version.
-func (db *DB) Open(bootstrap bool, project string) error {
+func (db *DB) Open(ext extensions.Extensions, bootstrap bool, project string) error {
 	ctx, cancel := context.WithTimeout(db.ctx, 30*time.Second)
 	defer cancel()
 
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -35,7 +35,7 @@ func (db *DB) Open(ext extensions.Extensions, bootstrap bool, project string) er
 		return err
 	}
 
-	err = db.waitUpgrade(bootstrap)
+	err = db.waitUpgrade(bootstrap, ext)
 	if err != nil {
 		return err
 	}
@@ -53,7 +53,7 @@ func (db *DB) Open(ext extensions.Extensions, bootstrap bool, project string) er
 // waitUpgrade compares the version information of all cluster members in the database to the local version.
 // If this node's version is ahead of others, then it will block on the `db.upgradeCh` or up to a minute.
 // If this node's version is behind others, then it returns an error.
-func (db *DB) waitUpgrade(bootstrap bool) error {
+func (db *DB) waitUpgrade(bootstrap bool, ext extensions.Extensions) error {
 	checkSchemaVersion := func(schemaVersion uint64, clusterMemberVersions []uint64) (otherNodesBehind bool, err error) {
 		nodeIsBehind := false
 		for _, version := range clusterMemberVersions {
@@ -76,6 +76,33 @@ func (db *DB) waitUpgrade(bootstrap bool) error {
 			// to upgrade. Let's error out and shutdown
 			// since we need a greater version.
 			return false, fmt.Errorf("This node's version is behind, please upgrade")
+		}
+
+		return nodeIsBehind, nil
+	}
+
+	checkAPIExtensions := func(currentAPIExtensions extensions.Extensions, clusterMemberAPIExtensions []extensions.Extensions) (otherNodesBehind bool, err error) {
+		logger.Warnf("Local API extensions: %v, cluster members API extensions: %v", currentAPIExtensions, clusterMemberAPIExtensions)
+
+		nodeIsBehind := false
+		for _, extensions := range clusterMemberAPIExtensions {
+			if currentAPIExtensions.IsSameVersion(extensions) == nil {
+				// API extensions are equal, there's hope for the
+				// update. Let's check the next node.
+				continue
+			} else if extensions == nil || currentAPIExtensions.Version() > extensions.Version() {
+				// Our version is bigger, we should stop here
+				// and wait for other nodes to be upgraded and
+				// restarted.
+				nodeIsBehind = true
+				continue
+			} else {
+				// Another node has a version greater than ours
+				// and presumeably is waiting for other nodes
+				// to upgrade. Let's error out and shutdown
+				// since we need a greater version.
+				return false, fmt.Errorf("This node's API extensions are behind, please upgrade")
+			}
 		}
 
 		return nodeIsBehind, nil
@@ -122,7 +149,42 @@ func (db *DB) waitUpgrade(bootstrap bool) error {
 
 	err := db.retry(context.TODO(), func(_ context.Context) error {
 		_, err := newSchema.Ensure(db.db)
-		return err
+		if err != nil {
+			return err
+		}
+
+		if !bootstrap {
+			otherNodesBehindAPI := false
+			// Perform the API extensions check.
+			err = query.Transaction(context.TODO(), db.db, func(ctx context.Context, tx *sql.Tx) error {
+				err := cluster.UpdateClusterMemberAPIExtensions(tx, ext, db.listenAddr.URL.Host)
+				if err != nil {
+					return fmt.Errorf("Failed to update API extensions when joining cluster: %w", err)
+				}
+
+				clusterMembersAPIExtensions, err := cluster.GetClusterMemberAPIExtensions(ctx, tx)
+				if err != nil {
+					return fmt.Errorf("Failed to get other members' API extensions: %w", err)
+				}
+
+				otherNodesBehindAPI, err = checkAPIExtensions(ext, clusterMembersAPIExtensions)
+				if err != nil {
+					return err
+				}
+
+				return nil
+			})
+			if err != nil {
+				return err
+			}
+
+			if otherNodesBehindAPI {
+				otherNodesBehind = true
+				return schema.ErrGracefulAbort
+			}
+		}
+
+		return nil
 	})
 
 	// If we are not bootstrapping, wait for an upgrade notification, or wait a minute before checking again.
@@ -130,7 +192,7 @@ func (db *DB) waitUpgrade(bootstrap bool) error {
 		logger.Warn("Waiting for other cluster members to upgrade their versions", logger.Ctx{"address": db.listenAddr.String()})
 		select {
 		case <-db.upgradeCh:
-		case <-time.After(time.Minute):
+		case <-time.After(30 * time.Second):
 		}
 	}
 
--- a/internal/db/db_test.go
+++ b/internal/db/db_test.go
@@ -3,6 +3,7 @@ package db
 import (
 	"context"
 	"database/sql"
+	"encoding/json"
 	"fmt"
 	"testing"
 	"time"
@@ -14,6 +15,7 @@ import (
 
 	"github.com/canonical/microcluster/cluster"
 	"github.com/canonical/microcluster/internal/db/update"
+	"github.com/canonical/microcluster/internal/extensions"
 )
 
 type dbSuite struct {
@@ -25,7 +27,7 @@ func TestDBSuite(t *testing.T) {
 }
 
 // Ensures waitUpgrade properly waits or returns an error with mismatched cluster versions.
-func (s *dbSuite) Test_waitUpgrade() {
+func (s *dbSuite) Test_waitUpgradeSchema() {
 	type versions struct {
 		schemaInt uint64
 		schemaExt uint64
@@ -162,6 +164,9 @@ func (s *dbSuite) Test_waitUpgrade() {
 		tx, err := db.db.BeginTx(ctx, nil)
 		s.NoError(err)
 
+		apiExtensions, err := extensions.NewExtensionRegistry(true)
+		s.NoError(err)
+
 		// Generate a cluster member for the local node.
 		_, err = cluster.CreateInternalClusterMember(ctx, tx, cluster.InternalClusterMember{
 			Name:           fmt.Sprintf("cluster-member-%d", 0),
@@ -169,6 +174,7 @@ func (s *dbSuite) Test_waitUpgrade() {
 			Certificate:    fmt.Sprintf("test-cert-%d", 0),
 			SchemaInternal: 0,
 			SchemaExternal: 0,
+			APIExtensions:  apiExtensions,
 			Heartbeat:      time.Time{},
 			Role:           "voter",
 		})
@@ -183,6 +189,7 @@ func (s *dbSuite) Test_waitUpgrade() {
 				Certificate:    fmt.Sprintf("test-cert-%d", j+1),
 				SchemaInternal: clusterMember.schemaInt + 1,
 				SchemaExternal: clusterMember.schemaExt + 1,
+				APIExtensions:  apiExtensions,
 				Heartbeat:      time.Time{},
 				Role:           "voter",
 			})
@@ -228,7 +235,7 @@ func (s *dbSuite) Test_waitUpgrade() {
 		db.schema = manager.Schema()
 
 		// Run the upgrade function to wait, error, or succeed.
-		err = db.waitUpgrade(false)
+		err = db.waitUpgrade(false, apiExtensions)
 		if t.expectErr != nil {
 			s.Equal(t.expectErr, err)
 		} else if t.expectWait {
@@ -275,6 +282,368 @@ func (s *dbSuite) Test_waitUpgrade() {
 
 		s.Equal(internalVersions, schemaInternal)
 		s.Equal(externalVersions, schemaExternal)
+	}
+}
+
+func (s *dbSuite) Test_waitUpgradeAPI() {
+	tests := []struct {
+		name                       string
+		upgradedLocalAPIExtensions extensions.Extensions
+		clusterMembersExtensions   []extensions.Extensions
+		expectErr                  error
+		expectWait                 bool
+	}{
+		{
+			name:                       "No upgrade, no other nodes",
+			upgradedLocalAPIExtensions: extensions.Extensions{},
+		},
+		{
+			name:                       "API upgrade, no other nodes",
+			upgradedLocalAPIExtensions: extensions.Extensions{"internal:a", "ext"},
+		},
+		{
+			name:                       "All other nodes ahead",
+			upgradedLocalAPIExtensions: extensions.Extensions{"internal:a", "ext"},
+			clusterMembersExtensions:   []extensions.Extensions{{"internal:a", "ext", "ext2"}, {"internal:a", "ext", "ext2"}},
+			expectErr:                  fmt.Errorf("This node's API extensions are behind, please upgrade"),
+		},
+		{
+			name:                       "Some other nodes ahead",
+			upgradedLocalAPIExtensions: extensions.Extensions{"internal:a", "ext"},
+			clusterMembersExtensions:   []extensions.Extensions{{"internal:a", "ext"}, {"internal:a", "ext", "ext2"}},
+			expectErr:                  fmt.Errorf("This node's API extensions are behind, please upgrade"),
+		},
+		{
+			name:                       "All other nodes behind",
+			upgradedLocalAPIExtensions: extensions.Extensions{"internal:a", "ext", "ext2"},
+			clusterMembersExtensions:   []extensions.Extensions{{"internal:a", "ext"}, {"internal:a", "ext"}},
+			expectWait:                 true,
+		},
+		{
+			name:                       "Some other nodes behind",
+			upgradedLocalAPIExtensions: extensions.Extensions{"internal:a", "ext", "ext2"},
+			clusterMembersExtensions:   []extensions.Extensions{{"internal:a", "ext", "ext2"}, {"internal:a", "ext"}},
+			expectWait:                 true,
+		},
+		{
+			name:                       "Some nodes behind, others ahead",
+			upgradedLocalAPIExtensions: extensions.Extensions{"internal:a", "ext", "ext2"},
+			clusterMembersExtensions:   []extensions.Extensions{{"internal:a", "ext"}, {"internal:a", "ext", "ext2", "ext3"}},
+			expectErr:                  fmt.Errorf("This node's API extensions are behind, please upgrade"),
+		},
+	}
+
+	for i, t := range tests {
+		s.T().Logf("%s (case %d)", t.name, i)
+
+		db, err := NewTestDB([]schema.Update{})
+		s.NoError(err)
+
+		ctx := context.Background()
+		tx, err := db.db.BeginTx(ctx, nil)
+		s.NoError(err)
+
+		// Generate a cluster member for the local node.
+		_, err = cluster.CreateInternalClusterMember(ctx, tx, cluster.InternalClusterMember{
+			Name:           fmt.Sprintf("cluster-member-%d", 0),
+			Address:        fmt.Sprintf("10.0.0.%d:8443", 0),
+			Certificate:    fmt.Sprintf("test-cert-%d", 0),
+			SchemaInternal: 0,
+			SchemaExternal: 0,
+			APIExtensions:  t.upgradedLocalAPIExtensions,
+			Heartbeat:      time.Time{},
+			Role:           "voter",
+		})
+
+		s.NoError(err)
+
+		// Generate a cluster member entry for all other expected nodes.
+		for j, clusterMemberExtensions := range t.clusterMembersExtensions {
+			_, err = cluster.CreateInternalClusterMember(ctx, tx, cluster.InternalClusterMember{
+				Name:           fmt.Sprintf("cluster-member-%d", j+1),
+				Address:        fmt.Sprintf("10.0.0.%d:8443", j+1),
+				Certificate:    fmt.Sprintf("test-cert-%d", j+1),
+				SchemaInternal: 1,
+				SchemaExternal: 1,
+				APIExtensions:  clusterMemberExtensions,
+				Heartbeat:      time.Time{},
+				Role:           "voter",
+			})
+
+			s.NoError(err)
+		}
+
+		s.NoError(tx.Commit())
+
+		// Reset the local schema so we can override the updates.
+		_, err = db.db.Exec("delete from schemas")
+		s.NoError(err)
+
+		stmt := `INSERT INTO schemas (version, type, updated_at) VALUES (?, ?, strftime("%s"))`
+
+		// Apply the local updates to the schema table.
+		manager := &update.SchemaUpdateManager{}
+
+		_, err = db.db.Exec(stmt, 0, 0)
+		s.NoError(err)
+		updates := []schema.Update{func(ctx context.Context, tx *sql.Tx) error { return nil }}
+		manager.SetInternalUpdates(updates)
+
+		_, err = db.db.Exec(stmt, 0, 1)
+		s.NoError(err)
+		updates = []schema.Update{func(ctx context.Context, tx *sql.Tx) error { return nil }}
+		manager.SetExternalUpdates(updates)
+
+		if t.expectWait {
+			db.upgradeCh <- struct{}{}
+		}
+
+		// Set a no-op schema manager so that we can go through the schema upgrade logic without running any real updates.
+		db.schema = manager.Schema()
+
+		// Run the upgrade function to wait, error, or succeed.
+		err = db.waitUpgrade(false, t.upgradedLocalAPIExtensions)
+		if t.expectErr != nil {
+			s.Equal(t.expectErr, err)
+		} else if t.expectWait {
+			s.Equal(schema.ErrGracefulAbort, err)
+		} else {
+			s.NoError(err)
+		}
+
+		tx, err = db.db.BeginTx(ctx, nil)
+		s.NoError(err)
+
+		res, err := query.SelectStrings(ctx, tx, "SELECT api_extensions FROM internal_cluster_members ORDER BY id")
+		s.NoError(err)
+		allExtensions := make([]extensions.Extensions, 0)
+		for _, r := range res {
+			e := extensions.Extensions{}
+			err = json.Unmarshal([]byte(r), &e)
+			s.NoError(err)
+			allExtensions = append(allExtensions, e)
+		}
+
+		s.NoError(tx.Commit())
+
+		s.Equal(len(t.clusterMembersExtensions)+1, len(allExtensions))
+
+		if t.expectErr == nil && t.expectWait == false {
+			for _, c := range t.clusterMembersExtensions {
+				err = t.upgradedLocalAPIExtensions.IsSameVersion(c)
+				s.NoError(err)
+			}
+		}
+	}
+}
+
+func (s *dbSuite) Test_waitUpgradeSchemaAndAPI() {
+	type versionsWithExtensions struct {
+		schemaInt uint64
+		schemaExt uint64
+		ext       extensions.Extensions
+	}
+
+	tests := []struct {
+		name              string
+		upgradedLocalInfo versionsWithExtensions
+		clusterMembers    []versionsWithExtensions
+		expectErr         error
+		expectWait        bool
+	}{
+		{
+			name: "Local node in sync, no other nodes",
+			upgradedLocalInfo: versionsWithExtensions{
+				schemaInt: 0,
+				schemaExt: 0,
+				ext:       extensions.Extensions{"internal:a"},
+			},
+		},
+		{
+			name: "Local node behind in schema and API",
+			upgradedLocalInfo: versionsWithExtensions{
+				schemaInt: 0,
+				schemaExt: 0,
+				ext:       extensions.Extensions{"internal:a"},
+			},
+			clusterMembers: []versionsWithExtensions{
+				{schemaInt: 1, schemaExt: 1, ext: extensions.Extensions{"internal:a"}},
+				{schemaInt: 1, schemaExt: 1, ext: extensions.Extensions{"internal:a"}},
+			},
+			expectErr: fmt.Errorf("This node's version is behind, please upgrade"),
+		},
+		{
+			name: "Local node ahead, all other nodes behind",
+			upgradedLocalInfo: versionsWithExtensions{
+				schemaInt: 2,
+				schemaExt: 2,
+				ext:       extensions.Extensions{"internal:a", "b", "c"},
+			},
+			clusterMembers: []versionsWithExtensions{
+				{schemaInt: 1, schemaExt: 1, ext: extensions.Extensions{"internal:a", "b", "c"}},
+				{schemaInt: 1, schemaExt: 1, ext: extensions.Extensions{"internal:a", "b", "c"}},
+			},
+			expectWait: true,
+		},
+		{
+			name: "Mixed node versions, some ahead and some behind",
+			upgradedLocalInfo: versionsWithExtensions{
+				schemaInt: 1,
+				schemaExt: 1,
+				ext:       extensions.Extensions{"internal:a", "b"},
+			},
+			clusterMembers: []versionsWithExtensions{
+				{schemaInt: 2, schemaExt: 2, ext: extensions.Extensions{"internal:a", "b"}},
+				{schemaInt: 0, schemaExt: 0, ext: extensions.Extensions{"internal:a", "b"}},
+			},
+			expectErr: fmt.Errorf("This node's version is behind, please upgrade"),
+		},
+	}
+
+	for i, t := range tests {
+		s.T().Logf("%s (case %d)", t.name, i)
+
+		db, err := NewTestDB([]schema.Update{})
+		s.NoError(err)
+
+		ctx := context.Background()
+		tx, err := db.db.BeginTx(ctx, nil)
+		s.NoError(err)
+
+		// Generate a cluster member for the local node.
+		_, err = cluster.CreateInternalClusterMember(ctx, tx, cluster.InternalClusterMember{
+			Name:           fmt.Sprintf("cluster-member-%d", 0),
+			Address:        fmt.Sprintf("10.0.0.%d:8443", 0),
+			Certificate:    fmt.Sprintf("test-cert-%d", 0),
+			SchemaInternal: 0,
+			SchemaExternal: 0,
+			APIExtensions:  extensions.Extensions{},
+			Heartbeat:      time.Time{},
+			Role:           "voter",
+		})
+
+		s.NoError(err)
+
+		// Generate a cluster member entry for all other expected nodes.
+		for j, clusterMember := range t.clusterMembers {
+			_, err = cluster.CreateInternalClusterMember(ctx, tx, cluster.InternalClusterMember{
+				Name:           fmt.Sprintf("cluster-member-%d", j+1),
+				Address:        fmt.Sprintf("10.0.0.%d:8443", j+1),
+				Certificate:    fmt.Sprintf("test-cert-%d", j+1),
+				SchemaInternal: clusterMember.schemaInt + 1,
+				SchemaExternal: clusterMember.schemaExt + 1,
+				APIExtensions:  clusterMember.ext,
+				Heartbeat:      time.Time{},
+				Role:           "voter",
+			})
+
+			s.NoError(err)
+		}
+
+		s.NoError(tx.Commit())
+
+		// Reset the local schema so we can override the updates.
+		_, err = db.db.Exec("delete from schemas")
+		s.NoError(err)
+
+		stmt := `INSERT INTO schemas (version, type, updated_at) VALUES (?, ?, strftime("%s"))`
+
+		// Apply the local updates to the schema table.
+		manager := &update.SchemaUpdateManager{}
+		updates := []schema.Update{}
+		for j := 0; j < int(t.upgradedLocalInfo.schemaInt)+1; j++ {
+			_, err = db.db.Exec(stmt, j, 0)
+			s.NoError(err)
+
+			updates = append(updates, func(ctx context.Context, tx *sql.Tx) error { return nil })
+		}
+
+		manager.SetInternalUpdates(updates)
+
+		updates = []schema.Update{}
+		for j := 0; j < int(t.upgradedLocalInfo.schemaExt)+1; j++ {
+			_, err = db.db.Exec(stmt, j, 1)
+			s.NoError(err)
+
+			updates = append(updates, func(ctx context.Context, tx *sql.Tx) error { return nil })
+		}
+
+		manager.SetExternalUpdates(updates)
+
+		if t.expectWait {
+			db.upgradeCh <- struct{}{}
+		}
+
+		// Set a no-op schema manager so that we can go through the schema upgrade logic without running any real updates.
+		db.schema = manager.Schema()
+
+		// Run the upgrade function to wait, error, or succeed.
+		err = db.waitUpgrade(false, t.upgradedLocalInfo.ext)
+		if t.expectErr != nil {
+			s.Equal(t.expectErr, err)
+		} else if t.expectWait {
+			s.Equal(schema.ErrGracefulAbort, err)
+		} else {
+			s.NoError(err)
+		}
+
+		tx, err = db.db.BeginTx(ctx, nil)
+		s.NoError(err)
+
+		schemaInternal, err := query.SelectIntegers(ctx, tx, "SELECT schema_internal FROM internal_cluster_members ORDER BY id")
+		s.NoError(err)
+
+		schemaExternal, err := query.SelectIntegers(ctx, tx, "SELECT schema_external FROM internal_cluster_members ORDER BY id")
+		s.NoError(err)
+
+		res, err := query.SelectStrings(ctx, tx, "SELECT api_extensions FROM internal_cluster_members ORDER BY id")
+		s.NoError(err)
+		allExtensions := make([]extensions.Extensions, 0)
+		for _, r := range res {
+			e := extensions.Extensions{}
+			err = json.Unmarshal([]byte(r), &e)
+			s.NoError(err)
+			allExtensions = append(allExtensions, e)
+		}
+
+		s.NoError(tx.Commit())
+
+		s.Equal(len(t.clusterMembers)+1, len(schemaInternal))
+		s.Equal(len(t.clusterMembers)+1, len(schemaExternal))
+
+		// If there's an error, the internal_cluster_members table will be rolled back.
+		minVersion := t.upgradedLocalInfo.schemaInt + 1
+		if t.expectErr != nil {
+			minVersion = 0
+		}
+
+		internalVersions := []int{int(minVersion)}
+		for _, c := range t.clusterMembers {
+			internalVersions = append(internalVersions, int(c.schemaInt)+1)
+		}
+
+		// If there's an error, the internal_cluster_members table will be rolled back.
+		minVersion = t.upgradedLocalInfo.schemaExt + 1
+		if t.expectErr != nil {
+			minVersion = 0
+		}
+
+		externalVersions := []int{int(minVersion)}
+		for _, c := range t.clusterMembers {
+			externalVersions = append(externalVersions, int(c.schemaExt)+1)
+		}
+
+		s.Equal(internalVersions, schemaInternal)
+		s.Equal(externalVersions, schemaExternal)
+
+		s.Equal(len(t.clusterMembers)+1, len(allExtensions))
+
+		if t.expectErr == nil && t.expectWait == false {
+			for _, c := range t.clusterMembers {
+				err = t.upgradedLocalInfo.ext.IsSameVersion(c.ext)
+				s.NoError(err)
+			}
+		}
 	}
 }
 
--- a/internal/db/update/update.go
+++ b/internal/db/update/update.go
@@ -28,6 +28,7 @@ func NewSchema() *SchemaUpdateManager {
 			updateInternal: {
 				updateFromV0,
 				updateFromV1,
+				updateFromV2,
 			},
 		},
 	}
@@ -59,6 +60,32 @@ func (m *SchemaUpdateManager) Schema() *SchemaUpdate {
 
 func (m *SchemaUpdateManager) AppendSchema(extensions []schema.Update) {
 	m.updates[updateExternal] = extensions
+}
+
+func updateFromV2(ctx context.Context, tx *sql.Tx) error {
+	stmt := `
+CREATE TABLE internal_cluster_members_new (
+    id                      INTEGER   PRIMARY  KEY    AUTOINCREMENT  NOT  NULL,
+    name                    TEXT      NOT      NULL,
+    address                 TEXT      NOT      NULL,
+    certificate             TEXT      NOT      NULL,
+    schema_internal         INTEGER   NOT      NULL,
+  	schema_external         INTEGER   NOT      NULL,
+    heartbeat               DATETIME  NOT      NULL,
+    role                    TEXT      NOT      NULL,
+    api_extensions          TEXT      NOT      NULL DEFAULT '[]',
+    UNIQUE(name),
+    UNIQUE(certificate)
+);
+
+INSERT INTO internal_cluster_members_new (id, name, address, certificate, schema_internal, schema_external, heartbeat, role, api_extensions)
+SELECT id, name, address, certificate, schema_internal, schema_external, heartbeat, role, '[]' FROM internal_cluster_members;
+
+DROP TABLE internal_cluster_members;
+ALTER TABLE internal_cluster_members_new RENAME TO internal_cluster_members;
+`
+	_, err := tx.ExecContext(ctx, stmt)
+	return err
 }
 
 // updateFromV1 fixes a bug in the schemas table. Previously there was no way to tell when an update was internal, so the last external update would be re-run instead.
--- a/internal/extensions/extensions.go
+++ b/internal/extensions/extensions.go
@@ -1,0 +1,195 @@
+package extensions
+
+import (
+	"database/sql/driver"
+	"encoding/json"
+	"fmt"
+	"regexp"
+
+	"github.com/canonical/lxd/shared"
+)
+
+var internalExtensionRegex = regexp.MustCompile(`^internal:([a-z0-9]|[a-z0-9]_[a-z0-9])+$`)
+var externalExtensionRegex = regexp.MustCompile(`^([a-z0-9]|[a-z0-9]_[a-z0-9])+$`)
+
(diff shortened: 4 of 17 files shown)


masnax (changes requested):
Thanks for this, just a couple things to change:

Firstly, we can't rely on the database to store the schema extensions for two reasons: 
 * If there's an API extension mismatch, we we would have to shut down the database which is expensive and slow.
 * More importantly, when joining a cluster, you don't have an open database yet, so when you set up the join operation you can't tell the existing cluster what API extensions you're able to support.

Instead I think we should follow an approach similar to LXD where we have a `[]string` of API extensions defined. In microcluster's case, this should be a function rather than a global variable, like we have for schema extensions here: https://github.com/canonical/microcluster/blob/7d8d3f4c1e7b178eaa13be1aadf513de57f9bc88/internal/db/update/update.go#L39-L45 
If the extensions are stored in memory, they will update immediately when we restart the daemon on an update, or start it for the first time.

Additionally, I think we should expose the concept of `vendor` and just have one big set of extensions, and error out if there's a conflict. Internally, we should have 2 lists for `internal` and `external` but when a user consumes from the API we should show them just 1 unified list.


FInally, to handle the API extensions on the joining side, we need to pass the lists with rest of the joining information. When we ask to join an existing cluster, we send that cluster a `JoinConfig` payload. We should include the list of schema extensions so the existing cluster can see if we are compatible, and only then allow us to open the database: https://github.com/canonical/microcluster/blob/7d8d3f4c1e7b178eaa13be1aadf513de57f9bc88/internal/rest/resources/control.go#L73-L82

The existing cluster should then record the size  in the `internal_cluster_members` table. Here I think we should have 2 new columns for `internal_api_extensions` and `external_api_extensions` which store the size of the respective extension sets, which will represent the "api version". We already do this for the schema version: https://github.com/canonical/microcluster/blob/7d8d3f4c1e7b178eaa13be1aadf513de57f9bc88/internal/rest/resources/cluster.go#L132-L140

tomponline:
> Additionally, I think we should not expose the concept of `vendor` and just have one big set of extensions, and error out if there's a conflict.

Agreed.

gabrielmougard:
@masnax @tomponline let me know what you think about this new take. It's interesting to see this PR with https://github.com/canonical/microovn/pull/121 as it shows how an integration can be easily done with a MicroCluster-based service.

masnax (changes requested):
Looking good, just one more nit about the naming structure and one last point to add:

What happens on an upgrade? If you have a cluster of microovns currently sitting at 10 API extensions, and you `snap refresh` one of them so it has 11 API extensions, how should the cluster behave?

It looks like right now you're registering the API extensions on start, and comparing them to other nodes when a join occurs. That means the cluster will ignore extension changes when a node is upgraded.  So we also need to compare extensions if we've restarted our node (which would happen on `snap refresh`). 

We do this for schema versions here: https://github.com/canonical/microcluster/blob/24e407f22615c00f3bb84e074f4c1c4d9807b647/internal/db/db.go#L23-L111

How this works is, if we open the database for any reason other than a bootstrap (so a join or a restart), we compare the schema version in the database to what we have registered locally. First, we update our node's version in `internal_cluster_members` so other nodes can compare themselves to us as well. The comparison works like this:
* If our locally registered schema version matches the all other nodes' versions recorded in `internal_cluster_members` then we open the database and proceed as normal.
* If our locally registered schema version is smaller than another node, that means we need to upgrade locally so we error out.
* If our locally registered schema version is larger than another node, that means we need to wait until the other nodes upgrade. So we block on `db.upgradeCh`: https://github.com/canonical/microcluster/blob/24e407f22615c00f3bb84e074f4c1c4d9807b647/internal/db/db.go#L86-L94 This channel closes when we make this `PATCH` request to dqlite: https://github.com/canonical/microcluster/blob/24e407f22615c00f3bb84e074f4c1c4d9807b647/internal/daemon/daemon.go#L450-L464 Any node that successfully opens its database (detects that its version matches all other records in `internal_cluster_members`) will make the PATCH requests to all nodes excluding itself, unblocking them.

So what I think you'll need to do here is modify `UpdateClusterMemberSchemaVersion` and `GetClusterMemberSchemaVersion` to return both the schema and internal and external API versions, and make sure all 3 are checked and updated in the `checkVersions` function inside `db.Open`.

tomponline:
@masnax does microcloud request the other members do a snap refresh when one gets upgraded like LXD does?

masnax:
> @masnax does microcloud request the other members do a snap refresh when one gets upgraded like LXD does?

I didn't know about this, sounds interesting.

tomponline:
> > @masnax does microcloud request the other members do a snap refresh when one gets upgraded like LXD does?
> 
> I didn't know about this, sounds interesting.

I think its this:

https://github.com/search?q=repo%3Acanonical%2Flxd%20LXD_CLUSTER_UPDATE&type=code
https://github.com/canonical/lxd-pkg-snap/blob/e896009ab5a74aa6ca9372d220015cbbf9919d7f/snapcraft/commands/daemon.start#L27

sabaini:
Thanks, this looks useful! As a general comment, it would be great to have the example app updated to make use of the extension by way of documentation

internal/db/extensions.go
```diff
@@ -0,0 +1,195 @@
...
+
+// Extensions represents a registry of extensions.
+type Extensions struct {
+	// internal extensions are related to MicroCluster capabilities, not capabilities of
+	// MicroCluster-backed services.
+	//
+	// To distinguish between internal and external extensions, internal extensions
+	// are not prefixed with a `<vendor>:` value.
+	// e.g, `runtime_extension_v1`, `runtime_extension_v2`, ...
+	internal map[string]struct{}
```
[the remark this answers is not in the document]
gabrielmougard: I use the `internal` field as a fast lookup datastructure so that's why I use a `map[string]struct{}`

internal/db/extensions.go
```diff
@@ -0,0 +1,195 @@
...
+		if !internalExtensionRegex.MatchString(extension) {
+			return &invalidExternalExtensionErr{}
+		}
+
+		return nil
+	}
+
+	// Aside from the internal extensions, we need the extension to be like
+	// `<vendor>:<extension>` with only lowercase letters and underscores.
+	if !externalExtensionRegex.MatchString(extension) {
```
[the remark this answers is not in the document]
gabrielmougard: fixed

masnax (changes requested):
Just a few more things and I think we should be good to go :)

internal/extensions/extensions.go
```diff
@@ -0,0 +1,233 @@
...
+	// MicroCluster-backed services.
+	//
+	// To distinguish between internal and external extensions, internal extensions
+	// are prefixed with a `internal:` value.
+	// e.g, `internal:runtime_extension_v1`, `internal:runtime_extension_v2`, ...
+	internal map[string]struct{}
+
+	// External extensions are related to capabilities of MicroCluster-backed services.
+	// e.g, `microovn_custom_encapsulation_ip`
+	External map[string]struct{}
```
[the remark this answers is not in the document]
gabrielmougard: fixed

internal/db/db.go
```diff
@@ -39,24 +40,73 @@ func (db *DB) Open(bootstrap bool, project string) error {
...
+			versioningInfo, err := cluster.GetClusterMemberSchemaVersionsAndAPIExtensions(ctx, tx)
 			if err != nil {
 				return fmt.Errorf("Failed to get other members' schema versions: %w", err)
 			}
 
-			for _, version := range versions {
-				if schemaVersion == version {
-					// Versions are equal, there's hope for the
+			for _, versionInfo := range versioningInfo {
+				lInternal, remLInternal := extensions.IsSubset(internalExtensionsDB, versionInfo.InternalAPIExtensions) // this node's internal extensions are a subset of the other node's internal extensions
```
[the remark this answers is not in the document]
gabrielmougard: I like this approach! But in the case of two nodes having the same number of extensions but not exactly the same ones, comparing the length of the extensions would not work right (maybe that this scenario will never happen in practice..)? That's why I tried to compare extension set rather than the number of extensions.. It might be overengineered. But let's keep things simple: I'll do as you suggested ;)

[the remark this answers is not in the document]
gabrielmougard: fixed

gabrielmougard:
@masnax I have an open question: it'd be nice to integrate the testing of this extension system in, let say, the MicroCloud integration test suite. Is it the best place for testing such thing? I was thinking about these test scenarios:

## Bootstrap
* Bootstrap a cluster (all nodes have the same extensions)

## Bootstrap +  Join
* Bootstrap a cluster (all nodes have the same extensions) + Join a node (with the same extensions)
* Bootstrap a cluster (all nodes have the same extensions) + Join a node (without extensions system)
* Bootstrap a cluster (all nodes have the same extensions) + Join a node (with extensions but different from the bootstrapped nodes)

## Boostrap +  Upgrade
* Bootstrap a cluster (all nodes have no extension system) + Upgrade a node (with the latest extensions)
* Bootstrap a cluster (all nodes have an extension system but these are not the latest) + Upgrade a node (with the latest extensions)

____

If we were to do that kind of testing, how would the setup be done in the test suite? Will we have to manually build different versions of MicroCloud with snapcraft prior to launching the test suite to be able to create these heterogeneous cluster? This approach seems quite complex and very time consuming... I wondered if you had a better idea.

gabrielmougard:
@masnax I managed to bootstrap an initial cluster member through MicroOVN:
```bash
$ microovn cluster bootstrap
$ microovn cluster sql "SELECT * FROM internal_cluster_members;"
+----+-------------+-------------------+------------------------------------------------------------------+--------+-------------------------------------+-------+-------------------------------+-------------------------+
| id |    name     |      address      |                           certificate                            | schema |              heartbeat              | role  |    internal_api_extensions    | external_api_extensions |
+----+-------------+-------------------+------------------------------------------------------------------+--------+-------------------------------------+-------+-------------------------------+-------------------------+
| 1  | gab-desktop | 192.168.1.38:6443 | -----BEGIN CERTIFICATE-----                                      | 4      | 2024-02-27T15:50:46.148257347+01:00 | voter | internal:runtime_extension_v1 | custom_encapsulation_ip |
|    |             |                   | MIIB8jCCAXigAwIBAgIQHZLPsKL7yExQMe4nsh6jcTAKBggqhkjOPQQDAzApMQww |        |                                     |       |                               |                         |
|    |             |                   | CgYDVQQKEwNMWEQxGTAXBgNVBAMMEHJvb3RAZ2FiLWRlc2t0b3AwHhcNMjQwMjI3 |        |                                     |       |                               |                         |
|    |             |                   | MTQ1MDE0WhcNMzQwMjI0MTQ1MDE0WjApMQwwCgYDVQQKEwNMWEQxGTAXBgNVBAMM |        |                                     |       |                               |                         |
|    |             |                   | EHJvb3RAZ2FiLWRlc2t0b3AwdjAQBgcqhkjOPQIBBgUrgQQAIgNiAASk9o+DEQ08 |        |                                     |       |                               |                         |
|    |             |                   | aRlr5AR+zgv6LOHaSb29T21tn4G18vnCF9sApZ/KmxHhM+5W4qHKACyJN2dHanz5 |        |                                     |       |                               |                         |
|    |             |                   | WW43hLQad6w3fnB8SFQ53uLiDKtZb68KeMkgWk3oDJxeBNDtHAXi98SjZTBjMA4G |        |                                     |       |                               |                         |
|    |             |                   | A1UdDwEB/wQEAwIFoDATBgNVHSUEDDAKBggrBgEFBQcDATAMBgNVHRMBAf8EAjAA |        |                                     |       |                               |                         |
|    |             |                   | MC4GA1UdEQQnMCWCC2dhYi1kZXNrdG9whwR/AAABhxAAAAAAAAAAAAAAAAAAAAAB |        |                                     |       |                               |                         |
|    |             |                   | MAoGCCqGSM49BAMDA2gAMGUCMQCMjiPVww/dPoRCiRbjQ5iE3axQaLadYtbWeVtx |        |                                     |       |                               |                         |
|    |             |                   | LnUfGtEND3rBioT4m+BRdvdkiR8CMDdiPfemuHl7vyJtE2YzhNXwYqUhrjZbcdDT |        |                                     |       |                               |                         |
|    |             |                   | TCO+zbs9KVSnJ/h8J6udd8sqJEnjVw==                                 |        |                                     |       |                               |                         |
|    |             |                   | -----END CERTIFICATE-----                                        |        |                                     |       |                               |                         |
|    |             |                   |                                                                  |        |                                     |       |                               |                         |
+----+-------------+-------------------+------------------------------------------------------------------+--------+-------------------------------------+-------+-------------------------------+-------------------------+

```

masnax:
>     * [ ]  Bootstrap a cluster (all nodes have no extension system) + Upgrade a node (with the latest extensions)
>       @masnax I have a `Unable to start daemon: Daemon failed to start: Failed to re-establish cluster connection: Failed to update schema version when joining cluster: no such column: internal_api_extensions`. I thought that during a `snap refresh` (or in my case a `snap install <local_snap_file>.snap --dangerous`), the DB schema was re-applied. How to handle that case properly?

So what you're doing here is creating the `checkVersion` function, which is passed to `newSchema.Check` here: https://github.com/canonical/microcluster/pull/86/files#diff-7e4ba9264afd9eae1dce0c185a9141edf9c02bb44b0b76d19877389b71624006L77

And this basically just assigns the function to a field in `newSchema`: https://github.com/canonical/microcluster/blob/24e407f22615c00f3bb84e074f4c1c4d9807b647/internal/db/update/schema.go#L33-L38

Finally, when we call `newSchema.Ensure` here: https://github.com/canonical/microcluster/pull/86/files#diff-7e4ba9264afd9eae1dce0c185a9141edf9c02bb44b0b76d19877389b71624006R104-R107

But `Ensure` runs `checkVersion` before we apply the schema updates: https://github.com/canonical/microcluster/blob/24e407f22615c00f3bb84e074f4c1c4d9807b647/internal/db/update/schema.go#L82-L93

The schema updates come afterward, here: https://github.com/canonical/microcluster/blob/24e407f22615c00f3bb84e074f4c1c4d9807b647/internal/db/update/schema.go#L95-L107


This is an unfortunate consequence of changing the schema update logic. In this case we need to update the relevant tables before do any checks. Sadly, I think we'll need to add a special case before we run `checkVersion`, rather than relying on an `updateFromV{n}`functions.

masnax:
So in this case you'll have to grab the `updateFromV1` function and call it before running `checkVersion`. 

Looking at this function, it seems like it can break if it runs more than once, so we should probably change it to create a second table, then move the rows over from the old table, then delete the old table and rename the new table.

We occasionally do this in LXD like this: https://github.com/canonical/lxd/blob/898d3e02ff6b190e199d2dda4324cb0248a987c5/lxd/db/cluster/update.go#L221-L244

masnax:
>     * [ ]  **[WIP]** Bootstrap a cluster (all nodes have an extension system but these are not the latest) + Upgrade a node (with the latest extensions) (@masnax how can you refresh a snap with a new `.snap` file ? We can't use `snap refresh` right?)

You can't use a new `.snap` file from my understanding, but you can leverage `microcloudd.debug` as a debug binary:

``bash
# Install the snap.
snap install --dangerous ./microcloud.snap

# Make a change to the codebase and build
make

# Copy the binary to the snap state directory (don't forget the double-d for the daemon binary).
cp "${GOPATH}/bin/microcloudd" /var/snap/microcloud/common/microcloudd.debug
systemctl restart snap.microcloud.daemon
```

**Comment by masnax** [comment]
> > ```
> > * [ ]  **[WIP]** Bootstrap a cluster (all nodes have an extension system but these are not the latest) + Upgrade a node (with the latest extensions) (@masnax how can you refresh a snap with a new `.snap` file ? We can't use `snap refresh` right?)
> > ```
> 
> You can't use a new `.snap` file from my understanding, but you can leverage `microcloudd.debug` as a debug binary:
> 
> ```shell
> # Install the snap.
> snap install --dangerous ./microcloud.snap
> 
> # Make a change to the codebase and build
> make
> 
> # Copy the binary to the snap state directory (don't forget the double-d for the daemon binary).
> cp "${GOPATH}/bin/microcloudd" /var/snap/microcloud/common/microcloudd.debug
> systemctl restart snap.microcloud.daemon
> ```

This should be a close emulation of `snap refresh`.

**Comment by gabrielmougard** [comment]
@masnax I managed to solve the upgrade issue but there's an other problem that I think could be related to https://github.com/canonical/microcluster/issues/66

- Bootstraping a new cluster works fine.
- Then there is the case of a cluster update. In this case, we have two sub-cases:
  - "Update a cluster with old extensions; there is no structural changes in the DB". This case works fine (at the beginning, after the update, the member's daemon is marked as not initialized but after a while, when all the nodes are updated, they come back online)
  - "Update a 'legacy' cluster; understand a cluster with no API extension system. A structural DB change is necessary" . I can actually update the DB structure in the good way: Here is an example after updating the first node `m1`:

```
root@m2:~# microovn cluster sql "SELECT * FROM internal_cluster_members;"
+----+------+--------------------+------------------------------------------------------------------+--------+--------------------------------+-------+-------------------------------+-------------------------+
| id | name |      address       |                           certificate                            | schema |           heartbeat            | role  |    internal_api_extensions    | external_api_extensions |
+----+------+--------------------+------------------------------------------------------------------+--------+--------------------------------+-------+-------------------------------+-------------------------+
| 1  | m1   | 10.12.134.172:6443 | -----BEGIN CERTIFICATE-----                                      | 4      | 2024-03-08T12:34:44.568361958Z | voter | internal:runtime_extension_v1 | custom_encapsulation_ip |
|    |      |                    | MIIB2DCCAV2gAwIBAgIQbxazH6ZaWJqaY3w+tsIigjAKBggqhkjOPQQDAzAgMQww |        |                                |       |                               |                         |
|    |      |                    | CgYDVQQKEwNMWEQxEDAOBgNVBAMMB3Jvb3RAbTEwHhcNMjQwMzA4MTIzMzM2WhcN |        |                                |       |                               |                         |
|    |      |                    | MzQwMzA2MTIzMzM2WjAgMQwwCgYDVQQKEwNMWEQxEDAOBgNVBAMMB3Jvb3RAbTEw |        |                                |       |                               |                         |
|    |      |                    | djAQBgcqhkjOPQIBBgUrgQQAIgNiAARBRNvAMLrTcyCOCnsqyI24J5rpF9JoIdMr |        |                                |       |                               |                         |
|    |      |                    | 7cWq4ExoEOfIUcIc0mbhTPOw7w3PP8fnr7dVpArXs97WOIbu0WFH4E1BAsGOM1VO |        |                                |       |                               |                         |
|    |      |                    | 4pc0kD9duvmr1Wr1cBqscGNHId5dTH6jXDBaMA4GA1UdDwEB/wQEAwIFoDATBgNV |        |                                |       |                               |                         |
|    |      |                    | HSUEDDAKBggrBgEFBQcDATAMBgNVHRMBAf8EAjAAMCUGA1UdEQQeMByCAm0xhwR/ |        |                                |       |                               |                         |
|    |      |                    | AAABhxAAAAAAAAAAAAAAAAAAAAABMAoGCCqGSM49BAMDA2kAMGYCMQDyIuXk6cmq |        |                                |       |                               |                         |
|    |      |                    | 5u0sdA7BBVxFbMtWPHF/7hxAmxmsO/RW/C6vA/NU7B4wRAtGGqmA5CUCMQCiM0/e |        |                                |       |                               |                         |
|    |      |                    | 1rQ1CqlhUBuNvJgvQpZIHdGbgga0InPO3AVoOe7TJHE2BP7CtDojI9UELKk=     |        |                                |       |                               |                         |
|    |      |                    | -----END CERTIFICATE-----                                        |        |                                |       |                               |                         |
|    |      |                    |                                                                  |        |                                |       |                               |                         |
| 2  | m2   | 10.12.134.224:6443 | -----BEGIN CERTIFICATE-----                                      | 3      | 2024-03-08T12:34:44.633496063Z | voter | <nil>                         | <nil>                   |
|    |      |                    | MIIB2DCCAV6gAwIBAgIRALkHSZPxE2TAS9YDMo3XWFswCgYIKoZIzj0EAwMwIDEM |        |                                |       |                               |                         |
|    |      |                    | MAoGA1UEChMDTFhEMRAwDgYDVQQDDAdyb290QG0yMB4XDTI0MDMwODEyMzI1NFoX |        |                                |       |                               |                         |
|    |      |                    | DTM0MDMwNjEyMzI1NFowIDEMMAoGA1UEChMDTFhEMRAwDgYDVQQDDAdyb290QG0y |        |                                |       |                               |                         |
|    |      |                    | MHYwEAYHKoZIzj0CAQYFK4EEACIDYgAENyus3L/JQZRFX/x3Qjak3Lk3KH+6jU22 |        |                                |       |                               |                         |
|    |      |                    | G9qi9fR7oD5tqBMbBTEtZTvFVOZvKCu9s9VsCukKufkaSb7si1zZ1dUQm/vBpjt+ |        |                                |       |                               |                         |
|    |      |                    | YYkkrQkmKEknyEvGpTt4nImXKZKW6EGNo1wwWjAOBgNVHQ8BAf8EBAMCBaAwEwYD |        |                                |       |                               |                         |
|    |      |                    | VR0lBAwwCgYIKwYBBQUHAwEwDAYDVR0TAQH/BAIwADAlBgNVHREEHjAcggJtMocE |        |                                |       |                               |                         |
|    |      |                    | fwAAAYcQAAAAAAAAAAAAAAAAAAAAATAKBggqhkjOPQQDAwNoADBlAjBcaJoE0Rk3 |        |                                |       |                               |                         |
|    |      |                    | MxwwKmatV5ySRtlizx5QrQ/sEQ3yVKwLYW5klq2yCCT8HG+vFgTnPIsCMQDDj1E8 |        |                                |       |                               |                         |
|    |      |                    | je2VNPkEQLO3lIV8ilSO7uyBn529DBDrsuE89bYqZDNtt9TNzPs/WBkLjUQ=     |        |                                |       |                               |                         |
|    |      |                    | -----END CERTIFICATE-----                                        |        |                                |       |                               |                         |
|    |      |                    |                                                                  |        |                                |       |                               |                         |
| 3  | m3   | 10.12.134.18:6443  | -----BEGIN CERTIFICATE-----                                      | 3      | 2024-03-08T12:34:44.626596573Z | voter | <nil>                         | <nil>                   |
|    |      |                    | MIIB1zCCAV2gAwIBAgIQQ5TH0aGm5wnKFMaGKiJZkTAKBggqhkjOPQQDAzAgMQww |        |                                |       |                               |                         |
|    |      |                    | CgYDVQQKEwNMWEQxEDAOBgNVBAMMB3Jvb3RAbTMwHhcNMjQwMzA4MTIzMzA2WhcN |        |                                |       |                               |                         |
|    |      |                    | MzQwMzA2MTIzMzA2WjAgMQwwCgYDVQQKEwNMWEQxEDAOBgNVBAMMB3Jvb3RAbTMw |        |                                |       |                               |                         |
|    |      |                    | djAQBgcqhkjOPQIBBgUrgQQAIgNiAAR0vfVuu5CQ2kZX3LpU0Lgp3gwXbKIsLPOg |        |                                |       |                               |                         |
|    |      |                    | aq2rfGg1IL6GbePHhwWEwLOiB3/OreKQLJJZbM0ohScl9oUi6oEJqyoRtCb/KbYX |        |                                |       |                               |                         |
|    |      |                    | v4CoPwQce4qgzdDOb7/7gjC8ZZwq6hajXDBaMA4GA1UdDwEB/wQEAwIFoDATBgNV |        |                                |       |                               |                         |
|    |      |                    | HSUEDDAKBggrBgEFBQcDATAMBgNVHRMBAf8EAjAAMCUGA1UdEQQeMByCAm0zhwR/ |        |                                |       |                               |                         |
|    |      |                    | AAABhxAAAAAAAAAAAAAAAAAAAAABMAoGCCqGSM49BAMDA2gAMGUCMQC8qd2i/FDc |        |                                |       |                               |                         |
|    |      |                    | wEwOi5folOEsX5B+2nOTR00hzgSTB2dGU3P7veStAEIjqAHKSDRCCJcCMEzmzSTu |        |                                |       |                               |                         |
|    |      |                    | jLEO+qKqTITb9WyaQWH1NBnz6T7CUfUl4c0DRtabUpgypjgr1IJv8moTDQ==     |        |                                |       |                               |                         |
|    |      |                    | -----END CERTIFICATE-----                                        |        |                                |       |                               |                         |
|    |      |                    |                                                                  |        |                                |       |                               |                         |
+----+------+--------------------+------------------------------------------------------------------+--------+--------------------------------+-------+-------------------------------+-------------------------+
```
  I eventually do that on all the nodes, but I end up having `Error: Daemon not yet initialized` even after waiting for a while. In the journalctl logs, I can see a `time="2024-03-08T11:41:15Z" level=error msg="Failed to send database upgrade request"`.
  
What do you think here is going wrong?

**Comment by masnax** [comment]
> I eventually do that on all the nodes, but I end up having `Error: Daemon not yet initialized` even after waiting for a while. In the journalctl logs, I can see a `level=error msg="Failed to send database upgrade request" error="Patch \"https://10.12.134.18:6443/cluster/internal/database\": Unable to connect to \"10.12.134.18:6443\": dial tcp 10.12.134.18:6443: connect: connection refused"`. Besides, all my `snap.microovn.*` services are active and running on all my nodes, so why am I seeing `connection refused` ? (thus leading to the fact that `NotifyUpgraded()` is never called, leaving the daemons on all the nodes to be uninitialized)
> 
> What do you think here is going wrong?
> 
> (I attached my journalctl logs of one of my node [microovn.log](https://github.com/canonical/microcluster/files/14539049/microovn.log))

I've tried running your branch locally with the example package and it works fine on my end (although I have to swap the example package schema extensions to `CREATE TABLE IF NOT EXISTS`)

Are you waiting long enough between starting microovn and restarting it? It takes about 30 seconds for the heartbeat to actually send the trust yamls to every cluster member, so that might be why you're getting `connection refused`. You can check the `internal_cluster_members` table to see if the nodes are reporing `PENDING` for their role first.

Coincidentally there's a PR up now (#89) which should fix that.

**Comment by masnax** [comment]
I think adding a whole other mechanism for the override is a bit overkill. You can just re-use the same update function directly, but add a check for existence of the columns to it, which will make it do nothing in those cases.

I think keeping track of a whole separate set of versions is going to get really confusing in the future. Especially with a version index included, as it's hard to tell what that's referencing.

Eventually we will stop supporting microcluster versions that don't have this quirk, and with both schema and API updates in place we won't need to mess with this block anymore, so I think it's fair to just call `updateFromV1` in two places, once before the main update block, and once after, but with a check that the columns don't already exist.

So how about something like this:
```go
// updateFromV1 adds API extension columns to the internal_cluster_members table.
// As these columns are required to check for other schema/API updates, we need to run it before other updates for existing deployments.
// So first we need to check if the columns were added earlier when the regular update logic runs.
func updateFromV1(ctx context.Context, tx *sql.Tx) error {
    stmt := `
SELECT count(name)
FROM pragma_table_info('internal_cluster_members')
WHERE name IN ('internal_api_extensions', 'external_api_extensions');
`

    var count int
    err := tx.QueryRow(stmt).Scan(&count)
    if err != nil && !errors.Is(err, sql.ErrNoRows) {
        return err
    }

    if err == nil && count != 2 {
        return fmt.Errorf("Invalid 'internal_cluster_members' schema")
    }

    // If we got an ErrNoRows error, then we still need to update the table.
    if err != nil {
        stmt := `
CREATE TABLE internal_cluster_members_new (
    id                      INTEGER   PRIMARY  KEY    AUTOINCREMENT  NOT  NULL,
    name                    TEXT      NOT      NULL,
    address                 TEXT      NOT      NULL,
    certificate             TEXT      NOT      NULL,
    schema                  INTEGER   NOT      NULL,
    heartbeat               DATETIME  NOT      NULL,
    role                    TEXT      NOT      NULL,
    internal_api_extensions TEXT,
    external_api_extensions TEXT,
    UNIQUE(name),
    UNIQUE(certificate)
);

INSERT INTO internal_cluster_members_new (id, name, address, certificate, schema, heartbeat, role, internal_api_extensions, external_api_extensions)
SELECT id, name, address, certificate, schema, heartbeat, role, NULL, NULL FROM internal_cluster_members;

DROP TABLE internal_cluster_members;
ALTER TABLE internal_cluster_members_new RENAME TO internal_cluster_members;
`
        _, err := tx.ExecContext(ctx, stmt)
        return err
    }

    return nil
}
```
Then we can add a helper that can export a particular version:
```go
// OverrideUpdate manually executes the update corresponding to the given version number, inclusive of external schema updates.
func (s *SchemaUpdate) OverrideUpdate(ctx context.Context, tx *sql.Tx, version int) error {
    if len(s.updates) < version {
        return fmt.Errorf("Invalid override version: %d", version)
    }

    return s.updates[version-1](ctx, tx)
}
```

And we can execute `newSchema.OverrideUpdate(ctx, tx, 2)` at the top of `checkVersion`. 

Additionally, keep in mind we always try to run all the updates so it's never okay to use `ADD COLUMN` as it will give an error the next time there's a schema update.

masnax:
> > I eventually do that on all the nodes, but I end up having `Error: Daemon not yet initialized` even after waiting for a while. In the journalctl logs, I can see a `level=error msg="Failed to send database upgrade request" error="Patch \"https://10.12.134.18:6443/cluster/internal/database\": Unable to connect to \"10.12.134.18:6443\": dial tcp 10.12.134.18:6443: connect: connection refused"`. Besides, all my `snap.microovn.*` services are active and running on all my nodes, so why am I seeing `connection refused` ? (thus leading to the fact that `NotifyUpgraded()` is never called, leaving the daemons on all the nodes to be uninitialized)
> > What do you think here is going wrong?
> > (I attached my journalctl logs of one of my node [microovn.log](https://github.com/canonical/microcluster/files/14539049/microovn.log))
> 
> I've tried running your branch locally with the example package and it works fine on my end (although I have to swap the example package schema extensions to `CREATE TABLE IF NOT EXISTS`)
> 
> Are you waiting long enough between starting microovn and restarting it? It takes about 30 seconds for the heartbeat to actually send the trust yamls to every cluster member, so that might be why you're getting `connection refused`. You can check the `internal_cluster_members` table to see if the nodes are reporing `PENDING` for their role first, or you can check that a cluster of size `n` has `n` yamls in its `var/snap/microovn/common/state/truststore` directory. 
> 
> Coincidentally there's a PR up now (#89) which should fix that.

masnax:
I made a mistake in the block above, I was using `ErrNoRows` but with `count` there will always be a row. I've fixed it now, in case you see the old one and get confused :)

masnax:
Looks like you've just found a bug with the schema update mechanism!

It seems we don't properly handle internal schema updates. If there are pre-existing external schema updates, and you add a new internal one, it tries to be smart and only run the newest update. However since the lists are appended together, it instead just re-runs the last external schema update. Not good at all.

This might be why you're struggling to get it to work with MicroOVN.

I've filed an issue for it here: https://github.com/canonical/microcluster/issues/91

markylaing:
@gabrielmougard is this blocked by #94?

gabrielmougard:
> @gabrielmougard is this blocked by #94?

yes

markylaing:
@gabrielmougard Let me know when you've rebased to fix the conflicts and I'll review :)

gabrielmougard:
@masnax can you give a feedback on this updated version of the PR? I made the existing unit tests pass but I'm not sure which tests we would need to create to fully test this API extension system... Thanks!

internal/db/update/schema.go
```diff
@@ -85,6 +85,13 @@ func (s *SchemaUpdate) Ensure(db *sql.DB) (int, error) {
 				return err
 			}
 
+			// updateFromV2 introduces the notion of internal and external APi extensions. It needs to be run before we check the API extensions version
+			// in the check hook.
+			err = updateFromV2(ctx, tx)
+			if err != nil {
+				return err
+			}
```
markylaing, line 93: I'm uneasy about this update being another special case, I thought the idea was that after #94 we'll have a slice of update functions to run, rather than specific ones called here. 

The check hook ought to fail if the schema versions are different before it checks for differences in API version. Then we will know that the update has been applied on all members. 

Also it looks like there's a misspelling of `API` here (lowercase `i`).

internal/extensions/extensions.go
```diff
@@ -0,0 +1,219 @@
...
+}
+
+// NewExtensionRegistry creates a new instance of the Extensions struct, which represents a registry of extensions.
+// We have the option to populate the internal extensions or not. Having s option becomes handy
+// when we want to create a custom registry of extensions (see `NewExtensionRegistryFromList`)
+func NewExtensionRegistry(populateInternal bool) (*Extensions, error) {
+	internal := make(map[string]struct{})
+	if populateInternal {
+		// Populate internal extensions here.
+		internal["internal:runtime_extension_v1"] = struct{}{}
```
markylaing, line 95: Can we move this to a separate list of strings as in LXD? Then iterate over them here.

internal/extensions/extensions.go
```diff
@@ -0,0 +1,219 @@
...
+	//
+	// To distinguish between internal and external extensions, internal extensions
+	// are prefixed with a `internal:` value.
+	// e.g, `internal:runtime_extension_v1`, `internal:runtime_extension_v2`, ...
+	internal map[string]struct{}
+
+	// External extensions are related to capabilities of MicroCluster-backed services.
+	// e.g, `microovn_custom_encapsulation_ip`
+	external map[string]struct{}
+}
```
markylaing, line 41: I think having a structure like this might be more than is required. In fact, it also loses information about the order that the extensions occurred (not sure if that matters). Since internal extensions are prefixed with `internal:`, can we just have `type Extensions []string`?

markylaing (changes requested):
Overall there are a few things to think about here. My main concern is adding what appears to be another special case in the schema update logic to include API extensions. Happy to have a call to discuss :)

internal/db/update/schema.go
masnax: So the schema/API update process looks like this, which might help describe the problem a bit:

* Open the database
* Fetch the schema and API versions of all cluster members to compare against our own.
* Update the row corresponding to ourselves in `internal_cluster_members` with the API and schema versions determined when the daemon started.
* Compare the updated schema/API versions to the rest of the cluster
* Potentially error out if we are behind
* By this point we are sure our daemon is valid, so actually run the schema updates that we are missing
* Wait for other nodes if their version is behind


We need the `internal_cluster_members` table to actually have all that information. So we're stuck in a loop where we need `updateFromV2` to have run before we check for any schema updates. 

That leaves us with two rather ugly possible solutions. Either we duplicate the whole block, to fully separate API and schema version checking, which means we need to have unique queries for schema and API updates and verification. Or instead we can manually call `updateFromV2` before we attempt to fetch from the `internal_cluster_members` table for the first time.

markylaing: I'm not clear on why the whole block needs to be duplicated to perform the API version check though.

Once we've passed the schema check, can we not query for each nodes API extensions and just check that the contents are identical?

masnax: It's definitely more resilient in case we ever need to make a schema update to the `internal_cluster_members` table again.

internal/db/db_test.go
```diff
@@ -40,115 +44,163 @@ func (s *dbSuite) Test_waitUpgrade() {
 	}{
 		{
 			name:              "No upgrade, no other nodes",
-			upgradedLocalInfo: versions{schemaInt: 0, schemaExt: 0},
+			upgradedLocalInfo: versions{schemaInt: 0, schemaExt: 0, apiExtensionInt: "internal:ext_a,internal:ext_b", apiExtensionExt: "ext_c,ext_d"},
```
masnax, line 47: I think having all these API updates in this test is a bit overkill. You've added (I think) the same extensions to each entry, but I think we should limit the test input to only the values that we want to change across the tests, otherwise it just makes it hard to read what the test is doing. If you need to set a default value, you can do that in the actual test loop below.

I think maybe you can keep this test the same and rename it to `waitUpgradeSchema`, and then add a new test case for `waitUpgradeAPI` which mainly changes the `apiExtensionsInt` and `apiExtensionsExt` in all the same cases as this test. That might make it easier to follow what each test is doing as there's not so many moving variables.

We can also have a 3rd test function that checks if both API and Schema versions are being updated at the same time.

internal/db/update/schema.go
markylaing: Yeah that's my worry. We have an established system now for schema updates and while it's a little repetitive to perform some of the same steps for the API extensions it's probably better to do this check separately IMO

internal/extensions/extensions.go
```diff
@@ -0,0 +1,219 @@
...
+	return fmt.Sprintf("Internal extension format is invalid (%v). It should be in the format `internal:<extension>` with only lowercase letters and underscores", e.invalidExt)
+}
+
+type invalidExternalExtensionErr struct {
+	invalidExt string
+}
+
+func (e *invalidExternalExtensionErr) Error() string {
+	return fmt.Sprintf("External extension format is invalid (%v). It should be in the format `<extension>` with only lowercase letters and underscores", e.invalidExt)
+}
```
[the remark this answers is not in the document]
gabrielmougard: fixed

internal/extensions/extensions.go
gabrielmougard: fixed

internal/extensions/extensions.go
```diff
@@ -0,0 +1,219 @@
...
+	for extension := range e.internal {
+		internalKeys = append(internalKeys, extension)
+	}
+
+	for extension := range e.external {
+		externalKeys = append(externalKeys, extension)
+	}
+
+	return strings.Join(internalKeys, ","), strings.Join(externalKeys, ",")
+}
```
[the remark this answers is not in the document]
gabrielmougard: fixed

internal/extensions/extensions.go
```diff
@@ -0,0 +1,219 @@
...
+	for extension := range e.internal {
+		extensions = append(extensions, extension)
+	}
+
+	for extension := range e.external {
+		extensions = append(extensions, extension)
+	}
+
+	return extensions
+}
```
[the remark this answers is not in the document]
gabrielmougard: fixed

internal/extensions/extensions.go
gabrielmougard: fixed

internal/db/update/update.go
```diff
@@ -61,6 +62,50 @@ func (m *SchemaUpdateManager) AppendSchema(extensions []schema.Update) {
...
+    UNIQUE(certificate)
+);
+
+INSERT INTO internal_cluster_members_new (id, name, address, certificate, schema_internal, schema_external, heartbeat, role, api_extensions_internal, api_extensions_external)
+SELECT id, name, address, certificate, schema_internal, schema_external, heartbeat, role, NULL, NULL FROM internal_cluster_members;
+
+DROP TABLE internal_cluster_members;
+ALTER TABLE internal_cluster_members_new RENAME TO internal_cluster_members;
+`
+		_, err := tx.ExecContext(ctx, stmt)
```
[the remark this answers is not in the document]
gabrielmougard: fixed

internal/db/db_test.go
gabrielmougard: fixed.

internal/db/update/schema.go
gabrielmougard: fixed

gabrielmougard:
@markylaing @masnax updated.

Also, @masnax, In my case in order to run the `db_test.go` unit tests, I have to edit the following for it to work:

```patch
diff --git a/internal/db/db_test.go b/internal/db/db_test.go
index 9858f4b..9aab1bc 100644
--- a/internal/db/db_test.go
+++ b/internal/db/db_test.go
@@ -661,7 +661,7 @@ func NewTestDB(extensionsExternal []schema.Update) (*DB, error) {
 		return nil, err
 	}
 
-	err = cluster.PrepareStmts(db.db, cluster.GetCallerProject(), false)
+	err = cluster.PrepareStmts(db.db, "cluster", false)
 	if err != nil {
 		return nil, err
 	}
```

Else, `cluster.GetCallerProject()` translates to `"internal"` which is not the same project used by the prepared statements in the `cluster` package. I can't really figure out why the unit tests are passing on github without this change...

masnax:
> @markylaing @masnax updated.
> 
> Also, @masnax, In my case in order to run the `db_test.go` unit tests, I have to edit the following for it to work:
> 
> ```diff
> diff --git a/internal/db/db_test.go b/internal/db/db_test.go
> index 9858f4b..9aab1bc 100644
> --- a/internal/db/db_test.go
> +++ b/internal/db/db_test.go
> @@ -661,7 +661,7 @@ func NewTestDB(extensionsExternal []schema.Update) (*DB, error) {
>  		return nil, err
>  	}
>  
> -	err = cluster.PrepareStmts(db.db, cluster.GetCallerProject(), false)
> +	err = cluster.PrepareStmts(db.db, "cluster", false)
>  	if err != nil {
>  		return nil, err
>  	}
> ```
> 
> Else, `cluster.GetCallerProject()` translates to `"internal"` which is not the same project used by the prepared statements in the `cluster` package. I can't really figure out why the unit tests are passing on github without this change...

So `PrepareStmts` will always prepare statements from `microcluster` regardless of the output of `cluster.GetCallerProject()`:
https://github.com/canonical/microcluster/blob/05982b6af9fe1d585aea6aff6749373f69b14c64/cluster/stmt.go#L49-L53

So it's very strange that you're running into issues. It's even stranger that you're getting `internal` as output. `GetCallerProject` will take the 3rd path after 'src' (so `microcloud` from `src/github.com/canonical/microcloud/....`): 
https://github.com/canonical/microcluster/blob/05982b6af9fe1d585aea6aff6749373f69b14c64/cluster/stmt.go#L114-L119

What's the absolute path to your copy of `microcluster`?

masnax:
@gabrielmougard needs a rebase now, sorry.

masnax:
Sorry about the milestone, that was my jumping on the keyboard :)

gabrielmougard:
> @gabrielmougard
> 
> So `PrepareStmts` will always prepare statements from `microcluster` regardless of the output of `cluster.GetCallerProject()`:
> 
> https://github.com/canonical/microcluster/blob/05982b6af9fe1d585aea6aff6749373f69b14c64/cluster/stmt.go#L49-L53
> 
> So it's very strange that you're running into issues. It's even stranger that you're getting `internal` as output. `GetCallerProject` will take the 3rd path after 'src' (so `microcluster` from `src/github.com/canonical/microcluster/....`):
> 
> https://github.com/canonical/microcluster/blob/05982b6af9fe1d585aea6aff6749373f69b14c64/cluster/stmt.go#L114-L119
> 
> What's the absolute path to your copy of `microcluster`?

It's `/home/gab/go/src/github.com/microcluster`

masnax:
> It's `/home/gab/go/src/github.com/microcluster` and my GOPATH is `/home/gab/go`

Ah yeah it's because you're missing the full path. The [gopath reference](https://go.dev/wiki/GOPATH) mentions the paths in src should be `github.com/y/z`:

> The go get command downloaded source code satisfying import "example.com/y/z" to $GOPATH/src/example.com/y/z.

internal/extensions/extensions.go
```diff
@@ -0,0 +1,238 @@
...
+				return nil, err
+			}
+
+			i := sort.Search(len(*extensions), func(i int) bool {
+				return (*extensions)[i] >= internalExtension
+			})
+
+			*extensions = append(*extensions, internalExtension)
+			copy((*extensions)[i+1:], (*extensions)[i:])
+			(*extensions)[i] = internalExtension
```
masnax, line 99: Why does this need to be so complex? 

Why can't we just do
```go
if populateInternal {
  # Instantiate extensions with a size of 0 but a capacity equal to the size of the internal extension list.
  extensions = make(Extensions, 0, len(*internalExtensions))
  
  for _, internalExtension := range *internalExtensions {
      extensions = append(extensions, internalExtension)
  }
}

return extensions
```

internal/extensions/extensions.go
```diff
@@ -0,0 +1,238 @@
...
+
+			return !currentIsInternal && (*registry)[i] >= extension
+		})
+
+		*registry = append(*registry, extension)
+		copy((*registry)[i+1:], (*registry)[i:])
+		(*registry)[i] = extension
+	}
+
+	return registry, nil
```
masnax, line 147: Same here, it's not clear to me why we need to have all of this, and can't just directly append to `registry` and return.

internal/extensions/extensions.go
```diff
@@ -0,0 +1,238 @@
...
+		}
+
+		i := sort.Search(len(*e), func(i int) bool {
+			currentIsInternal := validateInternalExtension((*e)[i]) == nil
+			return !currentIsInternal && (*e)[i] >= extension
+		})
+
+		*e = append(*e, extension)
+		copy((*e)[i+1:], (*e)[i:])
+		(*e)[i] = extension
```
masnax, line 170: Here as well, in relation to appending.

internal/extensions/extensions.go
gabrielmougard: I guess, what I wanted to do here, is to insert the extension in the registry at the right index so that these are always sorted (sorted internal extensions and then sorted external extensions):
`internal:a, internal:b, .., internal_e, not_internal_a, not_internal_b`

If we just do:

```go
for _, internalExtension := range *internalExtensions {
      // validate first
      // ... 
      extensions = append(extensions, internalExtension)
}
```
if `*internalExtensions` is not sorted, this approach will keep `extensions` not sorted. The idea of inserting the entry in a sorted way, is to keep the data predictable in the database and when comparing two extensions (instead of always calling the sort function on the registry before inserting it to the DB or when comparing the entries with an other extensions)

internal/extensions/extensions.go
gabrielmougard: same reason as above

internal/extensions/extensions.go
gabrielmougard: same reason as above

internal/extensions/extensions.go
masnax: Sorry, I'm not following you about the sorting. You're sorting alphabetically by name of the extension, but if we're sorting at all, I would think it should be by recency of the API extension. Is there a benefit to sorting alphanumerically?

* `internalExtensions` is a list of arbitrary strings that begin with `internal:`, the list should be ordered by when the api extension was introduced by default, as we should always be appending to the end of the list when we add a new one.

* `extensions` is being instantiated in this function so it's always empty, and by the end it should contain all the same items as `internalExtensions`.

Then you run `sort.Search` and compare `(*extensions)[i] >= internalExtension` but this would compare the actual length of the strings, but this breaks the order that the API extensions were introduced in, and instead sorts them alphanumerically. 

I'm not sure if it's such a good idea to lose that order, and it's not clear to me what we gain with alphanumeric sorting instead.

internal/extensions/extensions.go
masnax: So here, it looks like the lists are also sorted alphanumerically, but we still have a preference for `internal:` in the front, which means the full sort method we have is "alphanumeric, except for the prefix `internal:` which is front-loaded"

This makes it even more confusing to me why "alphanumeric" is helpful here.

roosterfish (changes requested):
Sorry for the late review, looks good to me, just a comment on the regex. 
But I agree with the others the extension sorting feels quite complex.

internal/extensions/extensions.go
```diff
@@ -0,0 +1,238 @@
...
+	"encoding/json"
+	"errors"
+	"fmt"
+	"regexp"
+	"sort"
+
+	"github.com/canonical/lxd/shared"
+)
+
+var internalExtensionRegex = regexp.MustCompile(`^internal:[a-z0-9][a-z_0-9]*[a-z0-9]$`)
```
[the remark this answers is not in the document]
gabrielmougard: fixed

internal/extensions/extensions.go
gabrielmougard: fixed

internal/extensions/extensions.go
gabrielmougard: fixed

internal/extensions/extensions.go
gabrielmougard: You're right, after thinking about it, it does not really bring anything to sort it alphanumerically. I'm removing this.

internal/extensions/extensions_test.go
```diff
@@ -0,0 +1,150 @@
...
+
+func TestValidateExternalExtension(t *testing.T) {
+	cases := []struct {
+		name      string
+		extension string
+		wantError bool
+	}{
+		{"Valid Extension", "valid_extension", false},
+		{"Empty Extension", "", true},
+		{"Invalid Characters", "Invalid-Extension!", true},
```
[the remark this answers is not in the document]
gabrielmougard: fixed

internal/extensions/extensions_test.go
```diff
@@ -0,0 +1,148 @@
...
+		{"Invalid Start Underscore", "_invalidextension", true},
+		{"Invalid End Underscore", "invalidextension_", true},
+		{"Invalid Non Alphanumeric", "invalid-extension", true},
+		{"Invalid Start Numeric", "123_invalid", false},
```
roosterfish, line 26: `wantError` is set to `false` here so should the test name say `Valid Start Numeric` instead?

gabrielmougard: yes! thanks for seeing this

gabrielmougard: fixed

gabrielmougard: fixed

masnax (approved):
LGTM now!

@gabrielmougard Have you checked that this works properly with existing & new deployments of micro* apps before we merge?

@roosterfish @markylaing I won't dismiss your reviews yet in case you have any further input.

roosterfish (approved):
Thanks for reducing the complexity on the `NewExtensionRegistry*()` functions and fixing the regex. Looks good to me!

tomponline:
@markylaing please can you approve and merge if you're happy with this.

internal/extensions/extensions.go
```diff
@@ -0,0 +1,213 @@
...
+// from API requests.
+func (e *Extensions) UnmarshalJSON(data []byte) error {
+	var tmp []string
+	if err := json.Unmarshal(data, &tmp); err != nil {
+		return err
+	}
+
+	*e = Extensions(tmp)
+	return nil
+}
```
markylaing, line 167: Now that `Extensions` is a `[]string` I'm not sure we need the `json.Marshaler` `json.Unmarshaler` implementations as I believe the `json` library will understand the underlying type.

I could be wrong though

internal/db/update/update.go
```diff
@@ -61,6 +62,32 @@ func (m *SchemaUpdateManager) AppendSchema(extensions []schema.Update) {
...
+CREATE TABLE internal_cluster_members_new (
+    id                      INTEGER   PRIMARY  KEY    AUTOINCREMENT  NOT  NULL,
+    name                    TEXT      NOT      NULL,
+    address                 TEXT      NOT      NULL,
+    certificate             TEXT      NOT      NULL,
+    schema_internal         INTEGER   NOT      NULL,
+  	schema_external         INTEGER   NOT      NULL,
+    heartbeat               DATETIME  NOT      NULL,
+    role                    TEXT      NOT      NULL,
+    api_extensions          TEXT      NOT      NULL DEFAULT '{}',
```
markylaing, line 76: Since extensions are a `[]string`, shouldn't the default value be `[]`?

internal/extensions/extensions.go
```diff
@@ -0,0 +1,213 @@
...
+	bytes, ok := value.([]byte)
+	if !ok {
+		return errors.New("type assertion to []byte failed, incompatible type")
+	}
+
+	return json.Unmarshal(bytes, e)
+}
+
+// IsSameVersion checks if the source registry supports the target registry.
+func (e *Extensions) IsSameVersion(t *Extensions) error {
```
markylaing, line 194: Since `Extensions` is a slice, it is already a pointer. I think most of these methods can have the value rather the reference as the receiver.

The `UnmarshalJSON` implementation would be an exception as the `json` library must be given a reference, but that method may or may not be needed (see above).

markylaing (changes requested):
Sorry it's taken a while to get to this.

In general I don't think we need to pass around a `*extensions.Extensions` as a slice is already a pointer.

The comment I've added about the schema update will need to be addressed before merging but other comments can be addressed in a follow up PR if you choose. Thanks.

internal/db/update/update.go
gabrielmougard: yes it should! I'm testing this with MicroOVN right now and I'm running into an unmarshalling issue when upgrading a legacy cluster. Thanks!

internal/extensions/extensions.go
gabrielmougard: fixed

internal/extensions/extensions.go
gabrielmougard: fixed

internal/db/db.go
```diff
@@ -121,6 +149,37 @@ func (db *DB) waitUpgrade(bootstrap bool) error {
...
+				if otherNodesBehindInternal {
+					otherNodesBehind = true
+					return schema.ErrGracefulAbort
+				}
+
+				return nil
+			})
+		}
+
 		return err
```
masnax, line 183: This was a sneaky error! If you return `schema.ErrGracefulAbort` from within the transaction, then the transaction will never be committed and when each node updates its local API version in the database, it won't be reflected on other nodes.

You can fix it by moving the check for `otherNodesBehind` outside the transaction.

```suggestion
		if !bootstrap {
		        otherNodesBehindAPI := false
			// Perform the API extensions check.
			err = query.Transaction(context.TODO(), db.db, func(ctx context.Context, tx *sql.Tx) error {
				err := cluster.UpdateClusterMemberAPIExtensions(tx, ext, db.listenAddr.URL.Host)
				if err != nil {
					return fmt.Errorf("Failed to update API extensions when joining cluster: %w", err)
				}

				clusterMembersAPIExtensions, err := cluster.GetClusterMemberAPIExtensions(ctx, tx)
				if err != nil {
					return fmt.Errorf("Failed to get other members' API extensions: %w", err)
				}

				otherNodesBehindAPI, err := checkAPIExtensions(ext, clusterMembersAPIExtensions)
				if err != nil {
					return err
				}

				return nil
			})
			if err != nil {
			  return err
			}
			
			if otherNodesBehindInternal {
				otherNodesBehind = true
				return schema.ErrGracefulAbort
		        }
		}

		return nil
```

masnax (changes requested):

internal/db/db.go
masnax: Also, I think we should halve the timeout here: https://github.com/canonical/microcluster/blob/94483f4d619243d598d4b5e512735035d1e34445/internal/db/db.go#L132

The reason for that is because if an API and schema update happen simultaneously, we'll have to wait up to 2 whole minutes for each node to pick up the change.

masnax: You also may run into the bug fixed by #117 so you may want to rebase off of that to test locally.

gabrielmougard: Amazing! thanks a lot!

gabrielmougard: fixed

masnax (approved):

masnax:
I've tested that this works with new clusters and pre-existing clusters, so we can finally merge this one :)

PR #88, opened 2024-03-05
include name in encoded token when using `ListJoinTokens()`

neoaggelos:
### Summary

When using `ListJoinTokens()` the tokens do not match the join token that was given by `NewJoinToken()`, as it does not include the token name in the encoded string.

Note that this does not affect joining nodes, as either of the tokens works (the `name` field of the token is not used).


roosterfish (approved):
Thanks, looks good to me. @masnax can you please double check?

masnax:
Thanks, good catch :)

PR #89, opened 2024-03-08
Ensure joining nodes are immediately trusted

masnax:
This cleans up some of the handling around authentication and trust when a system joins microcluster, or restarts.

Previously, we were relying on the dqlite heartbeat to carry trust information to existing nodes when a new node has joined the cluster. This meant that immediately after a node joins, the only systems that trust it are the one that handled its join request first, and potentially also the leader if that node wasn't the leader. 

To get around this, there is now a `PUT /cluster/1.0/cluster` method which forwards a joining node's truststore information to all other nodes. Since we have this in place now, we can also ensure that all nodes fully trust the new node before any of the `PreJoin`, `OnNewMember` and `PostJoin` hooks run.

For the reason above, the `POST /cluster/1.0/cluster` endpoint was actually untrusted by default, and authenticated with the join token. However, this endpoint is reused to send a cluster notification for the `OnNewMember` hook, which would bypass this authentication entirely as there's no join token. To fix that, the `authenticate` function is now exported, so that we can manage authentication a bit better, and make these endpoints require authentication except for the specific case of handling the join token.

Additionally cleans up some errors and makes the default `/cluster/1.0` endpoint untrusted.


masnax:
Might be a good idea to wait until #86 is merged, and then add an API extension for this, as it does affect the API.

tomponline:
> Might be a good idea to wait until #86 is merged, and then add an API extension for this, as it does affect the API.

Agreed good idea.

masnax:
> > PUT /cluster/1.0/cluster
> 
> This seems rather stammering - can we just use `PUT /cluster/1.0`?

`/cluster` represents pre-packaged microcluster endpoints

`/cluster/1.0` is publicly available endpoints provided by microcluster

`/cluster/1.0/cluster` is the subset of endpoints relating to management of the cluster itself. 

I'm not sure if making those top-level public endpoints is appropriate.

masnax:
> > > PUT /cluster/1.0/cluster
> > 
> > 
> > This seems rather stammering - can we just use `PUT /cluster/1.0`?
> 
> `/cluster` represents pre-packaged microcluster endpoints
> 
> `/cluster/1.0` is publicly available endpoints provided by microcluster. (there's also `/cluster/internal` and `cluster/control` for the fully internal and control socket endpoints respectively).
> 
> `/cluster/1.0/cluster` is the subset of endpoints relating to management of the cluster itself.
> 
> I'm not sure if making those top-level public endpoints is appropriate.

tomponline:
> top-level public endpoints

What does that mean, what are the implications you're thinking of?

PR #93, opened 2024-03-15
ensure daemon stop sequence runs with if context is cancelled

neoaggelos:
### Summary

Ensure `microcluster` shutdown sequence does not hang

### Changes

- Use a `sync.Once` to call the daemon.Stop() sequence
- Don't check if ShutdownCtx is already cancelled, that will fail the check if the parent context is cancelled by other means.
- Use a buffered channel to retrieve the result of the Stop(), otherwise the goroutine can block while sending at `ShutdownDoneCh`

### Details

Without the proposed changes, trying to stop a microcluster service often hangs with. This could be because there is a race condition: if the shutdownCtx is already cancelled, then the goroutine that calls `d.Stop()` will never be called.

```
Mar 15 08:05:24 t1 k8s.k8sd[1677]: + exec /snap/k8s/x1/bin/k8sd --port=6400 --state-dir=/var/snap/k8s/common/var/lib/k8sd/state
Mar 15 08:05:24 t1 k8s.k8sd[1677]: time="2024-03-15T08:05:24Z" level=warning msg="Failed to parse new remotes from truststore"
Mar 15 08:05:24 t1 k8s.k8sd[1677]: time="2024-03-15T08:05:24Z" level=info msg=" - binding control socket" socket=/var/snap/k8s/common/var/lib/k8sd/state/control.socket
Mar 15 08:05:24 t1 k8s.k8sd[1677]: time="2024-03-15T08:05:24Z" level=warning msg="microcluster database is uninitialized"
Mar 15 08:05:24 t1 k8s.k8sd[1677]: time="2024-03-15T08:05:24Z" level=warning msg="Failed to parse new remotes from truststore"
Mar 15 08:05:24 t1 k8s.k8sd[1677]: time="2024-03-15T08:05:24Z" level=info msg=" - binding https socket" network="[::]:6400"

<--- issued "snap stop k8s.k8sd"

Mar 15 08:05:36 t1 k8s.k8sd[1677]: time="2024-03-15T08:05:36Z" level=info msg="Closing filesystem watcher"
Mar 15 08:05:36 t1 k8s.k8sd[1677]: time="2024-03-15T08:05:36Z" level=info msg="Received signal" signal=terminated
Mar 15 08:05:36 t1 k8s.k8sd[1677]: time="2024-03-15T08:05:36Z" level=warning msg="Ignoring signal, shutdown already in progress" signal=terminated
Mar 15 08:05:36 t1 systemd[1]: Stopping Service for snap application k8s.k8sd...
Mar 15 08:06:06 t1 systemd[1]: snap.k8s.k8sd.service: State 'stop-sigterm' timed out. Killing.
Mar 15 08:06:06 t1 systemd[1]: snap.k8s.k8sd.service: Killing process 1677 (k8sd) with signal SIGKILL.
Mar 15 08:06:06 t1 systemd[1]: snap.k8s.k8sd.service: Killing process 1702 (k8sd) with signal SIGKILL.
Mar 15 08:06:06 t1 systemd[1]: snap.k8s.k8sd.service: Killing process 1703 (k8sd) with signal SIGKILL.
Mar 15 08:06:06 t1 systemd[1]: snap.k8s.k8sd.service: Killing process 1705 (k8sd) with signal SIGKILL.
Mar 15 08:06:06 t1 systemd[1]: snap.k8s.k8sd.service: Killing process 1706 (k8sd) with signal SIGKILL.
Mar 15 08:06:06 t1 systemd[1]: snap.k8s.k8sd.service: Killing process 1707 (k8sd) with signal SIGKILL.
Mar 15 08:06:06 t1 systemd[1]: snap.k8s.k8sd.service: Killing process 1708 (k8sd) with signal SIGKILL.
Mar 15 08:06:06 t1 systemd[1]: snap.k8s.k8sd.service: Killing process 1709 (k8sd) with signal SIGKILL.
Mar 15 08:06:06 t1 systemd[1]: snap.k8s.k8sd.service: Killing process 1710 (k8sd) with signal SIGKILL.
Mar 15 08:06:06 t1 systemd[1]: snap.k8s.k8sd.service: Killing process 1712 (k8sd) with signal SIGKILL.
Mar 15 08:06:06 t1 systemd[1]: snap.k8s.k8sd.service: Killing process 1713 (k8sd) with signal SIGKILL.
Mar 15 08:06:06 t1 systemd[1]: snap.k8s.k8sd.service: Killing process 1714 (n/a) with signal SIGKILL.
Mar 15 08:06:06 t1 systemd[1]: snap.k8s.k8sd.service: Killing process 1715 (n/a) with signal SIGKILL.
Mar 15 08:06:06 t1 systemd[1]: snap.k8s.k8sd.service: Main process exited, code=killed, status=9/KILL
Mar 15 08:06:06 t1 systemd[1]: snap.k8s.k8sd.service: Failed with result 'timeout'.
Mar 15 08:06:06 t1 systemd[1]: Stopped Service for snap application k8s.k8sd.
Mar 15 08:06:20 t1 systemd[1]: Started Service for snap application k8s.k8sd.
```

### With changes

Below are the shutdown logs with the applied changes (does not hang anymore):

```
Mar 15 08:09:03 t1 k8s.k8sd[3586]: time="2024-03-15T08:09:03Z" level=info msg="Preparing statements for Go project \"k8sd\""


<-- issued "snap stop k8s.k8sd"


Mar 15 08:10:06 t1 k8s.k8sd[3586]: time="2024-03-15T08:10:06Z" level=info msg="Received signal" signal=terminated
Mar 15 08:10:06 t1 k8s.k8sd[3586]: time="2024-03-15T08:10:06Z" level=warning msg="Ignoring signal, shutdown already in progress" signal=terminated
Mar 15 08:10:06 t1 k8s.k8sd[3586]: time="2024-03-15T08:10:06Z" level=info msg="Initiate daemon stop sequence"
Mar 15 08:10:06 t1 k8s.k8sd[3586]: time="2024-03-15T08:10:06Z" level=info msg="Closing filesystem watcher"
Mar 15 08:10:06 t1 k8s.k8sd[3586]: time="2024-03-15T08:10:06Z" level=error msg="Failed to initiate heartbeat round" address="10.0.0.47:6400" error="Post \"http://control.socket/cluster/internal/heartbeat\": context canceled"
Mar 15 08:10:06 t1 systemd[1]: Stopping Service for snap application k8s.k8sd...
Mar 15 08:10:06 t1 k8s.k8sd[3586]: time="2024-03-15T08:10:06Z" level=info msg="Stopping REST API handler - closing socket" socket=/var/snap/k8s/common/var/lib/k8sd/state/control.socket
Mar 15 08:10:06 t1 k8s.k8sd[3586]: time="2024-03-15T08:10:06Z" level=info msg="Received shutdown signal - aborting unix socket server startup"
Mar 15 08:10:06 t1 k8s.k8sd[3586]: time="2024-03-15T08:10:06Z" level=info msg="Stopping REST API handler - closing https socket" address="10.0.0.47:6400"
Mar 15 08:10:06 t1 k8s.k8sd[3586]: time="2024-03-15T08:10:06Z" level=info msg="Received shutdown signal - aborting https socket server startup"
Mar 15 08:10:06 t1 k8s.k8sd[3586]: time="2024-03-15T08:10:06Z" level=info msg="Daemon stopped"
Mar 15 08:10:06 t1 systemd[1]: snap.k8s.k8sd.service: Succeeded.
Mar 15 08:10:06 t1 systemd[1]: Stopped Service for snap application k8s.k8sd.
Mar 15 08:10:06 t1 systemd[1]: Started Service for snap application k8s.k8sd.
```

### DIFF
--- a/microcluster/app.go
+++ b/microcluster/app.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"sync"
 	"time"
 
 	"github.com/canonical/lxd/lxd/db/schema"
@@ -96,21 +97,30 @@ func (m *MicroCluster) Start(apiEndpoints []rest.Endpoint, schemaExtensions map[
 		return fmt.Errorf("Unable to start daemon: %w", err)
 	}
 
+	stop := sync.OnceFunc(func() {
+		logger.Info("Initiate daemon stop sequence")
+		err := d.ShutdownCtx.Err()
+		if stopErr := d.Stop(); stopErr != nil {
+			err = stopErr
+		}
+
+		d.ShutdownDoneCh <- err
+	})
+
 	for {
 		select {
+		case err = <-d.ShutdownDoneCh:
+			close(d.ShutdownDoneCh)
+			return err
+		case <-d.ShutdownCtx.Done():
+			stop()
 		case sig := <-sigCh:
 			logCtx := logger.AddContext(logger.Ctx{"signal": sig})
 			logCtx.Info("Received signal")
 			if d.ShutdownCtx.Err() != nil {
 				logCtx.Warn("Ignoring signal, shutdown already in progress")
-			} else {
-				go func() {
-					d.ShutdownDoneCh <- d.Stop()
-				}()
 			}
-
-		case err = <-d.ShutdownDoneCh:
-			return err
+			stop()
 		}
 	}
 }
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -70,7 +70,7 @@ func NewDaemon(ctx context.Context, project string) *Daemon {
 	return &Daemon{
 		ShutdownCtx:    ctx,
 		ShutdownCancel: cancel,
-		ShutdownDoneCh: make(chan error),
+		ShutdownDoneCh: make(chan error, 1),
 		ReadyChan:      make(chan struct{}),
 		project:        project,
 	}


tomponline (commented):
If shutdown can be initiated multiple times it suggests the check for an existing ongoing shutdown is being missed somewhere.

Rather than protecting this with a sync.Once, could we identity the scenario where an ongoing shutdown sequence is not detected, such that a correct error message is returned like line 116?

tomponline:
> * Don't check if ShutdownCtx is already cancelled, that will fail the check if the parent context is cancelled by other means.

This might be worth digging into, im not sure that sounds correct that a shutdownctx can be cancelled by a parent context (i.e it shouldn't have a parent context).

roosterfish (changes requested):
I think making the `ShutdownDoneCh` buffered is a good idea as it will block and never read again from the channel causing the hang you are experiencing.

Regarding the `ShutdownCtx` couldn't we just add another case for checking if it is cancelled and if so forward the error message into the now buffered `ShutdownDoneCh`:

```go
case <-d.ShutdownCtx.Done():
    d.ShutdownDoneCh <- d.ShutdownCtx.Err()
```

microcluster/app.go
```diff
@@ -96,21 +97,25 @@ func (m *MicroCluster) Start(apiEndpoints []rest.Endpoint, schemaExtensions map[
...
+			return err
+		case <-d.ShutdownCtx.Done():
+			stop()
 		case sig := <-sigCh:
 			logCtx := logger.AddContext(logger.Ctx{"signal": sig})
 			logCtx.Info("Received signal")
 			if d.ShutdownCtx.Err() != nil {
 				logCtx.Warn("Ignoring signal, shutdown already in progress")
-			} else {
-				go func() {
```
tomponline, line 107: I like that using sync.OnceFunc and buffered channel gets rid of the go routine though :)

microcluster/app.go
```diff
@@ -96,21 +97,25 @@ func (m *MicroCluster) Start(apiEndpoints []rest.Endpoint, schemaExtensions map[
...
+		case err = <-d.ShutdownDoneCh:
+			close(d.ShutdownDoneCh)
+			return err
+		case <-d.ShutdownCtx.Done():
+			stop()
 		case sig := <-sigCh:
 			logCtx := logger.AddContext(logger.Ctx{"signal": sig})
 			logCtx.Info("Received signal")
 			if d.ShutdownCtx.Err() != nil {
 				logCtx.Warn("Ignoring signal, shutdown already in progress")
```
tomponline, line 116: we should probably have a `continue` statement here so the loops restarts again without calling `stop()` unnecessarily if a shutdown is already started.

microcluster/app.go
tomponline: One drawback of getting rid of the go routine is that if subsequent signals are sent to the daemon, because the loop will be potentially blocked on calling `stop()` for the first time, the signals wont be handled and the `logCtx.Warn("Ignoring signal, shutdown already in progress")` part wont be reached.

tomponline (approved):
Overall I think this looks good, thanks!

neoaggelos:
> I think making the `ShutdownDoneCh` buffered is a good idea as it will block and never read again from the channel causing the hang you are experiencing.
> 
> Regarding the `ShutdownCtx` couldn't we just add another case for checking if it is cancelled and if so forward the error message into the now buffered `ShutdownDoneCh`:
> 
> ```go
> case <-d.ShutdownCtx.Done():
>     d.ShutdownDoneCh <- d.ShutdownCtx.Err()
> ```

I don't think that's right, who calls `d.Done()` in that case?

roosterfish:
> > I think making the `ShutdownDoneCh` buffered is a good idea as it will block and never read again from the channel causing the hang you are experiencing.
> > Regarding the `ShutdownCtx` couldn't we just add another case for checking if it is cancelled and if so forward the error message into the now buffered `ShutdownDoneCh`:
> > ```go
> > case <-d.ShutdownCtx.Done():
> >     d.ShutdownDoneCh <- d.ShutdownCtx.Err()
> > ```
> 
> I don't think that's right, who calls `d.Stop()` in that case?

This would require keeping the `go func() { ... }()` which will call `d.Stop()` in case of a signal from outside.

roosterfish (dismissed):
Thanks for splitting up the commit and putting the package prefixes. Looks good to me.

microcluster/app.go
```diff
@@ -96,21 +97,29 @@ func (m *MicroCluster) Start(apiEndpoints []rest.Endpoint, schemaExtensions map[
 		return fmt.Errorf("Unable to start daemon: %w", err)
 	}
 
+	stop := sync.OnceFunc(func() {
+		logger.Info("Initiate daemon stop sequence")
+		if err := d.Stop(); err != nil {
+			d.ShutdownDoneCh <- err
+			return
+		}
+		d.ShutdownDoneCh <- d.ShutdownCtx.Err()
```
tomponline, line 106: This needs a new empty line before it otherwise it'll fail our CI checks (well it would on LXD at least).

tomponline: Wont this mean that a "normal" shutdown will cause a Canceled error to be sent down the `d.ShutdownDoneCh` which will cause the daemon to exit with a non-zero exit code, possibly causing it to be restarted systemd?

tomponline: I'm not really seeing why we need to get the shutdown context error at all? (some comments in this part of the code base would be welcomed).

roosterfish: My argument was that if the daemon gets stopped due to the parent context being cancelled it can forward the actual reason why its now performing a stop.

tomponline: TBH I find this "parent context" thing a bit odd here. I guess this is because its a package rather than a daemon itself like LXD. But I think we need to differentiate between the **request** that triggers the shutdown (either a signal being received or the parent context being cancelled) and the **result** of the shutdown, which is what needs to influence the exit code of the program.

Also as an aside, I believe its somewhat an anti-pattern to be setting up signal handlers in an imported package.
Not something I would recommend.

Really the parent process should be setting those up and trigger a call to a .Stop() function or cancelling that parent context, rather than the package doing it itself.

neoaggelos: > Also as an aside, I believe its somewhat an anti-pattern to be setting up signal handlers in an imported package.
Not something I would recommend.
> 
> Really the parent process should be setting those up and trigger a call to a .Stop() function or cancelling that parent context, rather than the package doing it itself.

+100, and this is exactly what we do and how I bumped into the issue here. I would very much like to have a `DisableDefaultSignalHandlers bool` option in microcluster in the future to completely disable this, but that would have to be carefully setup to avoid breaking backwards compatibility.

This PR, admittedly, does not lead to a perfect world, but solves the problem with minimal changes to microcluster itself.

tomponline: CC @masnax

masnax:
So this perplexed me quite a bit until I looked over at the `k8sd` code and now it all fits together a bit.

So you set [signal.NotifyContext and then start MicroCluster](https://github.com/canonical/k8s-snap/blob/main/src/k8s/cmd/main.go#L21-L33) after passing the context around to [here](https://github.com/canonical/k8s-snap/blob/962746c7de9cf685a6bf1e73c1ab199fd2b0152c/src/k8s/cmd/k8sd/k8sd.go#L22) and then [here where it's finally assigned to the daemon](https://github.com/canonical/k8s-snap/blob/962746c7de9cf685a6bf1e73c1ab199fd2b0152c/src/k8s/pkg/k8sd/app/app.go#L40).

But the resulting context from `signal.NotifyContext` will [cancel in a goroutine if a signal is received.](https://cs.opensource.google/go/go/+/refs/tags/go1.22.1:src/os/signal/signal.go;l=290)

So in this case I don't think we should necessarily work around handling signals in two places with this approach as it's mostly papering over the real issue. Additionally, we should not be using the `cmd` parent context to directly manage the stop sequence in the microcluster daemon at all. 

> I would very much like to have a DisableDefaultSignalHandlers bool option in microcluster in the future to completely disable this, but that would have to be carefully setup to avoid breaking backwards compatibility.

If for the sake of consistency in your codebase you would prefer to always re-use the same signal notifier for all of the cases in `main.go`, I think adding the `DisableDefaultSignalHandlers` option is the best way to go forward, and I don't think it would be that difficult.

There's already a [microcluster.Args](https://github.com/canonical/microcluster/blob/8d4c3238a91bd3c4b6dba089881de2c8cd6e494d/microcluster/app.go#L40) struct for passing configuration to the daemon. We can simply add a `DisableDefaultSignalHandlers` boolean here, and in that case do the following, let me know what you think:

```go
        // Don't listen for signals if disable flag is set. The default behaviour will be to set it up if the flag is omitted, for backwards compatibility.
        sigCh := make(chan os.Signal, 1)
        if !m.args.DisableDefaultSignalHandlers {
          signal.Notify(sigCh, unix.SIGPWR)
	  signal.Notify(sigCh, unix.SIGINT)
	  signal.Notify(sigCh, unix.SIGQUIT)
	  signal.Notify(sigCh, unix.SIGTERM)

	  chIgnore := make(chan os.Signal, 1)
	  signal.Notify(chIgnore, unix.SIGHUP)
        }

        // Don't use the parent context for the daemon as that will certainly cause problems that we will have to work around.
	d := daemon.NewDaemon(context.Background(), cluster.GetCallerProject())
	err = d.Init(m.args.ListenPort, m.FileSystem.StateDir, m.FileSystem.SocketGroup, apiEndpoints, schemaExtensions, hooks)
	if err != nil {
		return fmt.Errorf("Unable to start daemon: %w", err)
	}

        // Use `sync.Once` to ensure we only run the stop sequence once.
        stop := sync.OnceFunc(func() {
          d.ShutdownDoneCh <- d.Stop()
        })

	for {
		select {
                // Add a case to handle an external shutdown sequence if the microcluster context is cancelled. This should be distinct from the daemon context.
                case <-m.ctx.Done():
                        logger.Info("Started external shutdown sequence", logger.Ctx{"error": m.ctx.Err()})
                        stop()

                // The following case will just block forever if m.args.DisableSignalHandlers is true.
		case sig := <-sigCh:
			logCtx := logger.AddContext(logger.Ctx{"signal": sig})
			logCtx.Info("Received signal")
			if d.ShutdownCtx.Err() != nil {
				logCtx.Warn("Ignoring signal, shutdown already in progress")
			}

                        stop()

                // This case will only happen after we run the stop func, so we can just close the buffered channel and return.
		case err = <-d.ShutdownDoneCh:
                        close(d.ShutdownDoneCh)                       
			return err
		}
	}

```

tomponline:
> So in this case I don't think we should necessarily work around handling signals in two places with this approach as it's mostly papering over the real issue. Additionally, we should not be using the cmd parent context to directly manage the stop sequence in the microcluster daemon at all.

What is the patent context for. If not to trigger cancelation of the server?

masnax:
> > So in this case I don't think we should necessarily work around handling signals in two places with this approach as it's mostly papering over the real issue. Additionally, we should not be using the cmd parent context to directly manage the stop sequence in the microcluster daemon at all.
> 
> What is the patent context passed to NewDaemon for. If not to trigger cancelation of the server?
> 
> Why is there a d.Stop() and a NewDaemon context?

The daemon needs to be able to cancel itself and propagate that up to `microcluster.App`, as well as an external action cancelling the parent context, and sending it down to `microcluster.App`, which in turn sends it further down to `Daemon`. 

If the the daemon's shutdown context is done, we can't tell if it was cancelled by `d.Stop` or some external error, which would tell us we still have to call `d.Stop`. 

Keeping them separate makes this easier to reason about:
* If `m.ctx` was cancelled, move down (to call `d.Stop`). 
* If `d.ctx` was cancelled, move up (to unblock `m.Start`). 

There's definitely other ways to do this but we'd probably have to introduce another channel and at that point I think it's unneeded complexity.

tomponline:
> > > So in this case I don't think we should necessarily work around handling signals in two places with this approach as it's mostly papering over the real issue. Additionally, we should not be using the cmd parent context to directly manage the stop sequence in the microcluster daemon at all.
> > 
> > 
> > What is the patent context passed to NewDaemon for. If not to trigger cancelation of the server?
> > Why is there a d.Stop() and a NewDaemon context?
> 
> The daemon needs to be able to cancel itself and propagate that up to `microcluster.App`, as well as an external action cancelling the parent context, and sending it down to `microcluster.App`, which in turn sends it further down to `Daemon`.
> 
> If the the daemon's shutdown context is done, we can't tell if it was cancelled by `d.Stop` or some external error, which would tell us we still have to call `d.Stop`.
> 
> Keeping them separate makes this easier to reason about:
> 
>     * If `m.ctx` was cancelled, move down (to call `d.Stop`).
> 
>     * If `d.ctx` was cancelled, move up (to unblock `m.Start`).
> 
> 
> There's definitely other ways to do this but we'd probably have to introduce another channel and at that point I think it's unneeded complexity.

I'm afraid I'm not following this. Please can you make a note to discuss it in our one to one on Monday. Overall I think this needs simplifying. Also I understand it is a Go anti pattern to store long lived contexts so it doesn't seem correct to have one passed to NewDaemon.

tomponline:
> > > > So in this case I don't think we should necessarily work around handling signals in two places with this approach as it's mostly papering over the real issue. Additionally, we should not be using the cmd parent context to directly manage the stop sequence in the microcluster daemon at all.
> > > 
> > > 
> > > What is the patent context passed to NewDaemon for. If not to trigger cancelation of the server?
> > > Why is there a d.Stop() and a NewDaemon context?
> > 
> > 
> > The daemon needs to be able to cancel itself and propagate that up to `microcluster.App`, as well as an external action cancelling the parent context, and sending it down to `microcluster.App`, which in turn sends it further down to `Daemon`.
> > If the the daemon's shutdown context is done, we can't tell if it was cancelled by `d.Stop` or some external error, which would tell us we still have to call `d.Stop`.
> > Keeping them separate makes this easier to reason about:
> > ```
> > * If `m.ctx` was cancelled, move down (to call `d.Stop`).
> > 
> > * If `d.ctx` was cancelled, move up (to unblock `m.Start`).
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > There's definitely other ways to do this but we'd probably have to introduce another channel and at that point I think it's unneeded complexity.
> 
> I'm afraid I'm not following this. Please can you make a note to discuss it in our one to one on Monday. Overall I think this needs simplifying. Also I understand it is a Go anti pattern to store long lived contexts so it doesn't seem correct to have one passed to NewDaemon.

https://go.dev/blog/context-and-structs

> Contexts should not be stored inside a struct type, but instead passed to each function that needs it.

masnax:
> > Contexts should not be stored inside a struct type, but instead passed to each function that needs it.
 
We can clear out the context fields from the structs to be more go-compliant but that's not the root of the problem. None of this code is highly coupled with that requirement.

Here's a quick rundown of the issue without any context fields.

```go
func (m *microcluster) Start(ctx context.Context) {
  // Spawn a new daemon. 
  d := NewDaemon()
  // Init the daemon, which starts the listener.
  d.Init(ctx)
  
  // Block until the daemon is stopped.
  for {
     select {
       // The daemon is stopped so we can return.
       case: <- d.ShutdownDoneCh::
          return
       // We received a signal to stop, so shutdown the daemon.
       case: <- ReceivedSignal():
         if ctx.Err() != nil {
           // skip as we --incorrectly-- expect a shutdown to be in progress.
         } else {
           d.Stop()
           d.ShutdownDoneCh <- nil
         }
     }
  }
}

func (d *Daemon) Stop() {
  d.CancelFunc()
  d.additionalCleanup()
}
```

So with the above, we
```go
// Get a context that might cancel itself in a goroutine.
ctx := ContextThatRandomlyMightCancel()
m.Start(ctx)

// If the ctx above is randomly cancelled, the `ctx.Err() != nil` check 
// above would result in a hang because we would never receive 
// from `d.ShutdownDoneCh`. 
```

So we need either a second context, or a different mechanism to check for the case of an in-progress stop sequence.

neoaggelos:
@masnax @tomponline 

OK, though I have also seen the issues raised with how microcluster deals with contexts in a few occations, and appreciate the discussion that has occured here, please let me know how you would like to proceed.

The `DisableDefaultSignalHandlers` only deals with one of the two issues, it still does not resolve calling the `d.Stop()` sequence (since it's currently only under the `sig := <-sigCh` case, and will not even be called if the context is already cancelled).

masnax:
Closing this as it should be fixed by the merged PR #95

PR #94, opened 2024-03-15
Separately handle internal and external schema updates

### LINKED ISSUE #91: Internal schema updates don't apply if external ones exist
If an external schema update already exists, appending a new internal schema update will try to run the newest update only. However, since the internal list is appended with the external list, it ends up re-running the last external update instead of the new internal one.

masnax:
Closes #91 and unblocks #33 

Updates the `schemas` table to replace the `version` column with `internal_version` and `external_version` to represent microcluster-internal schema updates from updates passed in by importers of microcluster. Since we only have an internal V0 currently, we will assume all existing schema updates after the first one are external for existing deployments. 

Necessarily, the `internal_cluster_members` table, and the corresponding API types for `HeartbeatInfo` and `ClusterMember` also have 2 fields for schema versions now.

Unlike the API extensions feature from #33 that labels each API extension so we can actually parse them and thus lets us easily hide the internal/external distinction, the database schema version is just an integer so we must expose it.

Previous to this PR, the behaviour of the schema update logic was to take the internal list and append the external list to it, then check the current version against the size of the list and apply all the schema updates from the list starting at the index of the current version. This posed a major problem in that introducing a new internal schema update would actually run the latest external schema update instead, as the list would get incremented by one, but the last entry to the list would always be an external update. This PR also addresses the rare edge case where an internal,external list like {3,2} might incorrectly be viewed as the same as {2,3}.

Similar to #33, since this schema update affects the schema validation process itself, the `updateFromV1` function will run both during the normal schema update process, in order to increment the schema version, as well as within the schema validation logic itself, after we check that a schema table exists and try to create it. This happens [here](https://github.com/canonical/microcluster/blob/6908c5c2454aad33874a75d1f82cbd9b59d8f63f/internal/db/update/schema.go#L73) before we attempt to fetch the schema version if the schema table exists, and [here](https://github.com/canonical/microcluster/blob/6908c5c2454aad33874a75d1f82cbd9b59d8f63f/internal/db/update/schema.go#L242) just after we create the schema table itself.


internal/db/update/schema.go
```diff
@@ -37,8 +37,8 @@ func (s *SchemaUpdate) Check(check schema.Check) {
 	s.check = check
 }
 
-func (s *SchemaUpdate) Version() int {
-	return len(s.updates)
+func (s *SchemaUpdate) Version() (int, int) {
```
gabrielmougard, line 40: Could we have a function prototype like `func (s *SchemaUpdate) Version() (internal int, external int)` to distinguish the return values (useful for `gopls` in VSCode when hovering the code)

gabrielmougard (approved):
A small suggestion but else, very good!

internal/db/update/schema.go
tomponline, line 40: Doc block or named var names need to explain what the 2 ints returned here are.

tomponline: Or perhaps return a [2]int array so you can treat the combined version as  a single data type?

internal/db/update/schema.go
```diff
@@ -67,15 +67,17 @@ func (s *SchemaUpdate) Ensure(db *sql.DB) (int, error) {
 			return fmt.Errorf("failed to check if schema table is there: %w", err)
 		}
 
-		current := 0
+		versions := []int{0, 0}
```
tomponline, line 70: This doesn't need pre-initialising.

tomponline, line 70: Rather than pre-initialising you could do a nil check on line 99.

internal/db/update/schema.go
```diff
@@ -126,25 +128,30 @@ func (s *SchemaUpdate) Ensure(db *sql.DB) (int, error) {
 func (s *SchemaUpdate) Dump(db *sql.DB) (string, error) {
 	var statements []string
 	err := query.Transaction(context.TODO(), db, func(ctx context.Context, tx *sql.Tx) error {
-		versions, err := query.SelectIntegers(ctx, tx, "SELECT version FROM schemas ORDER BY version")
-		if err != nil {
-			return err
-		}
+		current := 0
+		for _, versionType := range []int{0, 1} {
```
tomponline, line 132: 0 and 1 should be defined as constants imho to avoid the use of "magic numbers" in this important part of the code base.

internal/db/update/schema.go
```diff
@@ -126,25 +128,30 @@ func (s *SchemaUpdate) Ensure(db *sql.DB) (int, error) {
 func (s *SchemaUpdate) Dump(db *sql.DB) (string, error) {
 	var statements []string
 	err := query.Transaction(context.TODO(), db, func(ctx context.Context, tx *sql.Tx) error {
-		versions, err := query.SelectIntegers(ctx, tx, "SELECT version FROM schemas ORDER BY version")
-		if err != nil {
-			return err
-		}
+		current := 0
```
tomponline, line 131: current what? Naming could be improved here.

internal/db/update/schema.go
```diff
@@ -126,25 +128,30 @@ func (s *SchemaUpdate) Ensure(db *sql.DB) (int, error) {
...
+			versions, err := query.SelectIntegers(ctx, tx, "SELECT version FROM schemas WHERE type = ? ORDER BY version", versionType)
+			if err != nil {
+				return err
+			}
 
-		if len(versions) == 0 {
-			return fmt.Errorf("expected schema table to contain at least one row")
-		}
+			if len(versions) == 0 {
+				return fmt.Errorf("expected schema table to contain at least one row")
```
tomponline, line 139: Errors should start with capital letter?

internal/db/update/schema.go
```diff
@@ -126,25 +128,30 @@ func (s *SchemaUpdate) Ensure(db *sql.DB) (int, error) {
...
+				return err
+			}
+
+			current += versions[len(versions)-1]
 		}
 
-		current := versions[len(versions)-1]
-		if current != len(s.updates) {
-			return fmt.Errorf("Update level is %d, expected %d", current, len(s.updates))
+		if current != len(s.internalUpdates)+len(s.externalUpdates) {
```
tomponline, line 150: Would it be more helpful to do this check inside the for loop for each schema version type so that the error can indicate which schema type is having problems?

internal/db/update/schema.go
```diff
@@ -156,11 +163,19 @@ func (s *SchemaUpdate) Dump(db *sql.DB) (string, error) {
...
 
 	// Add a statement for inserting the current schema version row.
-	statements = append(
-		statements,
-		fmt.Sprintf(`
-INSERT INTO schemas (version, updated_at) VALUES (%d, strftime("%%s"))
-`, len(s.updates)))
+	for _, versionType := range []int{0, 1} {
+		updates := s.internalUpdates
+		if versionType == 1 {
```
tomponline, line 168: yeah we need to use constants - otherwise its going to get confusing.

tomponline: question: why have two variables s.internalUpdates and s.externalUpdates, rather than s.updates being a map[int][]schema.Update?

Also we should get into the habit of using "thing{Action/Type}" naming like we do elsewhere, so if we stick with 2 vars, they should be `updatesInternal` and `updatesExternal`.

tomponline: Could we introduce a data type for the schema type too so it becomes: `map[schemaType]schema.Update`?

internal/db/update/schema.go
```diff
@@ -306,28 +337,36 @@ SELECT COUNT(name) FROM sqlite_master WHERE type = 'table' AND name = 'schemas'
 // trigger the associated Update value. It's required that the minimum key in
 // the map is 1, and if key N is present then N-1 is present too, with N>1
 // (i.e. there are no missing versions).
-func NewFromMap(versionsToUpdates map[int]schema.Update) *SchemaUpdate {
+func NewFromMap(internalUpdates, externalUpdates map[int]schema.Update) *SchemaUpdate {
```
tomponline, line 340: We dont use the compressed argument form please.

internal/db/update/schema.go
```diff
@@ -306,28 +337,36 @@ SELECT COUNT(name) FROM sqlite_master WHERE type = 'table' AND name = 'schemas'
...
 	// Collect all version keys.
-	versions := []int{}
-	for version := range versionsToUpdates {
-		versions = append(versions, version)
-	}
 
-	// Sort the versions,
-	sort.Ints(versions)
+	schemaUpdate := &SchemaUpdate{}
+	for i, update := range []map[int]schema.Update{internalUpdates, externalUpdates} {
```
tomponline, line 344: As I've seen this pattern in a few places ` range []map[int]schema.Update{internalUpdates, externalUpdates}` is it worth making a data type that has a `ToOrderedSlice()` function that returns the map in the correct order?

internal/db/update/schema.go
```diff
@@ -306,28 +337,36 @@ SELECT COUNT(name) FROM sqlite_master WHERE type = 'table' AND name = 'schemas'
...
+		versions := []int{}
+		for version := range update {
+			versions = append(versions, version)
+		}
+
+		// Sort the versions,
+		sort.Ints(versions)
```
tomponline, line 351: It concerns me that we use `map[int]schema.Update` for an ordered list of updates, given that maps are iterated in random order. I realise this is a historical carry over from LXD.

But I wonder if we can do better here and use a []schema.Update instead?

tomponline (commented):
I'm going to leave my review there for now and revisit once you're ready for the next pass. Thanks

internal/db/update/schema.go
```diff
@@ -67,15 +67,17 @@ func (s *SchemaUpdate) Ensure(db *sql.DB) (int, error) {
 			return fmt.Errorf("failed to check if schema table is there: %w", err)
 		}
 
-		current := 0
+		versions := []int{0, 0}
 		if exists {
-			versions, err := query.SelectIntegers(ctx, tx, "SELECT version FROM schemas ORDER BY version")
+			// updateFromV1 changes the schema table and needs to be run before we calculate the schema version.
+			err := updateFromV1(ctx, tx)
```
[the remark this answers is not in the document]
masnax: No, `updateFromV0` already exists, it's the first schema update and the reason that the `updateFromV1` decrements entries into `external_updates` by 1.

internal/db/update/schema.go
masnax: I've made the schema update lists into slices now, but this may have adverse effects on any project that didn't structure their external schema update map quite like our internal one with ordered indexes.

masnax: The micro-projects are all fine on this front.

internal/db/update/schema.go
```diff
@@ -156,11 +163,19 @@ func (s *SchemaUpdate) Dump(db *sql.DB) (string, error) {
...
-		fmt.Sprintf(`
-INSERT INTO schemas (version, updated_at) VALUES (%d, strftime("%%s"))
-`, len(s.updates)))
+	for _, versionType := range []int{0, 1} {
+		updates := s.internalUpdates
+		if versionType == 1 {
+			updates = s.externalUpdates
+		}
+
+		statements = append(
```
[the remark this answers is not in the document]
masnax: I actually removed this whole code block since it's entirely unused.

internal/db/update/schema.go
masnax: As mentioned above, I've made all of the schema update maps into slices now.

masnax:
@tomponline Should be ready for another pass. Main changes here are the following:

- Added an `updateType` to represent internal and external schema updates in the database.

- Replaced the `map[int]schema.Update` maps with `[]schema.Update`. This also extends to external schema updates so there is a chance that it might have adverse effects if some project wasn't relying on ordered keys in the map like we were, but I think it's better to rip the bandaid off here.

- Removed the `Dump` and `SchemaDotGo` functions and their associated helpers since they are not used in microcluster at all.

internal/db/update/schema.go
```diff
@@ -126,25 +140,31 @@ func (s *SchemaUpdate) Ensure(db *sql.DB) (int, error) {
 func (s *SchemaUpdate) Dump(db *sql.DB) (string, error) {
 	var statements []string
 	err := query.Transaction(context.TODO(), db, func(ctx context.Context, tx *sql.Tx) error {
-		versions, err := query.SelectIntegers(ctx, tx, "SELECT version FROM schemas ORDER BY version")
-		if err != nil {
-			return err
-		}
+		currentVersion := 0
+		for versionType := range s.updates {
```
tomponline: Do we need to interate internal updates first before external ones? Like this the order is random. Does it make a difference?

internal/db/update/schema.go
```diff
@@ -190,37 +214,50 @@ ORDER BY name
...
+func ensureUpdatesAreApplied(ctx context.Context, tx *sql.Tx, versions []int, schemaUpdates map[updateType][]schema.Update, hook schema.Hook) error {
+	if versions == nil {
+		versions = []int{0, 0}
 	}
 
-	// If there are no updates, there's nothing to do.
-	if len(updates) == 0 {
-		return nil
-	}
+	for updateType, updates := range schemaUpdates {
```
tomponline, line 139: Do we need to iterate schemaUpdates such that internal updates are processed first, rather than in random order?

internal/db/update/schema.go
```diff
@@ -6,20 +6,31 @@ import (
...
-	"sort"
 	"strings"
 
 	"github.com/canonical/lxd/lxd/db/query"
 	"github.com/canonical/lxd/lxd/db/schema"
 	"github.com/canonical/lxd/shared"
 )
 
+// updateType represents whether the update is an internal or external schema update.
+type updateType int
```
tomponline, line 16: should this be a uint?

cluster/cluster_members.go
```diff
@@ -73,18 +74,19 @@ func (c InternalClusterMember) ToAPI() (*internalTypes.ClusterMember, error) {
 			Address:     address,
 			Certificate: *certificate,
 		},
-		Role:          string(c.Role),
-		SchemaVersion: c.Schema,
-		LastHeartbeat: c.Heartbeat,
-		Status:        internalTypes.MemberUnreachable,
+		Role:                  string(c.Role),
+		InternalSchemaVersion: c.InternalSchema,
```
tomponline, line 78: Lets use the normal `thingVariant` pattern, so `SchemaInternalVersion`

cluster/cluster_members.go
```diff
@@ -102,12 +104,28 @@ func UpdateClusterMemberSchemaVersion(tx *sql.Tx, version int, address string) e
 
 // GetClusterMemberSchemaVersions returns the schema versions from all cluster members that are not pending.
 // This helper is non-generated to work before generated statements are loaded, as we update the schema.
-func GetClusterMemberSchemaVersions(ctx context.Context, tx *sql.Tx) ([]int, error) {
-	sql := "SELECT schema FROM internal_cluster_members WHERE NOT role='pending'"
-	versions, err := query.SelectIntegers(ctx, tx, sql)
+func GetClusterMemberSchemaVersions(ctx context.Context, tx *sql.Tx) ([]int, []int, error) {
```
tomponline, line 107: would be good to use named return vars here for clarity to the caller.

internal/daemon/daemon.go
```diff
@@ -374,13 +374,15 @@ func (d *Daemon) StartAPI(bootstrap bool, initConfig map[string]string, newConfi
...
-			Heartbeat:   time.Time{},
-			Role:        cluster.Pending,
+			Name:           localNode.Name,
+			Address:        localNode.Address.String(),
+			Certificate:    localNode.Certificate.String(),
+			InternalSchema: internalSchema,
+			ExternalSchema: externalSchema,
+			Heartbeat:      time.Time{},
+			Role:           cluster.Pending,
 		}
```
tomponline, line 387: You could also call it here as `clusterMember.InternalSchema, clusterMember.ExternalSchema = d.db.Schema().Version()` and avoid the temp var.

internal/daemon/daemon.go
```diff
@@ -77,7 +77,7 @@ func NewDaemon(ctx context.Context, project string) *Daemon {
 }
 
 // Init initializes the Daemon with the given configuration, and starts the database.
-func (d *Daemon) Init(listenPort string, stateDir string, socketGroup string, extendedEndpoints []rest.Endpoint, schemaExtensions map[int]schema.Update, hooks *config.Hooks) error {
+func (d *Daemon) Init(listenPort string, stateDir string, socketGroup string, extendedEndpoints []rest.Endpoint, schemaExtensions []schema.Update, hooks *config.Hooks) error {
```
tomponline, line 80: Should the doc block explain how schemaExtensions should be ordered to help the caller?

microcluster/app.go
```diff
@@ -71,7 +71,7 @@ func App(ctx context.Context, args Args) (*MicroCluster, error) {
 
 // Start starts up a brand new MicroCluster daemon. Only the local control socket will be available at this stage, no
 // database exists yet. Any api or schema extensions can be applied here.
-func (m *MicroCluster) Start(apiEndpoints []rest.Endpoint, schemaExtensions map[int]schema.Update, hooks *config.Hooks) error {
+func (m *MicroCluster) Start(apiEndpoints []rest.Endpoint, schemaExtensions []schema.Update, hooks *config.Hooks) error {
```
tomponline, line 74: Should the doc block explain how schemaExtensions should be ordered to help the caller?

tomponline (changes requested):
Looking better. In general we tend to use `thingVariant` type patterns, so it would be good to go through this PR and change `internalSchema*` to `schemaInternal*` for vars and DB column names. It helps with auto complete in IDEs for example.

internal/db/update/schema.go
masnax: Thanks, I overlooked that. I think for sanity's sake the internal updates should come first, but either case has trade-offs.

- If all internal updates come first, but an external update depends on an internal table, a new internal update can it can change the table from under that pre-existing external update unexpectedly.

- If internal updates don't come first, then an external update can change an internal table and then the following internal updates could break.

It speaks to the larger problem that we have no way of restricting external schema updates to not depend on internal ones.

internal/daemon/daemon.go
masnax: I'm not following, do you mean we should add a comment explaining that external schema updates should be passed in as a list in the order the updates should be applied?

We should have such a comment but I think it belongs in `app.Start` where a project importing microcluster would see it.

microcluster/app.go
masnax: Ah okay, so both, got it :)

internal/db/update/schema.go
```diff
@@ -6,20 +6,31 @@ import (
...
+)
+
+// SchemaUpdate holds the configuration for executing schema updates.
 type SchemaUpdate struct {
-	updates []schema.Update // Ordered series of updates making up the schema
-	hook    schema.Hook     // Optional hook to execute whenever a update gets applied
-	fresh   string          // Optional SQL statement used to create schema from scratch
-	check   schema.Check    // Optional callback invoked before doing any update
-	path    string          // Optional path to a file containing extra queries to run
+	updates map[updateType][]schema.Update // Ordered series of internal updates making up the schema
```
tomponline, line 28: Internal updates? It presumably also supports external ones too due to the updateType key?

internal/db/update/schema.go
tomponline: Did this get addressed?

cluster/cluster_members.go
```diff
@@ -73,18 +74,19 @@ func (c InternalClusterMember) ToAPI() (*internalTypes.ClusterMember, error) {
...
+		Status:                internalTypes.MemberUnreachable,
 	}, nil
 }
 
 // UpdateClusterMemberSchemaVersion sets the schema version for the cluster member with the given address.
 // This helper is non-generated to work before generated statements are loaded, as we update the schema.
-func UpdateClusterMemberSchemaVersion(tx *sql.Tx, version int, address string) error {
-	stmt := "UPDATE internal_cluster_members SET schema=? WHERE address=?"
-	result, err := tx.Exec(stmt, version, address)
+func UpdateClusterMemberSchemaVersion(tx *sql.Tx, internalVersion int, externalVersion int, address string) error {
```
tomponline, line 87: would it be a problem to change these to be uints to better reflect what they are?

tomponline: And if they are coming to/from the db they should probably be uint64.

internal/daemon/daemon.go
```diff
@@ -77,6 +77,9 @@ func NewDaemon(ctx context.Context, project string) *Daemon {
 }
 
 // Init initializes the Daemon with the given configuration, and starts the database.
+// - `extendedEndpoints` is a list of endpoints to be served over `/1.0`.
```
tomponline, line 80: endpointsExtented

microcluster/app.go
```diff
@@ -71,6 +71,9 @@ func App(ctx context.Context, args Args) (*MicroCluster, error) {
 
 // Start starts up a brand new MicroCluster daemon. Only the local control socket will be available at this stage, no
 // database exists yet. Any api or schema extensions can be applied here.
+// - `schemaExtensions` is a list of schema updates in the order that they should be applied.
+// - `apiEndpoints` is a list of endpoints to be served over `/1.0`.
```
tomponline, line 75: endpointsAPI?

microcluster/app.go
```diff
@@ -71,6 +71,9 @@ func App(ctx context.Context, args Args) (*MicroCluster, error) {
...
+// - `hooks` are a set of functions that trigger at certain points during cluster communication.
 func (m *MicroCluster) Start(apiEndpoints []rest.Endpoint, schemaExtensions []schema.Update, hooks *config.Hooks) error {
```
tomponline, line 77: Apparently `const dotGoTemplate = "package %s\n\n" +` is unused?

tomponline (commented):
Looking good, a couple of small nits and questions.

I'm assuming all of these has been thoroughly tested in an existing cluster though. :)

internal/db/update/schema.go
masnax: Yes, here: https://github.com/canonical/microcluster/pull/94/files#diff-268ba13a60418730fa859d08731781b80293f421148bcfc6e507f8f2f8a95709R139-R141

microcluster/app.go
masnax: Yes, I missed removing the constant. thanks.

tomponline:
Once @markylaing or @roosterfish are happy with this ill do a final pass. Thanks

internal/db/db.go
```diff
@@ -38,37 +38,51 @@ func (db *DB) Open(bootstrap bool, project string) error {
...
+						otherNodesBehind = true
+						break
+					}
+
+					// Another node has a version greater than ours
+					// and presumeably is waiting for other nodes
+					// to upgrade. Let's error out and shutdown
+					// since we need a greater version.
+					otherNodesBehind = false
+					return fmt.Errorf("this node's version is behind, please upgrade")
```
markylaing, line 78: I'm finding this block a little hard to read. Would it be clearer if we define a function for the inner loop? Something like:
```golang
checkVersion := func (schemaVersion uint64, versions []uint64) error {
	for _, version := range versions {
		if schemaVersion == version {
			// Versions are equal, there's hope for the
			// update. Let's check the next node.
			continue
		}

		if schemaVersion > version {
			// Our version is bigger, we should stop here
			// and wait for other nodes to be upgraded and
			// restarted.
			otherNodesBehind = true
			break
		}

		// Another node has a version greater than ours
		// and presumeably is waiting for other nodes
		// to upgrade. Let's error out and shutdown
		// since we need a greater version.
		otherNodesBehind = false
		return fmt.Errorf("this node's version is behind, please upgrade")
	}
}
```

Then have two calls, one for the internal version and one for the external:
```golang
err = checkVersion(internalSchemaVersion, internalVersions)
if err != nil {
	return err
}

err = checkVersion(externalSchemaVersion, externalVersions)
if err != nil {
	return err
}
```

internal/db/update/update.go
```diff
@@ -21,76 +18,85 @@ CREATE TABLE schemas (
...
-	}
-
-	schema := NewFromMap(m.updates)
+// updateFromV1 fixes a bug in the schemas table. Previously there was no way to tell when an update was internal, so the last external update would be re-run instead.
+// To fix this, we introduce a `type` column to the schemas table. The first schema update will be considered internal, as that is what microcluster initially shipped with.
+// All other schema versions will be considered external.
+//
+// Because this update affects the schema update mechanism, it must necessarily be run manually before the regular schema updates.
+// So first it checks if the column exists already, and in such a case does nothing.
+func updateFromV1(ctx context.Context, tx *sql.Tx) error {
```
markylaing, line 52: I think as this schema update is quite critical it would be worth having a unit test to verify that it does exactly what is expected.

markylaing (changes requested):
I think we should add a unit test for the schema update to ensure it's working as expected. Other than that I just have the one nit about readability of the version checking loop in the schema update function.

cluster/cluster_members.go
```diff
@@ -73,18 +74,19 @@ func (c InternalClusterMember) ToAPI() (*internalTypes.ClusterMember, error) {
...
 	}, nil
 }
 
 // UpdateClusterMemberSchemaVersion sets the schema version for the cluster member with the given address.
 // This helper is non-generated to work before generated statements are loaded, as we update the schema.
-func UpdateClusterMemberSchemaVersion(tx *sql.Tx, version int, address string) error {
-	stmt := "UPDATE internal_cluster_members SET schema=? WHERE address=?"
-	result, err := tx.Exec(stmt, version, address)
+func UpdateClusterMemberSchemaVersion(tx *sql.Tx, internalVersion uint64, externalVersion uint64, address string) error {
+	stmt := "UPDATE internal_cluster_members SET internal_schema=?,external_schema=? WHERE address=?"
```
roosterfish, line 88: There looks to be a typo. The columns in the database schema are called `schema_internal` and `schema_external`.

cluster/cluster_members.go
```diff
@@ -102,12 +104,28 @@ func UpdateClusterMemberSchemaVersion(tx *sql.Tx, version int, address string) e
 
 // GetClusterMemberSchemaVersions returns the schema versions from all cluster members that are not pending.
 // This helper is non-generated to work before generated statements are loaded, as we update the schema.
-func GetClusterMemberSchemaVersions(ctx context.Context, tx *sql.Tx) ([]int, error) {
-	sql := "SELECT schema FROM internal_cluster_members WHERE NOT role='pending'"
-	versions, err := query.SelectIntegers(ctx, tx, sql)
+func GetClusterMemberSchemaVersions(ctx context.Context, tx *sql.Tx) (internalSchema []uint64, externalSchema []uint64, err error) {
+	sql := "SELECT internal_schema,external_schema FROM internal_cluster_members WHERE NOT role='pending'"
```
roosterfish, line 108: Same as above with the column names.

roosterfish (changes requested):
Looking good, just a small name mismatch in the DB columns. I agree with @markylaing on the test case.

cluster/cluster_members.go
masnax: Ah good catch, I swapped them around yesterday but missed these ones.

internal/db/db.go
masnax: I like this, though it would mean @gabrielmougard's PR would also need to implement a second function for comparing API extensions.

gabrielmougard: It's fine by me :+1:

tomponline:
@masnax when can these changes be addressed? Keen to get this merged to unblock @gabrielmougard 

Thanks

gabrielmougard:
> @tomponline @markylaing @roosterfish This should be ready for another look-over.
> 
> The main change here is just adding tests. A few small changes had to be made to facilitate testing:
> 
> * The schema update logic and waiting for other nodes has been moved to its own function and is now called from `db.Open` instead of being directly implemented in `db.Open`
> * There are a few additional setters for the `update.SchemaUpdateManager` so that we can manually change schema updates for tests.

Thanks for the update!

roosterfish (dismissed):
The tests are looking good and robust! I really like that the external schemas are now passed using a slice instead of the map as it feels much more intuitive.

internal/db/update/schema.go
```diff
@@ -155,11 +175,15 @@ func (s *SchemaUpdate) Dump(db *sql.DB) (string, error) {
...
-		statements,
-		fmt.Sprintf(`
-INSERT INTO schemas (version, updated_at) VALUES (%d, strftime("%%s"))
-`, len(s.updates)))
+	for versionType := range s.updates {
+		updates := s.updates[versionType]
+		statements = append(
+			statements,
+			fmt.Sprintf(`
+INSERT INTO schemas (version, type, updated_at) VALUES (%d, %d, strftime("%%s"))
```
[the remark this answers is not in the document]
markylaing: `strftime` is a sqlite function that takes a format argument (https://www.sqlite.org/lang_datefunc.html). So the double percentage is to tell `fmt.Sprintf` to actually put a percentage there.

internal/db/update/update_test.go
```diff
@@ -0,0 +1,168 @@
...
+		schema.updates[updateInternal] = append(schema.updates[updateInternal], t.upgradesInternal...)
+		schema.updates[updateExternal] = append(schema.updates[updateExternal], t.upgradesExternal...)
+
+		_, err = schema.Schema().Ensure(db)
+		s.NoError(err)
+
+		ctx := context.Background()
+		tx, err := db.BeginTx(ctx, nil)
+		s.NoError(err)
+		versions, err := query.SelectIntegers(ctx, tx, "SELECT MAX(version) FROM schemas WHERE type = 0 UNION ALL SELECT COALESCE(MAX(version), 0) FROM schemas WHERE type = 1")
```
markylaing, line 140: I think it would also be good to double check the values of `schema_internal` and `schema_external` in `internal_cluster_members`.

markylaing (changes requested):
Just one nit about the unit tests and then I think we're good to go :)

gabrielmougard (approved):
The PR looks good. And it's nice to have the DB test suite. Thanks!

internal/db/update/schema.go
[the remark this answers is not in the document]
roosterfish: Ah yes, thanks, I have overlooked that.

internal/db/update/schema.go
```diff
@@ -125,25 +139,31 @@ func (s *SchemaUpdate) Ensure(db *sql.DB) (int, error) {
...
+			}
+
+			currentVersion += versions[len(versions)-1]
 		}
 
-		current := versions[len(versions)-1]
-		if current != len(s.updates) {
-			return fmt.Errorf("Update level is %d, expected %d", current, len(s.updates))
+		schemaVersion := len(s.updates[updateInternal]) + len(s.updates[updateExternal])
+		if currentVersion != len(s.updates[updateInternal])+len(s.updates[updateExternal]) {
```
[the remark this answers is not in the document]
masnax: Did you grab this from a middle commit? This whole block is removed by 3832cff.

internal/db/update/update_test.go
masnax: Hmm, I suppose for safety I could try to spoof some values into this test too, as the upgrade test in `db_test.go` won't necessarily run `updateFromV1`.

masnax: I've now also added a test that spoofs a pre-existing 3 member cluster from before this change, and then applies `updateFromV1` to it to validate that both the `schemas` and `internal_cluster_members` have the schema version column split in two.

markylaing: Yep mocking values in these tests was what I was expecting, thanks :)

internal/db/update/schema.go
```diff
@@ -66,15 +78,17 @@ func (s *SchemaUpdate) Ensure(db *sql.DB) (int, error) {
...
-			versions, err := query.SelectIntegers(ctx, tx, "SELECT version FROM schemas ORDER BY version")
+			// updateFromV1 changes the schema table and needs to be run before we calculate the schema version.
+			err := updateFromV1(ctx, tx)
 			if err != nil {
 				return err
 			}
 
-			if len(versions) > 0 {
-				current = versions[len(versions)-1]
+			versions, err = query.SelectIntegers(ctx, tx, "SELECT COALESCE(MAX(version), 0) FROM schemas WHERE type = 0 UNION ALL SELECT COALESCE(MAX(version), 0) FROM schemas WHERE type = 1")
```
markylaing, line 88: One last thing (sorry). Each time I come back to this PR I need to remember/figure out why `COALESCE(MAX(version), 0)` is required. 

Could you please define this query as a const with a description of the format of the schema table and reasoning as to why this query is the best way to extract the details we want. For example, why can't we just count the rows where `type = 0` and then count the rows where `type = 1`.

markylaing (dismissed):
Sorry one more final nit about the SQL query for getting the schema versions.

internal/db/update/schema.go
tomponline: Sounds good @markylaing - its good to improve readability for other people.

internal/db/update/schema.go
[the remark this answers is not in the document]
roosterfish: Oh sorry, my mistake.

internal/db/update/schema.go
masnax: Done.

masnax:
@markylaing I've addressed your last comment, if there's nothing else I think @tomponline can take a look now?

markylaing:
> @markylaing I've addressed your last comment, if there's nothing else I think @tomponline can take a look now?

Yeah all good thanks 🙂

tomponline (approved):

tomponline:
@masnax Thanks!

Does this require changes to applications that use microcluster?

If so please can you post in the ~MicroCloud channel drawing attention to this and perhaps update the PR description with info on how users of microcluster can update their app to work with these changes?

Thanks

PR #95, opened 2024-03-19
Use one context for cancelling the daemon

masnax:
Simplifies the daemon's blocking action in response to #93 by @neoaggelos and a closer look at the `Daemon` and `App`.

This moves the blocking loop from `app.Start` into `daemon.Run` (previously `daemon.Init`) so that we no longer have to export the related context fields. Eventually they can be removed, but this prevents us from committing to them further in any solution to the bug that #93 addresses.

Additionally, `app.Start` now handles signals with `signal.NotifyContext`, which lets us treat all triggers of the daemon stop sequence in the same way, with one context. 

Any context passed to `app.Start` will be wrapped in a `signal.NotifyContext`, and the resulting child context will be used in `daemon.Run` to block on until it cancels from a signal or some other source outside the scope of `microcluster.App`. 

In order to not call `d.stop` more than once, l took some inspiration from #93, and made the method itself a `sync.OnceValue` when we instantiate the daemon. This also lets us catch the edge case where a REST API shutdown request and a direct shutdown signal happen concurrently by having `state.Stop` also use the same `sync.Once` stop function as the daemon itself.

All in all, this reduces the blocking loop to just 2 cases:
* Waiting for the parent context to cancel (from a signal, or unknown source), which means we should run the stop sequence
* Waiting to receive from `d.shutdownDoneCh`, which should be immediate in the above case, but might also come from the REST API.

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -59,18 +59,18 @@ type Daemon struct {
 	hooks config.Hooks // Hooks to be called upon various daemon actions.
 
 	ReadyChan      chan struct{}      // Closed when the daemon is fully ready.
-	ShutdownCtx    context.Context    // Cancelled when shutdown starts.
-	ShutdownDoneCh chan error         // Receives the result of the d.Stop() function and tells the daemon to end.
-	ShutdownCancel context.CancelFunc // Cancels the shutdownCtx to indicate shutdown starting.
+	shutdownCtx    context.Context    // Cancelled when shutdown starts.
+	shutdownDoneCh chan error         // Receives the result of the d.Stop() function and tells the daemon to end.
+	shutdownCancel context.CancelFunc // Cancels the shutdownCtx to indicate shutdown starting.
 }
 
 // NewDaemon initializes the Daemon context and channels.
 func NewDaemon(ctx context.Context, project string) *Daemon {
 	ctx, cancel := context.WithCancel(ctx)
 	return &Daemon{
-		ShutdownCtx:    ctx,
-		ShutdownCancel: cancel,
-		ShutdownDoneCh: make(chan error),
+		shutdownCtx:    ctx,
+		shutdownCancel: cancel,
+		shutdownDoneCh: make(chan error),
 		ReadyChan:      make(chan struct{}),
 		project:        project,
 	}
@@ -131,14 +131,14 @@ func (d *Daemon) init(listenPort string, extendedEndpoints []rest.Endpoint, sche
 		return fmt.Errorf("Failed to initialize trust store: %w", err)
 	}
 
-	d.db = db.NewDB(d.ShutdownCtx, d.serverCert, d.ClusterCert, d.os)
+	d.db = db.NewDB(d.shutdownCtx, d.serverCert, d.ClusterCert, d.os)
 
 	// Apply extensions to API/Schema.
 	resources.ExtendedEndpoints.Endpoints = append(resources.ExtendedEndpoints.Endpoints, extendedEndpoints...)
 
 	ctlServer := d.initServer(resources.UnixEndpoints, resources.InternalEndpoints, resources.PublicEndpoints, resources.ExtendedEndpoints)
-	ctl := endpoints.NewSocket(d.ShutdownCtx, ctlServer, d.os.ControlSocket(), d.os.SocketGroup)
-	d.endpoints = endpoints.NewEndpoints(d.ShutdownCtx, ctl)
+	ctl := endpoints.NewSocket(d.shutdownCtx, ctlServer, d.os.ControlSocket(), d.os.SocketGroup)
+	d.endpoints = endpoints.NewEndpoints(d.shutdownCtx, ctl)
 	err = d.endpoints.Up()
 	if err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


internal/daemon/daemon.go
```diff
@@ -71,7 +71,7 @@ type Daemon struct {
 func NewDaemon(project string) *Daemon {
 	d := &Daemon{
 		shutdownDoneCh: make(chan error),
-		ReadyChan:      make(chan struct{}),
+		ReadyChan:      make(chan struct{}, 1),
```
tomponline, line 74: this change doesnt correlate with the commit message - did you change the wrong channel?

internal/daemon/daemon.go
```diff
@@ -124,7 +124,14 @@ func (d *Daemon) Init(ctx context.Context, listenPort string, stateDir string, s
 
 	close(d.ReadyChan)
 
-	return nil
+	for {
+		select {
+		case <-ctx.Done():
+			d.shutdownDoneCh <- d.stop()
```
tomponline, line 130: It seems a bit odd to depend on the send to the channel to return below here (although i appreciate we do need to send to the channel).

I would suggest we read do:

```go
err = d.stop()
d.shutdownDoneCh <- err
return err
```

for clarity and decouple the case where there is an err send to d.shutdownDoneCh from externally.

internal/daemon/daemon.go
```diff
@@ -566,24 +582,12 @@ func (d *Daemon) State() *state.State {
 		Database:       d.db,
 		Remotes:        d.trustStore.Remotes,
 		StartAPI:       d.StartAPI,
-		Stop:           d.Stop,
+		Stop:           d.stop,
```
tomponline, line 592: @masnax i was thinking about how we can avoid the need of exporting the `ShutdownDoneCh` in this struct, and avoid the need for the caller of `stopErr := state.Stop()` from having to pass the `stopErr` to `state.ShutdownDoneCh` as it seems rather circular.

Instead how about we do change `state.Stop` to be `func() (exit func(), stopErr error)`.

Then in `shutdownPost` we can call `exit, stopErr := state.Stop()` and then rather than posting to `state.ShutdownDoneCh` here we just call `exit()` which then internally will send the stopErr returned to `d.shutdownDoneCh` (making the channel an internal implementation detail and not part of the API).

internal/daemon/daemon.go
masnax: Hmm, looks like a rebase artifact or something. Making this channel buffered was entirely unintentional.

internal/daemon/daemon.go
```diff
@@ -579,19 +579,25 @@ func (d *Daemon) State() *state.State {
...
+		Address:     d.Address,
+		Name:        d.Name,
+		Endpoints:   d.endpoints,
+		ServerCert:  d.ServerCert,
+		ClusterCert: d.ClusterCert,
+		Database:    d.db,
+		Remotes:     d.trustStore.Remotes,
+		StartAPI:    d.StartAPI,
+		Stop: func() (exit func(), stopErr error) {
+			err := d.stop()
```
tomponline, line 594: I was thinking more that d.stop() would return the exit function.

masnax: Hm, I'm not sure about that. Technically with this approach, and as you've mentioned already in your comment [here](https://github.com/canonical/microcluster/pull/95#discussion_r1530024835), the daemon doesn't even need to directly send to the `d.shutdownDoneCh` as it can just immediately return after calling `d.stop()`. Only the `state` needs to do that in the API, so having an exit function serves no purpose for the `Daemon` itself. Plus, returning a `func()` from a `sync.OnceValues` sounds like a recipe for confusion.

masnax: We also don't need to make `d.shutdownDoneCh` buffered anymore either, since the `state` will only ever call `exit()` once, and then every other time will return early because the context is already cancelled.

internal/daemon/daemon.go
```diff
@@ -59,18 +59,18 @@ type Daemon struct {
 	hooks config.Hooks // Hooks to be called upon various daemon actions.
 
 	ReadyChan      chan struct{}      // Closed when the daemon is fully ready.
-	ShutdownCtx    context.Context    // Cancelled when shutdown starts.
-	ShutdownDoneCh chan error         // Receives the result of the d.Stop() function and tells the daemon to end.
-	ShutdownCancel context.CancelFunc // Cancels the shutdownCtx to indicate shutdown starting.
+	shutdownCtx    context.Context    // Cancelled when shutdown starts.
+	shutdownDoneCh chan error         // Receives the result of the d.Stop() function and tells the daemon to end.
```
tomponline, line 63: Comment mentions d.Stop that is now d.stop

microcluster/app.go
```diff
@@ -82,15 +82,12 @@ func (m *MicroCluster) Start(apiEndpoints []rest.Endpoint, schemaExtensions map[
...
-	sigCh := make(chan os.Signal, 1)
-	signal.Notify(sigCh, unix.SIGPWR)
-	signal.Notify(sigCh, unix.SIGINT)
-	signal.Notify(sigCh, unix.SIGQUIT)
-	signal.Notify(sigCh, unix.SIGTERM)
-
 	chIgnore := make(chan os.Signal, 1)
 	signal.Notify(chIgnore, unix.SIGHUP)
 
+	ctx, cancel := signal.NotifyContext(m.ctx, unix.SIGPWR, unix.SIGTERM, unix.SIGINT, unix.SIGQUIT)
```
roosterfish, line 88: Neat, wasn't aware of `NotifyContext` until now.

internal/daemon/daemon.go
roosterfish: To add onto this isn't the `stopErr` var now completely ignored?

internal/daemon/daemon.go
tomponline: Not addressed

masnax: I added it to the commit which introduces the `exit()` function as that's the only place this is actually used now.

masnax: I've updated it in both commits now.

tomponline (approved):

neoaggelos (approved):
Tested locally, this works for us and can replace #93. Thanks a lot @masnax, when can we expect this to land?

PR #103, opened 2024-04-10
trust: expose trust.Location type

neoaggelos:
### Summary

Expose `trust.Location` type so that extended handlers can use `state.StartAPI()`

This is similar to https://github.com/canonical/microcluster/blob/0dcc76f5a09c8288da6a5aebe31f4c8e00d26a5a/state/state.go#L6

### DIFF
--- a/trust/trust.go
+++ b/trust/trust.go
@@ -1,0 +1,6 @@
+package trust
+
+import "github.com/canonical/microcluster/internal/trust"
+
+// Location exposes the internal trust location type for use with extended API handlers.
+type Location = trust.Location


tomponline:
Please can you advise what is the reason for the change?
I'd like to understand why microk8s needs this but the other micro* products (apparently) don't?

Thanks

markylaing:
`StartAPI` is called when a microcluster node successfully bootstraps the database or successfully joins an existing cluster. This is handled by microcluster itself, so extended handlers shouldn't need to call it. What led to the requirement for calling `StartAPI` from an extended handler?

neoaggelos:
The original problem we encounter is that our bootstrap process sometimes fails with a `context deadline exceeded` error during `.NewCluster()` with `timeout=0`. The trace looks like this:

- https://github.com/canonical/microcluster/blob/b39c0e97e8e47001559e5b8da86c948bedebf60d/internal/rest/client/control.go#L13
- https://github.com/canonical/microcluster/blob/00e59696646f6e69de068a0b0557750d49f488aa/internal/rest/client/client.go#L308
- https://github.com/canonical/microcluster/blob/00e59696646f6e69de068a0b0557750d49f488aa/internal/rest/client/client.go#L208-L214

Our requirements (wrt parsing bootstrap configuration, applying configs, etc) necessitate that we use a custom API for bootstrapping, and the handler in turn calls `microclusterapp.NewCluster()`. This always gets an implicit 30 second timeout that we cannot opt out from. One way to approach this, other than dropping this 30 second timeout in microcluster (which might break lots of downstream assumptions), is to instead call `state.StartAPI()` directly, since we are already in the server (which is also what a `.NewCluster()` bootstrap also does: https://github.com/canonical/microcluster/blob/10c6f6779a0b7a41d03d84fc942f29d5cb8a4145/internal/rest/resources/control.go#L29-L53

However, it is not possible to call `state.StartAPI()` in https://github.com/canonical/k8s-snap/blob/6178860a0b9fbab8453693868f35571c40c39dfa/src/k8s/pkg/k8sd/api/cluster_bootstrap.go#L13-L42, since the type definition includes an argument with the internal `trust.Location` type.

Hence this PR. Without this, even though `state.StartAPI()` is exposed, it is effectively unusable from the extended handlers.

tomponline:
> Our requirements (wrt parsing bootstrap configuration, applying configs, etc) necessitate that we use a custom API for bootstrapping.

It sounds like we should discuss this further and then come up with a plan on how to proceed, and whether this needs changes in microcluster.

@masnax might be able to offer suggestions if the requirements could be explained further? Thanks

markylaing:
> Our requirements (wrt parsing bootstrap configuration, applying configs, etc) necessitate that we use a custom API for bootstrapping, and the handler in turn calls microclusterapp.NewCluster(). This always gets an implicit 30 second timeout that we cannot opt out from.

We added support for passing configuration into the bootstrap process in this PR: https://github.com/canonical/microcluster/pull/72. These config keys are passed into the bootstrap hook for use by consuming applications.

With respect to the timeout, I think would should change `microclusterapp.NewCluster()` to accept a context, if no deadline is set, the default deadline (30s) can be applied.

tomponline:
@masnax @neoaggelos shall we close this for now?

PR #104, opened 2024-04-10
Daemon: Remove redundant call to `AddClusterMember`

roosterfish:
When joining a new node to the cluster this action is already performed as part of the `joinWithToken` function on the control endpoint: https://github.com/canonical/microcluster/blob/main/internal/rest/resources/control.go#L104.

If the first node gets bootstrapped, trust for itself is added as part of `StartAPI` here https://github.com/canonical/microcluster/blob/main/internal/daemon/daemon.go#L374.

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -29,7 +29,6 @@ import (
 	internalREST "github.com/canonical/microcluster/internal/rest"
 	internalClient "github.com/canonical/microcluster/internal/rest/client"
 	"github.com/canonical/microcluster/internal/rest/resources"
-	internalTypes "github.com/canonical/microcluster/internal/rest/types"
 	"github.com/canonical/microcluster/internal/state"
 	"github.com/canonical/microcluster/internal/sys"
 	"github.com/canonical/microcluster/internal/trust"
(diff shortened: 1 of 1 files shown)


roosterfish:
@masnax opening this as a draft PR because I am not sure if I am missing something else here.

masnax:
Ah yes, I was wondering when someone would raise an eyebrow at this :)

This is part of what's going to be addressed by https://github.com/canonical/microcluster/pull/89

Before we had so many hooks `PreJoin`, `PostJoin`, `OnNewMember`, etc., the main pattern I had used to actually run the hooks was to set the cluster notification header and re-use the same endpoint, but handle it differently.

The cluster notification header is set with the final argument [here](https://github.com/canonical/microcluster/blob/bcb26eb804fd991268760efce59e694194fa74fe/internal/daemon/daemon.go#L457) which applies [this function](https://github.com/canonical/microcluster/blob/bcb26eb804fd991268760efce59e694194fa74fe/internal/rest/client/client.go#L180-L197) as the proxy function for the client. This is sort of hinted at with the comment [here](https://github.com/canonical/microcluster/blob/bcb26eb804fd991268760efce59e694194fa74fe/internal/daemon/daemon.go#L472).


Then, when we call `AddClusterMember`, we forego the default actions and instead run [this block](https://github.com/canonical/microcluster/blob/bcb26eb804fd991268760efce59e694194fa74fe/internal/rest/resources/cluster.go#L53-L72) which executes the `OnNewMember` hook.

masnax:
The final solution in #89 isn't ideal either, and I've been debating redoing that PR and just moving all the hooks over to their own set of endpoints to make things clear and distinct.

roosterfish:
Appreciate the detailed response, so with the change in this PR it doesn't anymore send around the notification so it won't run the `OnNewMemberHook(s)` on the other cluster members? 

This explains why I could still form a functioning cluster, but this is breaking the hooks then, haven't tested this.
 
I'll close it and keep an eye on https://github.com/canonical/microcluster/pull/89. Let me know when you need a review.

roosterfish:
> This is sort of hinted at with the comment [here](https://github.com/canonical/microcluster/blob/bcb26eb804fd991268760efce59e694194fa74fe/internal/daemon/daemon.go#L472).

Yeah I saw this but got confused by the "upgrade" in there.

PR #105, opened 2024-04-15
Specify context in all microcluster methods

neoaggelos:
### Summary

- Remove context from `*microcluster.MicroCluster` struct
- Pass context during `(*MicroCluster).Start`, not during creation
- Update all `(*MicroCluster)` methods to accept a context and use that instead
- Update `microctl` for new API, use `cmd.Context()` instead of `context.Background()`
- Update `microd` for new API, use `cmd.Context()` instead of `context.Background()`

### DIFF
--- a/example/cmd/microctl/waitready.go
+++ b/example/cmd/microctl/waitready.go
@@ -1,6 +1,9 @@
 package main
 
 import (
+	"context"
+	"time"
+
 	"github.com/canonical/microcluster/microcluster"
 	"github.com/spf13/cobra"
 )
@@ -33,5 +36,11 @@ func (c *cmdWaitready) Run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	return m.Ready(cmd.Context(), c.flagTimeout)
+	ctx, cancel := cmd.Context(), func() {}
+	if c.flagTimeout > 0 {
+		ctx, cancel = context.WithTimeout(ctx, time.Duration(c.flagTimeout)*time.Second)
+	}
+	defer cancel()
+
+	return m.Ready(ctx)
 }
--- a/example/cmd/microctl/waitready.go
+++ b/example/cmd/microctl/waitready.go
@@ -1,8 +1,6 @@
 package main
 
 import (
-	"context"
-
 	"github.com/canonical/microcluster/microcluster"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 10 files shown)


masnax (approved):
Excellent, thanks so much for this: )

microcluster/app.go
```diff
@@ -117,7 +115,7 @@ func (m *MicroCluster) Status() (*internalTypes.Server, error) {
 
 // Ready waits for the daemon to report it has finished initial setup and is ready to be bootstrapped or join an
 // existing cluster.
-func (m *MicroCluster) Ready(timeoutSeconds int) error {
+func (m *MicroCluster) Ready(ctx context.Context, timeoutSeconds int) error {
```
roosterfish, line 118: Isn't the `timeoutSeconds` variable now redundant as we can potentially pass a `context.WithTimeout()`?

roosterfish (commented):
Thanks, just one remark on the `Ready()` function.

microcluster/app.go
tomponline: yeah this makes sense

neoaggelos: yes it is, same as in other methods like `NewCluster` and `JoinCluster`. Should I remove it on all?

neoaggelos: OK, I've updated it and pushed the changes

roosterfish (dismissed):
LGTM!

example/cmd/microctl/waitready.go
```diff
@@ -18,7 +21,7 @@ func (c *cmdWaitready) Command() *cobra.Command {
 		RunE:  c.Run,
 	}
 
-	cmd.Flags().IntVarP(&c.flagTimeout, "timeout", "t", 0, "Number of seconds to wait before giving up"+"``")
+	cmd.Flags().DurationVarP(&c.flagTimeout, "timeout", "t", 0, "Number of seconds to wait before giving up"+"``")
```
markylaing, line 24: Changing this to a duration means it is no longer a number of seconds? Presumably the user needs to specify something like "10s" instead? I'd leave it as an int and multiply `c.flatTimeout` by `time.Second` below.

markylaing (changes requested):
Just one nit :)

example/cmd/microctl/waitready.go
```diff
@@ -36,7 +36,10 @@ func (c *cmdWaitready) Run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	ctx, cancel := context.WithTimeout(cmd.Context(), c.flagTimeout)
+	ctx, cancel := cmd.Context(), func() {}
+	if c.flagTimeout > 0 {
+		ctx, cancel = context.WithTimeout(ctx, time.Duration(c.flagTimeout)*time.Second)
+	}
```
markylaing, line 42: Could you please amend the previous commit rather than adding a new one? Thanks

markylaing (dismissed):
Hi @neoaggelos sorry to nitpick, we try to keep a linear git history to make cherry-picking easier when maintaining snap channels. Can you please amend the commit which changed the flag to a duration rather than adding a new one? Thanks.

roosterfish (approved):
I understand the commit for changing the flag's type is gone and instead you went with the approach suggested by @markylaing.

masnax (approved):
Overall, this looks good to me so I'll be merging it now.

PR #107, opened 2024-04-16
Run PreRemove hook before modifying the database

masnax:
The PreRemove hook was intended to be used before the node was removed from the database, however it seems it was incorrectly only being run before the node was removed from dqlite.

This poses some issues because an application using microcluster may want to clean up entities associated to records in the database for a particular node, but by removing the node from the database first, `ON DELETE CASCADE` will clear all of that state.

### DIFF
--- a/internal/rest/resources/cluster.go
+++ b/internal/rest/resources/cluster.go
@@ -353,7 +353,7 @@ func clusterMemberDelete(s *state.State, r *http.Request) response.Response {
 		return response.SmartError(err)
 	}
 
-	// If we are not the leader, just update our trust store and forward the request.
+	// If we are not the leader, just forward the request.
 	if leaderInfo.Address != s.Address().URL.Host {
 		if allRemotes[name].Address.String() == s.Address().URL.Host {
 			// If the member being removed is ourselves and we are not the leader, then lock the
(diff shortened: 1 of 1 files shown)


internal/rest/resources/cluster.go
```diff
@@ -554,6 +546,14 @@ func clusterMemberDelete(s *state.State, r *http.Request) response.Response {
 		return response.SmartError(err)
 	}
 
+	// Remove the cluster member from the database.
```
roosterfish, line 536: Is there a reason `ResetClusterMember()` is called twice? Once before removing the cluster member from the DB and a second time after removing it from the trust store down below.

Would this mean the `PreRemove` hook is called a second time?

masnax: It's the same as `AddClusterMember`, based on the cluster notification request header it either runs the hook or the default implementation.

roosterfish: Ah I see, thanks!

masnax:
I want to wait to merge this until after we have API updates in place, as this does affect the API.

markylaing:
> > I want to wait to merge this until after we have API updates in place, as this does affect the API.
> 
> Which PR is this dependent on?

I'm guessing #86, so an API extension can be added for it.

masnax:
Yes that was the case with all of the outstanding PRs since many were modifications to the API. 

But since @tomponline merged a bunch of them already I suppose it doesn't matter anymore. We can merge this as is.

tomponline:
> But since @tomponline merged a bunch of them already I suppose it doesn't matter anymore. We can merge this as is.

Well lets wait a moment - if the previous PRs I merged that were approved also need API extensions then lets also add them too.

tomponline:
Its getting very hard to understand which PRs related to each other as so many are backed up.

PR #110, opened 2024-04-18
example: Add system tests for microd/microctl

MggMuggins:
Just a duplicate of what's in the readme so that we have a convenient way to spin up a cluster when hacking on microcloud.


example/test/main.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
roosterfish, line 1: It looks to be almost the same as the test added as part of https://github.com/canonical/microcluster/pull/106. 

I am wondering if it would make sense having the test case one instead as it deploys but also runs some checks? 
I have just recently learned about `LXD_INSPECT` in the LXD test suite.
We could introduce something like this so that the deployed MicroCluster can potentially be reused?

roosterfish (changes requested):

example/test/main.sh
MggMuggins: Yeah, I was planning to rebase #106 onto this but haven't pushed yet.

I'll add `LXD_INSPECT` this afternoon.

MggMuggins: Totally agree, thanks!

roosterfish (approved):
LGTM!

tomponline (approved):

PR #111, opened 2024-04-20
Clearer errors and logs

masnax:
* Makes the cluster join token error more descriptive to help debug join errors.
* Suppresses truststore initialization error, as we won't have any entries when the daemon first starts.

### DIFF
--- a/internal/trust/remotes.go
+++ b/internal/trust/remotes.go
@@ -73,7 +73,9 @@ func (r *Remotes) Load(dir string) error {
 		remoteData[remote.Name] = *remote
 	}
 
-	if len(remoteData) == 0 {
+	// If the refreshed truststore data is empty, and we already had data in the truststore,
+	// abort the refresh because an initialized system should always have truststore entries.
+	if len(remoteData) == 0 && len(r.data) != 0 {
 		logger.Warn("Failed to parse new remotes from truststore")
 
 		return nil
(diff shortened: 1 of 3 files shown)


internal/trust/remotes.go
```diff
@@ -73,7 +73,7 @@ func (r *Remotes) Load(dir string) error {
 		remoteData[remote.Name] = *remote
 	}
 
-	if len(remoteData) == 0 {
+	if len(remoteData) == 0 && len(r.data) != 0 {
```
roosterfish, line 76: In which situation would this condition be true? I understand in case of init `remoteData` is empty but what would be the effect of also checking `len(r.data) != 0`? Is it checking for the existence of the nodes own certificate? 
I would appreciate a comment here :)

roosterfish (changes requested):

internal/trust/remotes.go
masnax: We try to load the truststore as soon as the daemon starts, but it won't contain anything until the node is bootstrapped. 

This results in a bunch of logs being sent because the truststore was initialized/refreshed, but without any content (because there never was any to begin with).

So this instead makes sure that the truststore actually had content before, and only propagates the log and abort if the truststore refresh causes all entries to be lost. 

I'll add a comment too.

roosterfish: Thanks clear now.

roosterfish (approved):
Thanks for adding the comment, looks good to me.

tomponline (approved):

PR #112, opened 2024-04-20
Allow unauthenticated connections to /cluster/1.0

masnax:
Untrusted systems should still be able to view this non-sensitive information. (name, address, database status, and soon likely API extensions).

### DIFF
--- a/internal/rest/resources/api_1.0.go
+++ b/internal/rest/resources/api_1.0.go
@@ -5,7 +5,6 @@ import (
 
 	"github.com/canonical/lxd/lxd/response"
 
-	"github.com/canonical/microcluster/internal/rest/access"
 	internalTypes "github.com/canonical/microcluster/internal/rest/types"
 	"github.com/canonical/microcluster/internal/state"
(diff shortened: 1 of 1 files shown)


tomponline:
@masnax why do we need this?

tomponline:
>  Untrusted systems should still be able to view this non-sensitive information. (name, address, database status, and soon likely API extensions).

This doesn't explain what this PR fixes?

masnax:
> > Untrusted systems should still be able to view this non-sensitive information. (name, address, database status, and soon likely API extensions).
> 
> This doesn't explain what this PR fixes?

This is just server information about the cluster. Like LXD's `GetServer` on `/1.0`.

It was restricted to cluster members but all of this information is already available to all cluster members locally.

Nothing here is that sensitive so I thought it would be more flexible to allow untrusted systems to check it, so they can see if the cluster is functional or not. 

In LXD sensitive information is appended to the returned data only when the connection is from a trusted source, so I thought we could do something similar if we ever do want to add sensitive information here.

PR #113, opened 2024-04-20
Add truststore API

masnax:
This PR cleans up some of the patchwork in the join API.

#### `POST /cluster/internal/truststore`
* Adds an entry to the truststore of each node in the cluster. Accepts a `types.ClusterMemberLocal` which is usually used to convert to a `trust.Remote` type for the truststore.

#### `DELETE /cluster/internal/truststore`
* Replaces the truststore with the given entry excluded on each node of the cluster. Also accepts a `types.ClusterMemberLocal`.

#### Refactor local truststore management on joins
* Current behaviour 
  * Only the dqlite leader, and whichever node forwarded the request to the dqlite leader trust a joining node as it goes through the join process. The rest of the cluster is only updated 30 seconds later on the next heartbeat. 
* New behaviour
  * Just the dqlite leader will trust the joining node until it completes the join process. Finally, the joining node will inform the leader that it is done, at which point the leader will update the trust stores of all other nodes to include the joiner using the `/cluster/internal/truststore` endpoint.

#### Cleans up confusing functions for using cluster notifications.
  * Adds a `SetClusterNotification` function to the client for applying it automatically to all requests made by the client.
  * Renames `IsForwardedRequest` to `IsNotification` to be more consistent.
  * Changes `state.Cluster`, which returns a set of clients for all nodes in dqlite except the local one, to take a boolean flag as argument for setting the cluster notification header, as opposed to a request.
  
#### Adds non-dqlite version of `state.Cluster`
  * In case we want to exclusively talk to systems in the local trust store, this adds a `Cluster` method to `trust.Remotes` to fetch a client for each `Remote`. This will be used to reach the node that accepted a joiner into the cluster, so it can confirm that the cluster has successfully joined.
  
#### Simplify `StartAPI` using helpers.
* Using the `remote.Cluster` helper defined above, as well as moving the schema upgrade notification handling to its own helper reduces some of the complexity of `StartAPI` and makes it a bit easier to read.

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -445,26 +445,17 @@ func (d *Daemon) StartAPI(bootstrap bool, initConfig map[string]string, newConfi
 	}
 
 	// Get a client for every other cluster member in the newly refreshed local store.
-	cluster := make(client.Cluster, 0, d.trustStore.Remotes().Count()-1)
-	for _, addr := range d.trustStore.Remotes().Addresses() {
-		if d.address.URL.Host == addr.String() {
-			continue
-		}
-
-		publicKey, err := d.ClusterCert().PublicKeyX509()
-		if err != nil {
-			return err
-		}
-
-		url := api.NewURL().Scheme("https").Host(addr.String())
-		c, err := internalClient.New(*url, d.ServerCert(), publicKey, true)
-		if err != nil {
-			return err
-		}
-
-		cluster = append(cluster, client.Client{Client: *c})
+	publicKey, err := d.ClusterCert().PublicKeyX509()
+	if err != nil {
+		return err
 	}
 
+	cluster, err := d.trustStore.Remotes().Cluster(true, d.ServerCert(), publicKey)
+	if err != nil {
+		return err
+	}
+
+	localMemberInfo := internalTypes.ClusterMemberLocal{Name: localNode.Name, Address: localNode.Address, Certificate: localNode.Certificate}
 	if len(joinAddresses) > 0 {
 		err = d.hooks.PreJoin(d.State(), initConfig)
 		if err != nil {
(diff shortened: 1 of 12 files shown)


masnax:
I do plan to add some tests for the new endpoint as well.

internal/state/state.go
```diff
@@ -78,12 +76,8 @@ var OnNewMemberHook func(state *State) error
 var ReloadClusterCert func() error
 
 // Cluster returns a client for every member of a cluster, except
-// this one, with the UserAgentNotifier header set if a request is given.
-func (s *State) Cluster(r *http.Request) (client.Cluster, error) {
-	if r != nil {
-		r.Header.Set("User-Agent", request.UserAgentNotifier)
-	}
-
+// this one, with the UserAgentNotifier header set if isNotification is true.
```
roosterfish, line 79: As the header isn't set anymore in this function should the second line of the comment be removed?

internal/rest/client/truststore.go
```diff
@@ -0,0 +1,25 @@
...
+// AddTrustStoreEntry adds a new record to the truststore on all cluster members.
+func AddTrustStoreEntry(ctx context.Context, c *Client, args types.ClusterMemberLocal) error {
+	queryCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	defer cancel()
+
+	return c.QueryStruct(queryCtx, "POST", InternalEndpoint, api.NewURL().Path("truststore"), args, nil)
+}
+
+// DeleteTrustStoreEntry deletes the record corresponding to the given cluster member from the trust store.
+func DeleteTrustStoreEntry(ctx context.Context, c *Client, args types.ClusterMemberLocal) error {
```
roosterfish, line 20: Not sure about the other MicroCluster endpoints in regards to consistency, but would it make sense to have a `DELETE /cluster/internal/truststore/<member>` endpoint instead?

roosterfish: I am thinking about the `PUT/DELETE /cluster/1.0/cluster/<member>` endpoints that use the same concept.

internal/daemon/daemon.go
```diff
@@ -463,8 +463,38 @@ func (d *Daemon) StartAPI(bootstrap bool, initConfig map[string]string, newConfi
 		}
 	}
 
+	var lastErr error
+	var clusterConfirmation bool
+	err = cluster.Query(d.shutdownCtx, true, func(ctx context.Context, c *client.Client) error {
+		// No need to send a request to ourselves.
+		if d.address.URL.Host == c.URL().URL.Host {
+			return nil
+		}
```
roosterfish, line 472: Missing break after the `if`?

roosterfish (changes requested):
Thanks for making the concept of the cluster notification more clear in regards to the truststore updates. Just a few small comments from my side.

internal/state/state.go
masnax: It's not set in the function but it returns a client that applies it to all requests it makes. I'll clarify the comment.

internal/rest/client/truststore.go
masnax: Done.

roosterfish (approved):
LGTM!

PR #114, opened 2024-04-20
Add hooks API

### LINKED ISSUE #109: Uncouple hooks and truststore updates from other endpoints
Early in microcluster's development, we didn't want to have separate endpoints for managing hooks, but since then the number of hooks has tripled, so it's not feasible to re-use existing endpoints for this purpose, and makes things very confusing.

Instead we should have a new endpoint:

`/cluster/internal/hooks` 

I propose adding a `POST` method on this endpoint that takes a query parameter that iterates over constant integers corresponding to each hook. Then the endpoint can run the hook corresponding to the query parameter, and parse any payload that is necessary to pass into the hook.

Additionally, we also hide truststore updates in some endpoints, namely `POST` and `DELETE` `/cluster/1.0/cluster`. To make things clearer, we can add a new endpoint:

`/cluster/internal/truststore`

with two methods: 
`POST`: adds a new entry to the truststore on the given node.
`DELETE`: deletes an entry in the truststore on the given node.

masnax:
Closes #109 

Depends on #113 

#### `POST /cluster/internal/hooks/{hookType}`
 * Adds an endpoint to execute multi-node hooks on a particular cluster member. The hook type is specified by a `types.HookType` integer, and the arguments of the hook are passed in with the `types.HookNewMemberOptions` struct and the `types.HookRemoveMemberOptions` struct. This replaces the current implementation which relies on cluster notifications, which is confusing and bloats the cluster management API.

 * The only multi-node hooks are currently `OnNewMember`, `PreRemove`, and `PostRemove`, so only these are supported by the API at this time. The other hooks are local to a particular node and so do not need to be run by a different node over the network.
 
#### `/cluster/1.0/cluster` is disallowed if not initialized.
* The cluster management network requires the database, and so the database must be open for the API to work. It doesn't make sense to allow this the cluster management API to work before initialization.

* The `OnNewMember` hook might run on some nodes before they have started their database. This shouldn't be allowed as we don't know where the node is in the setup process yet.

#### Cluster notifications on `/cluster/1.0/cluster` do not execute hooks anymore
* As the hooks have their own endpoint, we should just use that instead of running the hooks as a cluster notification on existing APIs, which makes it hard to read the code when we see client functions run twice with different outcomes.

### DIFF
--- a/internal/rest/resources/cluster.go
+++ b/internal/rest/resources/cluster.go
@@ -49,29 +49,6 @@ var clusterMemberCmd = rest.Endpoint{
 }
 
 func clusterPost(s *state.State, r *http.Request) response.Response {
-	// If we received a forwarded request, assume the new member was successfully added on the leader,
-	// and execute the new member hook.
-	if client.IsNotification(r) {
-		ctx, cancel := context.WithTimeout(s.Context, 30*time.Second)
-		defer cancel()
-
-		// Wait for the database to be set up in case we received this request at the same time as joining ourselves.
-		for !s.Database.IsOpen() {
-			select {
-			case <-ctx.Done():
-				return response.SmartError(fmt.Errorf("Error waiting for peer to initialize: %w", ctx.Err()))
-			default:
-			}
-		}
-
-		err := state.OnNewMemberHook(s)
-		if err != nil {
-			return response.SmartError(fmt.Errorf("Failed to run post cluster member add actions: %w", err))
-		}
-
-		return response.EmptySyncResponse
-	}
-
 	if !s.Database.IsOpen() {
 		return response.Unavailable(fmt.Errorf("Daemon not yet initialized"))
 	}
@@ -254,18 +231,6 @@ var clusterDisableMu sync.Mutex
 
 // Re-execs the daemon of the cluster member with a fresh s.
 func clusterMemberPut(s *state.State, r *http.Request) response.Response {
-	force := r.URL.Query().Get("force") == "1"
-
-	// If we received a cluster notification, run the pre-removal hook and return.
-	if client.IsNotification(r) {
-		err := state.PreRemoveHook(s, force)
-		if err != nil {
-			return response.SmartError(fmt.Errorf("Failed to run pre-removal hook: %w", err))
-		}
-
-		return response.EmptySyncResponse
-	}
-
 	err := s.Database.Stop()
 	if err != nil {
 		return response.SmartError(fmt.Errorf("Failed shutting down database: %w", err))
@@ -327,17 +292,6 @@ func clusterMemberDelete(s *state.State, r *http.Request) response.Response {
 	name, err := url.PathUnescape(mux.Vars(r)["name"])
 	if err != nil {
 		return response.SmartError(err)
-	}
-
-	// If we received a forwarded request, assume the new member was successfully removed on the leader,
-	// and execute the post-remove hook.
-	if client.IsNotification(r) {
-		err := state.PostRemoveHook(s, force)
-		if err != nil {
-			return response.SmartError(fmt.Errorf("Failed to run post cluster member remove actions: %w", err))
-		}
-
-		return response.EmptySyncResponse
 	}
 
 	allRemotes := s.Remotes().RemotesByName()
@@ -534,7 +488,7 @@ func clusterMemberDelete(s *state.State, r *http.Request) response.Response {
 	}
 
 	// Tell the cluster member to run its PreRemove hook and return.
-	err = c.ResetClusterMember(s.Context, name, force)
+	err = internalClient.RunPreRemoveHook(ctx, c.UseTarget(name), internalTypes.HookRemoveMemberOptions{Force: force})
 	if err != nil {
 		return response.SmartError(err)
 	}
@@ -590,14 +544,27 @@ func clusterMemberDelete(s *state.State, r *http.Request) response.Response {
 		return response.SmartError(err)
 	}
 
+	// Run the PostRemove hook locally.
 	err = state.PostRemoveHook(s, force)
 	if err != nil {
 		return response.SmartError(err)
 	}
 
+	// Run the PostRemove hook on all other members.
+	remotes := s.Remotes()
 	err = cluster.Query(s.Context, true, func(ctx context.Context, c *client.Client) error {
 		c.SetClusterNotification()
-		return c.DeleteClusterMember(ctx, name, force)
+		addrPort, err := types.ParseAddrPort(c.URL().URL.Host)
+		if err != nil {
+			return err
+		}
+
+		remote := remotes.RemoteByAddress(addrPort)
+		if remote == nil {
+			return fmt.Errorf("No remote found at address %q run the post-remove hook", c.URL().URL.Host)
+		}
+
+		return internalClient.RunPostRemoveHook(ctx, c.Client.UseTarget(remote.Name), internalTypes.HookRemoveMemberOptions{Force: force})
 	})
 	if err != nil {
 		return response.SmartError(err)
--- a/internal/rest/resources/cluster.go
+++ b/internal/rest/resources/cluster.go
@@ -34,8 +34,7 @@ import (
 )
 
 var clusterCmd = rest.Endpoint{
-	Path:              "cluster",
-	AllowedBeforeInit: true,
+	Path: "cluster",
 
 	Post: rest.EndpointAction{Handler: clusterPost, AllowUntrusted: true},
 	Get:  rest.EndpointAction{Handler: clusterGet, AccessHandler: access.AllowAuthenticated},
@@ -49,10 +48,6 @@ var clusterMemberCmd = rest.Endpoint{
 }
 
 func clusterPost(s *state.State, r *http.Request) response.Response {
-	if !s.Database.IsOpen() {
-		return response.Unavailable(fmt.Errorf("Daemon not yet initialized"))
-	}
-
 	req := internalTypes.ClusterMember{}
 
 	// Parse the request.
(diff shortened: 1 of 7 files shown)


masnax:
This PR will also get some tests added to it.

tomponline:
needs a rebase

markylaing:
> POST /cluster/internal/hooks/{clusterMember}?hookType={int}
Adds an endpoint to execute multi-node hooks on a particular cluster member. The hook type is specified by a types.HookType integer, and the arguments of the hook are passed in with the types.HookConfig struct. This replaces the current implementation which relies on cluster notifications, which is confusing and bloats the cluster management API.
>
>The only multi-node hooks are currently OnNewMember and PostRemove, so only these are supported by the API at this time. Thqe other hooks are local to a particular node and so do not need to be run by a different node over the network.

I realise this is for internal use, but it might be clearer to have `/cluster/internal/hooks/{clusterMember}/{hookName}` and have names instead of integer types.

tomponline:
> I realise this is for internal use, but it might be clearer to have `/cluster/internal/hooks/{clusterMember}/{hookName}` and have names instead of integer types.

Agreed, easier to debug in logs.

masnax:
@markylaing @tomponline @roosterfish This is ready for review now, thanks.

masnax:
@markylaing @roosterfish @tomponline pinging again as this one is still waiting on reviews, thanks.

internal/rest/resources/hooks.go
```diff
@@ -0,0 +1,89 @@
...
+
+	"github.com/canonical/microcluster/internal/rest/access"
+	"github.com/canonical/microcluster/internal/rest/client"
+	"github.com/canonical/microcluster/internal/rest/types"
+	"github.com/canonical/microcluster/internal/state"
+	"github.com/canonical/microcluster/rest"
+)
+
+var hooksCmd = rest.Endpoint{
+	Path: "hooks/{clusterMember}/{hookType}",
```
roosterfish, line 20: Is there a requirement for having the `clusterMember` in the URL? 
Can't we call `hooks/{hookType}` on every member of the cluster directly if we already know the name (`clusterMember`) on which the hook should be executed? 

Below you do `if s.Name() != name {}` so do we require this "hidden" forwarding of the request if we could just call out to the member directly?

internal/rest/resources/cluster.go
```diff
@@ -534,7 +488,7 @@ func clusterMemberDelete(s *state.State, r *http.Request) response.Response {
 	}
 
 	// Tell the cluster member to run its PreRemove hook and return.
-	err = c.ResetClusterMember(s.Context, name, force)
+	err = internalClient.RunHook(ctx, c, name, internalTypes.PreRemove, internalTypes.HookConfig{Force: force})
```
roosterfish, line 482: Much easier to follow along now than having "reset" in the name :)

internal/rest/resources/cluster.go
```diff
@@ -590,14 +544,27 @@ func clusterMemberDelete(s *state.State, r *http.Request) response.Response {
...
+		if err != nil {
+			return err
+		}
+
+		remote := remotes.RemoteByAddress(addrPort)
+		if remote == nil {
+			return fmt.Errorf("No remote found at address %q run the post-remove hook", c.URL().URL.Host)
+		}
+
+		return internalClient.RunHook(ctx, &c.Client, remote.Name, internalTypes.PostRemove, internalTypes.HookConfig{Force: force})
```
roosterfish, line 558: As this is now calling the hook on each member, is there a situation where the actual API endpoint needs to forward the request to another member? Please see my comment from before for reference.

roosterfish (changes requested):

internal/rest/types/hooks.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+// HookConfig holds configuration pertaining to different hooks.
+type HookConfig struct {
+	// Force represents whether or not the hook should consider validation.
+	Force bool `json:"force" yaml:"force"`
+
+	// Name represents the name of the node that triggered the hook.
+	// This is distinct from the node that the hook is triggered on.
+	Name string `json:"name" yaml:"name"`
```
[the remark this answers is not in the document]
markylaing: Or perhaps `CallerNodeName`, `CallerMemberName`

internal/rest/resources/hooks.go
masnax: That's too restrictive. It boxes us into an "all or nothing" approach. For example, `PreRemove` only runs on the to-be-removed node, but can be triggered remotely by any node that received the request to remove that node.

masnax: Ah I think I misunderstood you, did you mean we should drop the name and the forwarding and just implicitly run the hook on the addresses we care about?

I was mainly including the node name to make it clear that POSTing to this API is node-specific, but if this isn't helpful then yes we can remove it. 

Actually the forwarding should be removed in general because microcluster has built-in forwarding that I forgot about. So if we do keep the node name we would just use it for validation and error out if it doesn't match.

roosterfish: Jup that was the idea after I saw there is a loop calling out to each of the members directly. So I was wondering if there even is a situation in which we would call this endpoint on a specific member but the request would be forwarded?

masnax: The forwarding is definitely unnecessary because microcluster actually has by-name forwarding built into every endpoint with a flag on the EndpointHandler. 

So if we do keep the name, it would just be to validate that we are on the right system and otherwise error out.

masnax: It's superfluous right now because the API is minimal, but we should be ready for future changes and make them easy to implement. For example, how would we reconcile a new cluster-wide hooks API with the existing one?

Right now, we need to do this:
```
// Run the hook on only these relevant cluster members. This endpoint only accepts PreRemove, PostRemove, OnNewMember.
for member in releventClusterMembers {
  POST /cluster/internal/hooks/{member}/{hookType}
}
```

But what if in the future we want to do this:
```
 // Run certain hooks on all cluster members. This endpoint only accepts cluster-wide hooks, hypothetically like OnHeartbeat, OnDqliteRoleChange.
 POST /cluster/internal/hooks/{hookType}
```

internal/rest/types/hooks.go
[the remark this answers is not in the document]
masnax: Issue with an approach like that is that the client function would have to accept `any` as the type for the payload and we would have to validate the payload either in the client function or on the server side, or both.

internal/rest/resources/hooks.go
masnax: Another reason we might want to reserve the top level `hooks` API is if we ever want to have state for the hooks, like turning them off or on, and preloading data to be used with them.

internal/rest/types/hooks.go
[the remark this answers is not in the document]
roosterfish: What about returning an interface from the client function? It could have the function `Triggerer() string` so for each specific hook type we can set special implementation on how to retrieve those details from the underlying data structure? And there could be a `Type() string` which returns the name of the hook so that it can be casted into any of the underlying structs if necessary?

internal/rest/resources/hooks.go
roosterfish: If we just have the `POST /cluster/internal/hooks/{hookType}` endpoint that triggers `hookType` only on the member where the endpoint is called, would we need any more in the future?

Having this single endpoint we can run the loop you have put as an example for each of the relevant cluster members since we also know it's address:

```
for member in releventClusterMembers {
  POST {address of relevant member}/cluster/internal/hooks/{hookType}
}
```

Do I miss anything here? Sorry if I am not following along with what you are describing.

internal/rest/types/hooks.go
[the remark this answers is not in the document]
masnax: At that point we might as well just have different client functions for each hook that we want, rather than directly specifying the hooktype to the client function.

internal/rest/resources/hooks.go
masnax: Actually, thinking more about this I think you're right, it's a bit unnecessary to keep the name around. As part of the built-in forwarding mechanism I mentioned above, we apply a query parameter `?target={clusterMember}` so if we need to make a distinction between global and per-node actions, we can leverage the query parameter.

masnax:
OK, changed this a bit now.

The endpoint is top-level `/cluster/internal/hooks/{hookType}` and has forwarding enabled to particular cluster members if `?target={clusterMember}` is set to true, with the `ProxyTarget=true` flag set in its `EndpointHandler`. 

Each hook has its own client function which take as argument one of two new API types: `types.HookRemoveMemberOptions` which sets the `force` flag, and `types.HookNewMemberOptions` which has the name of the new member. I'm thinking of extending the `OnNewMember`hook in a different PR to actually take in a `types.ClusterMember` so that the implementation of the hook can be tailored to the information we have on the cluster member without having to do another database fetch to grab it from within the hook.

internal/rest/types/hooks.go
```diff
@@ -0,0 +1,46 @@
...
+
+// HookRemoveMemberOptions holds configuration pertaining to the PreRemove and PostRemove hooks.
+type HookRemoveMemberOptions struct {
+	// Force represents whether to run the hook with the `force` option.
+	Force bool `json:"force" yaml:"force"`
+}
+
+type HookNewMemberOptions struct {
+	// NewMemberName is the name of the new cluster member that joined the cluster, triggering this hook.
+	NewMemberName string `json:"new_member_name" yaml:"new_member_name"`
```
roosterfish, line 45: Just a nit. Do we need the `NewMember` in both the name of the struct and the field? Maybe for the field `Name` would be enough :)

roosterfish (changes requested):
Thanks for the changes! Wasn't aware of the `?target=` query param but using this sounds even better. Just one remark.

internal/rest/types/hooks.go
masnax: Oops, fixed now.

roosterfish (approved):

Issue #96, opened 2024-03-26
Guard against MitM attack: remove node name from the token


louiseschmidtgen:
I propose enhancing the security of microcluster join tokens by removing the node name from the token payload. Currently, including the node name in the token compromises the validation process, particularly in guarding against potential attacks.

Utilizing Angelos's previous work as a starting point ([neoaggelos:microcluster:dev/token-names](https://github.com/canonical/microcluster/compare/main...neoaggelos:microcluster:dev/token-names?expand=1)), I suggest revising the client auth.

louiseschmidtgen:
@neoaggelos @tomponline @bschimke95

tomponline:
@louiseschmidtgen please could you describe a scenario where the node name in the token payload could be abused? What impact would that have? Thanks

@masnax i'd be interested to hear your thoughts on this? Is cluster member name only taken from the token?

bschimke95:
Adding some more context here: 
Currently, microcluster does not validate that the node is actual the one that we issued a token for. If an malicious actor obtains the join token it can join the cluster without further validation. For example, this is possible in the k8s snap:
```
# Token is generated on the cluster node
clusternode$ sudo k8s get-join-token mygoodnode
<token>

# Bad actor gets <token>, decodes it, gets node name and can now join some random other node:
mybadnode$ sudo k8s join-cluster <token> --name mygoodnode 
```
This works. The node name that is part of the token payload is not used at all in microcluster right now.

So, if a bad actor gets the token he can just read the payload and get the node name to join the cluster. 
To avoid that, we should remove the node name from the token payload and use the node name that the joining node provides to validate that the node uses the token that was issued for it.

tomponline:
I see thanks. So the main complaint is that it is possible to derive which cluster member the join token was issued for from the join token itself and thus makes it easier to use said join token if a malicious actor gets hold of it.

The join tokens themselves are meant to be secret - they represent a way to join a cluster so should not be made public.
However if the cluster member name inside the token isn't used then it would see sensible to remove it and make things more difficult for an attacker to use a join token if they get hold of one.

beliaev-maksim:
@bschimke95 will node name be sufficient ? This could be easily spoofed 

anyway implement real fingerprint validation where user can compare fingerprints on both main and worker nodes

beliaev-maksim:
@tomponline 
[...]

PR #115, opened 2024-04-22
Remove name from join token

MggMuggins:
Resolves the titular concern in #96 

@masnax I had mentioned that this caused cluster join to fail and... I can't reproduce it. Cluster forms up nicely. Likely had a messy worktree at the end of the day or some such. This passes what tests there are in #110.

### DIFF
--- a/cluster/token_records.go
+++ b/cluster/token_records.go
@@ -43,7 +43,6 @@ type InternalTokenRecordFilter struct {
 // ToAPI converts the InternalTokenRecord to a full token and returns an API compatible struct.
 func (t *InternalTokenRecord) ToAPI(clusterCert *x509.Certificate, joinAddresses []types.AddrPort) (*internalTypes.TokenRecord, error) {
 	token := internalTypes.Token{
-		Name:          t.Name,
 		Secret:        t.Secret,
 		Fingerprint:   shared.CertFingerprint(clusterCert),
 		JoinAddresses: joinAddresses,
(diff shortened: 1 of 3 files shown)


roosterfish (approved):
Looks good to me. 

I am wondering if we should hold this back until we have agreed on how to proceed with the join process? (https://discourse.ubuntu.com/t/secure-cluster-join-in-microcloud/44261/18). 

Overall I see it unrelated to the join process as it really is just a label right now. @masnax any thoughts on this?

tomponline:
@MggMuggins please can you see if LXD should also have the node name removed from its join token if its also not used? Thanks

tomponline:
> Looks good to me.
> 
> I am wondering if we should hold this back until we have agreed on how to proceed with the join process? (https://discourse.ubuntu.com/t/secure-cluster-join-in-microcloud/44261/18).
> 
> Overall I see it unrelated to the join process as it really is just a label right now. @masnax any thoughts on this?

Although I don't see an issue with merging this, I do also think we shouldn't close the related issue as the mitm scenario still exists and this change doesn't fix it.

MggMuggins:
@tomponline It looks like the cluster name is being sent from the new cluster joiner [here](https://github.com/canonical/lxd/blob/main/lxd/main_init.go#L168) and is required by the local LXD's API [here](https://github.com/canonical/lxd/blob/main/lxd/api_cluster.go#L338). It looks like the name is used for a couple of additional requests to `/internal` endpoints although my understanding of what goes on during the cluster join process gets fuzzy after the request is submitted to the local LXD.

TL;DR it doesn't look like it will be trivial to remove in LXD like it was here.

PR #117, opened 2024-04-26
Fix bug introduced with truststore API

masnax:
Couple things:

* The truststore API was being used to update the cluster member's local node records each time their databases start, but this was causing a quorum loss loop because if the daemon restarts, no node will be available to handle the request and the database will close prematurely.

* Additionally slightly related is that when we close the dqlite connection, we first try to shut down the database, but if other nodes have already closed their dqlite connections then we won't be able to close the database and the defunct connection will be left open, causing problems if the daemon restarts too quickly. I've changed this to skip the error check for `db.Close` since `dqlite.Close` will close it for us. This is the same behaviour as in LXD today.

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -463,33 +463,35 @@ func (d *Daemon) StartAPI(bootstrap bool, initConfig map[string]string, newConfi
 		}
 	}
 
-	var lastErr error
-	var clusterConfirmation bool
-	err = cluster.Query(d.shutdownCtx, true, func(ctx context.Context, c *client.Client) error {
-		// No need to send a request to ourselves.
-		if d.address.URL.Host == c.URL().URL.Host {
-			return nil
-		}
-
-		// At this point the joiner is only trusted on the node that was leader at the time,
-		// so find it and have it instruct all dqlite members to trust this system now that it is functional.
-		if !clusterConfirmation {
-			err := internalClient.AddTrustStoreEntry(ctx, &c.Client, localMemberInfo)
-			if err != nil {
-				lastErr = err
-			} else {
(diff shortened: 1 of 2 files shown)


internal/daemon/daemon.go
```diff
@@ -463,33 +463,35 @@ func (d *Daemon) StartAPI(bootstrap bool, initConfig map[string]string, newConfi
...
-		// At this point the joiner is only trusted on the node that was leader at the time,
-		// so find it and have it instruct all dqlite members to trust this system now that it is functional.
 		if !clusterConfirmation {
-			err := internalClient.AddTrustStoreEntry(ctx, &c.Client, localMemberInfo)
-			if err != nil {
-				lastErr = err
-			} else {
-				clusterConfirmation = true
-			}
+			return fmt.Errorf("Failed to confirm new member %q on any existing system (%d): %w", localMemberInfo.Name, len(cluster)-1, lastErr)
```
roosterfish, line 493: The other day in another review I saw a suggestion for using the `Remotes.Count()` func instead to get the overall number of cluster members. 
Is there a preference when one or the other should be used?

roosterfish (approved):
Looks good to me.

internal/daemon/daemon.go
masnax: It doesn't matter too much as long as you're not calling `Remotes()` again. 

In this case we have a client for a specific number of remotes in the `cluster` var, so the error should reflect that we tried to reach out to that many nodes. (Less one for ourselves).

Calling `Remotes()` again may yield a different number because of `fsnotify` which monitors the truststore for changes. So in a lot of places we instantiate `Remotes()` only once for the duration of the task.

PR #120, opened 2024-05-01
internal/rest/resources: Reorganize authentication handlers

masnax:
Moves the authentication handlers out into their own function that is exported and can be used to facilitate finer control of authentication per-endpoint.

Additionally allows passing in a map of x509.Certificate to authenticate against a set other than the default truststore, if desired. If used externally in a project specific AccessHandler, the request context will still report that the request is untrusted by the truststore, unless the certificate used to authenticate the request also exists in the truststore. In either case, the request will still go through.

@roosterfish This is what I was talking about today (or yesterday for you :))


rest/access/handlers.go
```diff
@@ -0,0 +1,53 @@
...
+// proceed. However in order to access any API route you must be authenticated, unless the handler's AllowUntrusted
+// property is set to true or you are an admin.
+func AllowAuthenticated(state *state.State, r *http.Request) response.Response {
+	return response.EmptySyncResponse
+}
+
+// Authenticate ensures the request certificates are trusted against the given set of trusted certificates.
+// - Requests over the unix socket are always allowed.
+// - HTTP requests require the TLS Peer certificate to match an entry in the supplied map of certificates.
+func Authenticate(state *state.State, r *http.Request, trustedCerts map[string]x509.Certificate) (bool, error) {
```
roosterfish, line 26: Thanks for putting this PR up! 

So if you start up a MicroCluster would it be the idea to use this `Authenticate` function as an access handler when setting up the slice of custom `[]rest.Endpoint`s? This means if the endpoint is called `handleAPIRequest` will then run this custom handler under the hood.

masnax: So if you look at MicroCloud's `rest.Endpoint`s, they are wrapped in a function that takes the `service.Handler` as an argument: https://github.com/canonical/microcloud/blob/facd3ae55bd74ba261b93a57bea2177e512f4e8d/microcloud/api/services.go#L47-L56

So if the ephemeral truststore is a field or method of `service.Handler`, we could have something like this:

```go
func serviceAccessHandler(sh *service.Handler) func(*state.State, *http.Request) response.Response {
    return func(s *state.State, r *http.Request) response.Response {
		if r.RemoteAddr == "@" {
			logger.Debug("Allowing unauthenticated request through unix socket")

			response.EmptySyncResponse
		}

                trusted, err := access.Authenticate(s, r, sh.EphemeralTruststore)
                if err != nil {
                        response.SmartError(fmt.Errorf("Failed to perform authentication check: %w", err)
                }

                if trusted {
                        return response.EmptySyncResponse
                }

		return response.SmartError(fmt.Errorf("Request is not trusted"))
	} 
}

var ServicesCmd = func(sh *service.Handler) rest.Endpoint {
	return rest.Endpoint{
		AllowedBeforeInit: true,
		Name:              "services",
		Path:              "services",

		Put: rest.EndpointAction{Handler: servicesPut, ProxyTarget: true, AccessHandler: serviceAccessHandler(sh)},
	}
}
```

And that will make this endpoint only authorize requests with a corresponding certificate in the ephemeral truststore (or on the local unix socket).

roosterfish: Ah ok, I was just about to ask why it has to be a field of `service.Handler` until I saw that this is instead coming from MicroCloud and isn't a datatype which is MicroCluster specific. 
But yeah it makes sense.

Do you want to get this merged? I was just thinking if it would be worth splitting the changes up into smaller commits.

masnax: Good point, I've split up the commits now.

roosterfish (approved):
LGTM

PR #121, opened 2024-05-01
Fix api extension regex

masnax:
The previous regex was failing on `custom_extension_a_0` which is used in the example package.

### DIFF
--- a/internal/extensions/extensions.go
+++ b/internal/extensions/extensions.go
@@ -9,8 +9,8 @@ import (
 	"github.com/canonical/lxd/shared"
 )
 
-var internalExtensionRegex = regexp.MustCompile(`^internal:([a-z0-9]|[a-z0-9]_[a-z0-9])+$`)
-var externalExtensionRegex = regexp.MustCompile(`^([a-z0-9]|[a-z0-9]_[a-z0-9])+$`)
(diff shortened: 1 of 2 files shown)


internal/extensions/extensions.go
```diff
@@ -9,8 +9,8 @@ import (
 	"github.com/canonical/lxd/shared"
 )
 
-var internalExtensionRegex = regexp.MustCompile(`^internal:([a-z0-9]|[a-z0-9]_[a-z0-9])+$`)
-var externalExtensionRegex = regexp.MustCompile(`^([a-z0-9]|[a-z0-9]_[a-z0-9])+$`)
+var internalExtensionRegex = regexp.MustCompile(`^internal:[a-zA-Z0-9]+(_[a-zA-Z0-9]+)*$`)
```
roosterfish, line 12: Good catch! 

But do we want to allow uppercase characters `A-Z` too as part of this change?

roosterfish (changes requested):

internal/extensions/extensions.go
masnax: Oops :)

roosterfish (approved):

PR #122, opened 2024-05-03
Disable foreign keys for internal updates

masnax:
We can't be sure which tables specified by external users of microcluster reference internal tables. So just broadly turn off foreign keys while applying internal schema updates, and turn it back on afterward.

### DIFF
--- a/internal/db/update/schema.go
+++ b/internal/db/update/schema.go
@@ -66,18 +66,26 @@ func (s *SchemaUpdate) Version() (internalVersion uint64, externalVersion uint64
 func (s *SchemaUpdate) Ensure(db *sql.DB) (int, error) {
 	var current int
 	aborted := false
+	versions := []int{0, 0}
+	var exists bool
 	err := query.Transaction(context.TODO(), db, func(ctx context.Context, tx *sql.Tx) error {
 		err := execFromFile(ctx, tx, s.path, s.hook)
 		if err != nil {
 			return fmt.Errorf("Failed to execute queries from %s: %w", s.path, err)
 		}
 
-		exists, err := doesSchemaTableExist(tx)
+		exists, err = doesSchemaTableExist(tx)
 		if err != nil {
 			return fmt.Errorf("Failed to check if schema table is there: %w", err)
 		}
 
-		var versions []int
+		return nil
+	})
+	if err != nil {
+		return -1, err
+	}
+
+	err = query.Transaction(context.TODO(), db, func(ctx context.Context, tx *sql.Tx) error {
 		if exists {
 			// updateFromV1 changes the schema table and needs to be run before we calculate the schema version.
 			err := updateFromV1(ctx, tx)
@@ -97,26 +105,37 @@ func (s *SchemaUpdate) Ensure(db *sql.DB) (int, error) {
 
 		if s.check != nil {
 			err := s.check(ctx, current, tx)
+			if err != nil && err != schema.ErrGracefulAbort {
+				return err
+			}
+
 			if err == schema.ErrGracefulAbort {
 				// Abort the update gracefully, committing what
 				// we've done so far.
 				aborted = true
-				return nil
-			}
-			if err != nil {
-				return err
 			}
 		}
 
+		return nil
+	})
+	if err != nil {
+		return -1, err
+	}
+
+	if aborted {
+		return current, schema.ErrGracefulAbort
+	}
+
+	err = query.Transaction(context.TODO(), db, func(ctx context.Context, tx *sql.Tx) error {
 		// When creating the schema from scratch, use the fresh dump if
 		// available. Otherwise just apply all relevant updates.
-		if versions == nil && s.fresh != "" {
+		if versions[updateExternal] == 0 && versions[updateInternal] == 0 && s.fresh != "" {
 			_, err = tx.ExecContext(ctx, s.fresh)
 			if err != nil {
 				return fmt.Errorf("Cannot apply fresh schema: %w", err)
 			}
 		} else {
-			err = ensureUpdatesAreApplied(ctx, tx, versions, s.updates, s.hook)
+			err = ensureUpdatesAreApplied(ctx, tx, updateInternal, versions[updateInternal], s.updates[updateInternal], s.hook)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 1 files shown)


masnax:
This one should get merged ASAP as the bug may result in data loss if anyone is using the edge of microcluster.

gabrielmougard (dismissed):
Looks good! thanks.

internal/db/update/schema.go
```diff
@@ -153,6 +153,15 @@ func ensureUpdatesAreApplied(ctx context.Context, tx *sql.Tx, versions []int, sc
 			return nil
 		}
 
+		// Ensure we turn off foreign keys so any external table we don't know about
```
roosterfish, line 156: If one of the external tables columns has a foreign key on any of the internal tables, how would this affect the relation? 
I am wondering if this could lead to an inconsistent state in the external tables as a user of the external table would expect the foreign key constraint to be working.

Wouldn't it be better to allow this behavior but make it clear that you should never reference those tables as they are internal?

masnax: It shouldn't matter since the foreign keys are turned back on after the internal updates are applied. Since all internal updates are applied before external ones, I don't see any way that this would interfere with external tables.

That being said, it appears my approach was a bit naive because you can't set `PRAGMA foreign_keys=OFF` from within a transaction, which means this PR is going to have to be a bit more involved.

roosterfish: Take the following as an example:

```sql
CREATE TABLE internal (
  column1 INTEGER PRIMARY KEY
)

CREATE TABLE external (
  column1 INTEGER
  FOREIGN KEY (column1) REFERENCES "internal" (column1)
)
```

Now let's assume there is a row in `external` which points to another row in `internal`. If we do `PRAGMA foreign_keys=OFF` (or something equal that works) and get rid of the row in `internal` and later on turn foreign keys back on, wouldn't the foreign key constraint `external(column1) -> internal(column1)` be broken?

masnax: Yes it would, however there are no pre-existing rows in any table, they are all determined based on user input so there would never be a valid schema update that purposely deletes a row from a table, that would be extremely dangerous. 


If we do for some reason have to delete a row, it makes way more sense to me to actually break the relation first because we shouldn't be arbitrarily deleting data for an external schema. At least in that case the external schema has some semblance of pathway for recovery, instead of having rows in its tables deleted, they would simply have dereferenced rows. 

In the present case, we are migrating rows to a different table which we then rename back to the old table name. This is a common occurrence for schema updates, but if we don't turn foreign keys off first each time, then when we drop the original table, any external tables with a foreign key to that one will be wiped. And since we we can't turn primary keys off in a transaction that means we can't run all updates in one big transaction like we do now either.

masnax: Since external schema updates are all run after the internal ones, these updates can be used to reconcile the external schema with the new internal one if the rows continue to exist in the external table. I can't think of any situation where we would be okay with deleting data external to microcluster from an internal schema update.

masnax:
Because `PRAGMA foreign_keys=OFF` has no effect if invoked a transaction, we have to split up the transaction into multiple parts. I've now updated this PR to split up the transaction where we run schema updates into 4 parts.

* The first transaction checks if the schemas table exists.
  * If the schema table does exist, then we check if we need to run the special v1 update that modifies the schema table so that the queries we make to it as part of the schema update process do not fail. 
  * If we need to run the v1 update, then we turn off foreign keys before the next transaction so that external references to the `internal_cluster_members` table are not wiped.

* The second transaction applies the special v1 update, and fetches the maximum internal and external schema versions from the schemas table. 
  * If we detect that some nodes are behind with the `schema.ErrGracefulAbort`, we return early.
  * If we need to return early, we re-enable foreign keys.
  * If we are not returning early, and our max internal schema version implies that we have updates to apply, then we turn off foreign keys outside the transaction if we didn't turn them off after the first transaction.

* The third transaction applies internal schema updates.
  * After all the internal schema updates are committed, we re-enable foreign keys before proceeding to external schema updates.

* The final transaction applies external schema updates.

tomponline:
Please can we chat about this next week at the sprint. This is starting to sound rather complicated and i dont understand why it has gotten to be this way. Thanks

tomponline:
Also LXD has been known to have to do things like this occasionally in patches, but to my knowledge its not done routinely like this.

masnax:
> Also LXD has been known to have to do things like this occasionally in patches, but to my knowledge its not done routinely like this.

LXD is quite a different situation than microcluster because LXD is in full control of its schema. If we need to make a change, we know and can manage any relationships between the tables. In microcluster this isn't the case. We don't know what the upstream project's supplied schema looks like so we must be very careful when we make changes to the internal schema, and we must consider that alterations to the internal schema can affect relations to external tables that we dont know about. 

It was intended for microcluster to supply its internal tables as a source of truth for the cluster state to upstream projects. MicroCeph for example needs to associate OSDs to the cluster member that supplied the disks, and does so through a reference to the `internal_cluster_members` table. 

So we are left with non-trivial options. I can think of these:

* We can leave things as they are and just fix the existing schema updates to use `ADD COLUMN` instead of replacing the whole table and remember to never drop any tables, rows, or columns in any schema update in the future. 

* We disable foreign keys on a case by case basis. I think we will end up doing this more often than not because any alteration to an existing table will be at risk. We only wouldn't need to disable foreign keys if we are just adding a table, and there's not really many more internal tables that microcluster needs. Since you can't disable foreign keys in a transaction, we would either have to explicitly run every schema update in its own transaction, or we would have to invoke something like this in each schema update call to `tx.Exec`:
```SQL
COMMIT TRANSACTION;
PRAGMA foreign_keys=OFF;
BEGIN TRANSACTION;
// body of schema update
COMMIT TRANSACTION;
PRAGMA foreign_keys=ON;
BEGIN TRANSACTION;
```

* We blanket disable foreign keys for the whole internal schema update process. 

* We can reasses the approach of allowing external tables to reference internal ones. This would require a hefty change in all upstream projects as they update their schemas. We would then also need to implement some synching mechanisms so that external tables are updated appropriately when a change is made on the internal side (eg. a cluster member is removed or modified).

tomponline:
Thanks Max. I think this warrants an architectural discussion around what micro cluster provides to its consumers. Thanks for laying out the landscape of options.

internal/db/update/schema.go
roosterfish: It's clear to me that the internal schema update should never do any modifications to any of the external schemas/data. 

However it feels like a huge burden on our end as we cannot modify the schema (e.g. rename rows) without fearing that this might break an external setup with active foreign keys on those rows. Maybe the example from before wasn't ideal, but what if we decide to change `column1` to `column2` in the updated `internal` table?

As an appendix to your list from below, we could provide views to export a defined set of data. This won't allow foreign key constraints (as rows in views cannot be referenced by foreign key) but it could be used to reference the internal data and we are always able to modify the underlying structure using internal schema updates.

tomponline: Couple of points/questions here:

1. Once we have the LTS branch of microcluster then we cannot make *any* schema changes inside that branch after the first release (this is our commitment to allow downgrades).
2. Is there anything stopping microcluster have its own internal core DB, and the microcluster consumers (microcloud, microovn, microceph etc) having their own app specific database, and accessing microcluster core data via an API. This would prevent cross-db references?

tomponline:
> * If we detect that some nodes are behind with the `schema.ErrGracefulAbort`, we return early.

Is this effectively the same as `patchDBNodesAutoInc` in LXD? Albeit it sounds more complex in the microcluster case somehow.

internal/db/update/schema.go
masnax: > On 2., I don't see how that solves the problem posed in step 1. It's just punting it to another area. If the upstream project relies on the API to balance its own database then we would be just as locked out of making API changes as we are currently for schema changes, because there's no reverting the effect that the API has on the upstream database once the nodes start up for the first time and rebalance.
> 
> On 3. Yes, the schema was designed to be as minimal as possible such that the only situations where we would need a schema update are when we are adding some feature (in this case: API extensions). We ended up with a second schema update from #94 because we found a bug with the schema update process that locked only the internal side out from adding such features. We offer a table for join tokens that just store each token's name and secret key, and a table for cluster members. External projects are expected to reference these tables and our schema updates should only be to extend the data we offer here.

I think this philosophy fits with point 1 because we won't be continually adding features to an LTS release.

masnax:
> > * If we detect that some nodes are behind with the `schema.ErrGracefulAbort`, we return early.
> 
> Is this effectively the same as `patchDBNodesAutoInc` in LXD? Albeit it sounds more complex in the microcluster case somehow.

I'm a bit confused by the line you quoted because I don't see the connection between that function and the error that makes us abort updates. 

The purpose of the function seems similar enough to me if I understand it correctly, the key difference I see is that needing foreign keys to be disabled is a special case for LXD's schema extensions but they are really the only type of schema updates I forsee us needing in microcluster. 

As for the complexity of needing to split the transaction up into 4 parts, which I'm assuming is what you meant with the line you quoted, the reason for that is just to accommodate the #94 schema update fix and the two sets of schema updates that are unique to microcluster. 

In LXD's case, the process looks like this:
* We check for existence of the schema table. 
* Then do some setup like creating that table if we need to, and checking if we need to apply any updates.
* Finally we apply the updates or return early.
* We also have patches that run at different stages of the daemon start proces. 

Microcluster is similar:
* Check for existence of the schema table.
*  Then we run the update from #94 to fix the broken default schema table. We could say this is analogous to a one-and-done patch without a whole mechanism behind it. 
* Then comes the setup like in LXD
* Apply internal updates if needed
* Apply external updates if needed. 

We need to disable foreign keys in two cases: Bullet points 2 and 4, hence needing to break the transaction up into parts.

This approach makes us only disable foreign keys if we are absolutely sure we are going to modify the internal schema. If we want, we can be more relaxed and just disable foreign keys blindly up until we reach the point where we want to apply external updates, at which point we can enable foreign keys first. That would get rid of all the `if weReallyNeedTo` type statements, and we would onlupy need 2 transactions (1 for bullet points 1-4, and 1 for bullet point 5).

internal/db/update/schema.go
tomponline: > On 2., I don't see how that solves the problem posed in step 1. It's just punting it to another area. If the upstream project relies on the API to balance its own database then we would be just as locked out of making API changes as we are currently for schema changes, because there's no reverting the effect that the API has on the upstream database once the nodes start up for the first time and rebalance.

The way I see that it solves/clarifies the problem as it defines clearly what is microcluster's API commitments to its consumers, rather than having an a potential unknown dependency on its own internal core tables.

If you are saying that the microcluster core internal tables *are* also an API, then it becomes quite simple really, we just cannot change the schema in a breaking way.

tomponline: TBH I am very confused by this now, will wait until next week when we can discuss in person.

masnax: > If you are saying that the microcluster core internal tables _are_ also an API, then it becomes quite simple really, we just cannot change the schema in a breaking way.

Yes, this is a good way to sum up my thoughts on this. The internal database is simply another API that microcluster provides, and since it's a database we don't have to implement all the mechanics that interconnect the upstream database to the data we supply, we can just rely on sqlite handling it for us.

masnax:
@roosterfish @tomponline

I've now tested both references with `FOREIGN KEY`, as well as `VIEW`s will still exist with a dropped or renamed column, and can be updated to reference a new column. 

For this to work, we need to set `PRAGMA foreign_keys=OFF` and `PRAGMA legacy_alter_table=ON` before each transaction in which we run any changes to the internal schema.

roosterfish:
> For this to work, we need to set `PRAGMA foreign_keys=OFF` and `PRAGMA legacy_alter_table=ON` before each transaction in which we run any changes to the internal schema.

Cool, so this means we can proceed with the plan which allows a downstream importer of MicroCluster to supply an external DB patch which will then align the external tables to the internal ones which have been intentionally "broken" by the internal schema update?

internal/db/update/schema.go
```diff
@@ -147,6 +191,14 @@ func (s *SchemaUpdate) Ensure(db *sql.DB) (int, error) {
 		return -1, err
```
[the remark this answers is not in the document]
masnax: If there is an error applying the schema update, the database will shut down. The next time it starts, it will go through applying the schema updates again, and only enable foreign keys before actually opening the database. 

For that reason, I don't think we need to concern ourselves with errors here because an error means you don't even have a database.

internal/db/update/schema.go
```diff
@@ -79,20 +80,45 @@ func (s *SchemaUpdate) Ensure(db *sql.DB) (int, error) {
...
+		if err != nil {
+			return -1, err
+		}
+	}
+
 	err = query.Transaction(context.TODO(), db, func(ctx context.Context, tx *sql.Tx) error {
-		if exists {
+		if exists && updateSchemaTable {
 			// updateFromV1 changes the schema table and needs to be run before we calculate the schema version.
 			err := updateFromV1(ctx, tx)
```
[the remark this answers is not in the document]
masnax: As mentioned in my other comment, if there is an error in these transactions, the database should not open, so we don't need to ensure that foreign keys are turned on.

internal/db/update/schema.go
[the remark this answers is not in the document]
roosterfish: Fair point if nothing in MicroCluster is usable after this point onwards.

internal/db/update/schema.go
```diff
@@ -127,61 +146,63 @@ func (s *SchemaUpdate) Ensure(db *sql.DB) (int, error) {
...
-			return fmt.Errorf("Schema version '%d' is more recent than expected '%d'", version, len(updates))
-		}
-
-		// If there are no updates, there's nothing to do.
-		if len(updates) == 0 {
-			return nil
-		}
+func ensureUpdatesAreApplied(ctx context.Context, tx *sql.Tx, updateType updateType, version int, updates []schema.Update, hook schema.Hook) error {
+	if version > len(updates) {
+		return fmt.Errorf("Schema version '%d' is more recent than expected '%d'", version, len(updates))
```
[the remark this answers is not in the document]
masnax: Fixed

roosterfish (approved):
LGTM!

PR #127, opened 2024-05-09
Misc fixes

masnax:
Cleans up a noisy log and fixes the example package, which was failing to run the `OnStart` hook because of an incorrectly ordered set of external API extensions.

### DIFF
--- a/example/cmd/microd/main.go
+++ b/example/cmd/microd/main.go
@@ -98,7 +98,7 @@ func (c *cmdDaemon) Run(cmd *cobra.Command, args []string) error {
 
 			// You can also create a new registry of extensions from a list of extensions.
 			// This is useful to communicate a system's extensions to other systems, for comparison purposes for example.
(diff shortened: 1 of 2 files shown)


example/cmd/microd/main.go
```diff
@@ -98,7 +98,7 @@ func (c *cmdDaemon) Run(cmd *cobra.Command, args []string) error {
 
 			// You can also create a new registry of extensions from a list of extensions.
 			// This is useful to communicate a system's extensions to other systems, for comparison purposes for example.
-			newExt, err := extensions.NewExtensionRegistryFromList([]string{"internal:runtime_extension_v1", "new_extension_at_runtime_1", "new_extension_at_runtime_2", "custom_extension_a_0", "custom_extension_a_1"})
+			newExt, err := extensions.NewExtensionRegistryFromList([]string{"internal:runtime_extension_v1", "custom_extension_a_0", "custom_extension_a_1", "new_extension_at_runtime_1", "new_extension_at_runtime_2"})
```
roosterfish, line 101: Was the `err = s.Extensions.IsSameVersion(newExt)` check failing as the order of the extensions didn't match?

masnax: Yeah, we decided to not sort the list of extensions to preserve the order they were introduced in, but this part was overlooked.

PR #130, opened 2024-05-22
Allow running extra network listeners

markylaing:
Allows microcluster downstreams to specify extra servers that they would like microcluster to set up. Handlers in the extension servers will have access to microcluster state.

To do:
- [x] Discuss implications for authentication.
- [x] Separate commits

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -306,7 +306,7 @@ func (d *Daemon) initStore() error {
 	return nil
 }
 
-func (d *Daemon) initServer(resources ...*resources.Resources) *http.Server {
+func (d *Daemon) initServer(resources ...rest.Resources) *http.Server {
 	/* Setup the web server */
 	mux := mux.NewRouter()
 	mux.StrictSlash(false)
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -68,6 +68,8 @@ type Daemon struct {
 
 	// stop is a sync.Once which wraps the daemon's stop sequence. Each call will block until the first one completes.
 	stop func() error
+
+	extensionServers []rest.Server
 }
 
 // NewDaemon initializes the Daemon context and channels.
@@ -95,8 +97,9 @@ func NewDaemon(project string) *Daemon {
 // Run initializes the Daemon with the given configuration, starts the database, and blocks until the daemon is cancelled.
 // - `extensionsAPI` is a list of endpoints to be served over `/1.0`.
 // - `extensionsSchema` is a list of schema updates in the order that they should be applied.
+// - `extensionServers` is a list of rest.Server that will be initialized and managed by microcluster.
 // - `hooks` are a set of functions that trigger at certain points during cluster communication.
-func (d *Daemon) Run(ctx context.Context, listenPort string, stateDir string, socketGroup string, extensionsAPI []rest.Endpoint, extensionsSchema []schema.Update, apiExtensions []string, hooks *config.Hooks) error {
+func (d *Daemon) Run(ctx context.Context, listenPort string, stateDir string, socketGroup string, extensionsAPI []rest.Endpoint, extensionsSchema []schema.Update, apiExtensions []string, extensionServers []rest.Server, hooks *config.Hooks) error {
 	d.shutdownCtx, d.shutdownCancel = context.WithCancel(ctx)
 	if stateDir == "" {
 		stateDir = os.Getenv(sys.StateDir)
@@ -116,6 +119,8 @@ func (d *Daemon) Run(ctx context.Context, listenPort string, stateDir string, so
 	if err != nil {
 		return fmt.Errorf("Failed to initialize directory structure: %w", err)
 	}
+
+	d.extensionServers = extensionServers
 
 	err = d.init(listenPort, extensionsAPI, extensionsSchema, apiExtensions, hooks)
 	if err != nil {
@@ -433,6 +438,12 @@ func (d *Daemon) StartAPI(bootstrap bool, initConfig map[string]string, newConfi
 			return err
 		}
 
+		// Add extension servers before post-bootstrap hook.
+		err = d.addExtensionServers()
+		if err != nil {
+			return err
+		}
+
 		err = d.hooks.PostBootstrap(d.State(), initConfig)
 		if err != nil {
 			return fmt.Errorf("Failed to run post-bootstrap actions: %w", err)
@@ -550,8 +561,32 @@ func (d *Daemon) StartAPI(bootstrap bool, initConfig map[string]string, newConfi
 		return err
 	}
 
+	// Add extension servers before post-join hook.
+	err = d.addExtensionServers()
+	if err != nil {
+		return err
+	}
+
 	if len(joinAddresses) > 0 {
 		return d.hooks.PostJoin(d.State(), initConfig)
+	}
+
+	return nil
+}
+
+// addExtensionServers initialises a new *endpoints.Network for each extension server and adds it to the Daemon endpoints.
+func (d *Daemon) addExtensionServers() error {
+	var networks []endpoints.Endpoint
+	for _, extensionServer := range d.extensionServers {
+		server := d.initServer(extensionServer.Resources...)
+		url := api.NewURL().Scheme(extensionServer.Protocol).Host(extensionServer.Address.String())
+		network := endpoints.NewNetwork(d.shutdownCtx, endpoints.EndpointNetwork, server, *url, extensionServer.Certificate)
+		networks = append(networks, network)
+	}
+
+	err := d.endpoints.Add(networks...)
+	if err != nil {
+		return err
 	}
 
 	return nil
--- a/internal/rest/rest.go
+++ b/internal/rest/rest.go
@@ -3,6 +3,7 @@ package rest
 import (
 	"context"
 	"database/sql"
+	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
@@ -201,8 +202,8 @@ func HandleEndpoint(state *state.State, mux *mux.Router, version string, e rest.
 			handleRequest = handleDatabaseRequest
 		}
 
-		trusted, err := access.Authenticate(state, r, state.Remotes().CertificatesNative())
-		if err != nil {
+		trusted, err := access.Authenticate(state, r, state.Address().URL.Host, state.Remotes().CertificatesNative())
+		if err != nil && !errors.As(err, &access.ErrInvalidHost{}) {
 			resp = response.Forbidden(fmt.Errorf("Failed to authenticate request: %w", err))
 		} else {
 			r = internalAccess.SetRequestAuthentication(r, trusted)
--- a/internal/rest/rest.go
+++ b/internal/rest/rest.go
@@ -24,37 +24,23 @@ import (
 )
 
 func handleAPIRequest(action rest.EndpointAction, state *state.State, w http.ResponseWriter, r *http.Request) response.Response {
-	trusted := r.Context().Value(request.CtxAccess)
-	if trusted == nil {
-		return response.Forbidden(nil)
-	}
-
-	trustedReq, ok := trusted.(internalAccess.TrustedRequest)
-	if !ok {
-		return response.Forbidden(nil)
-	}
-
-	// If the request is not trusted, only allow it if AllowUntrusted is set.
-	if !trustedReq.Trusted && !action.AllowUntrusted {
-		return response.Forbidden(nil)
-	}
-
 	if action.Handler == nil {
 		return response.NotImplemented(nil)
 	}
 
-	if action.AccessHandler != nil {
-		// Defer access control to custom handler.
-		accessResp := action.AccessHandler(state, r)
-		if accessResp != response.EmptySyncResponse {
-			return accessResp
+	// If allow untrusted is not set, the request must be authenticated via core authentication (e.g. certificate in truststore).
+	if !action.AllowUntrusted {
+		resp := access.AllowAuthenticated(state, r)
+		if resp != response.EmptySyncResponse {
+			return resp
 		}
-	} else if !action.AllowUntrusted {
-		// Set the default access handler if one isn't specified.
-		action.AccessHandler = access.AllowAuthenticated
-		accessResp := action.AccessHandler(state, r)
-		if accessResp != response.EmptySyncResponse {
-			return accessResp
+	}
+
+	// Run the custom access handler if set.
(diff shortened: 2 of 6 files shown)


rest/access/handlers.go
```diff
@@ -46,7 +45,7 @@ func Authenticate(state *state.State, r *http.Request, trustedCerts map[string]x
 			}
 		}
 	default:
-		return false, fmt.Errorf("Invalid request address %q", r.Host)
```
markylaing, line 49: @masnax @roosterfish @tomponline I don't necessarily think that this change is a good idea. I'll keep this PR as draft until we've discussed in more detail.

`Authenticate` is called on every API request for all listeners. As microcluster is being used by more and more applications for various purposes, it's becoming a little too prescriptive in the way that requests are handled. 

In general, I think microcluster needs to be changed to remove the `AccessHandler` and `AllowUntrusted` from `EndpointAction` and instead accept a slice of handlers as middleware. Each handler should return a `response.Response`, if the response.Response is nil, we move onto the next handler. Downstreams can then implement their own authentication (and other middleware e.g. for logging/tracing). We can expose some handlers to downstreams that they may want to use - such as a wrapper for `Authenticate`.

masnax: As for dropping `AllowUntrusted`, I'll give a bit of background as to how we treat trust in microcluster:

*  `Authenticate` is run by default using the default network address and the default truststore, and the trusted status of the request is stored in the request context's `CtxAccess`. This indicates whether the request is trusted by the core API and configuration. I think we should always keep a record of this for every endpoint in case we need to perform specific internal handling in the future. It will become very clear.

* Currently, if a downstream application wants to have their own access handler, they would have to set `AllowUntrusted=true` and create a custom `AccessHandler` that calls `Authenticate` with their specific parameters. In this case, the endpoint can be deemed trusted but the `CtxAccess` field in the request context will still be set to untrusted because the endpoint is not trusted according to the core configuration.

So EndpointActions behave like this:

```go
// Runs Authenticate with the core configuration, returns an error if untrusted.
rest.EndpointAction{Handler: endpointPost}

// Runs Authenticate with the core configuration, allows the request if trusted or untrusted.
rest.EndpointAction{Handler: endpointPost, AllowUntrusted: true}

// Runs myHandler, and if untrusted errors out. Then runs the core Authenticate, and if untrusted errors out.
rest.EndpointAction{Handler: endpointPost, AccessHandler: myHandler}

// Runs myHandler, and if untrusted errors out. Then runs the core Authenticate,  allows the request if untrusted.
rest.EndpointAction{Handler: endpointPost, AccessHandler: myHandler, AllowUntrusted: true}
```

I think it's useful to have `Authenticate` always run with the core configuration so we know whether an endpoint is trusted according to the default microcluster behaviour. 

If we get rid of the `AllowUntrusted` flag, that would require every endpoint to have an access handler. If an endpoint is supposed to be untrusted, we would need to define a no-op access handler that does not call `Authenticate` and also does not check `CtxAccess` populated by the internal `Authenticate` call. 

One thing I'm not sure of is what a list of access handlers would give us since you could just sequentially run the functions in an outer function anyway. It also makes it clearer to the writer that they need to either call a pre-defined access handler supplied by us or create their own, as opposed to specifying a list and digging through our code to see what the endpoint is doing with that list.

markylaing: > If we get rid of the `AllowUntrusted` flag, that would require every endpoint to have an access handler. If an endpoint is supposed to be untrusted, we would need to define a no-op access handler that does not call `Authenticate` and also does not check `CtxAccess` populated by the internal `Authenticate` call.

This is the same as the current `access.AllowAuthenticated` handler. I have a bone of contention with the naming of this handler because it doesn't actually check the value of `CtxAccess` and is a no-op. The naming assumes that the top level handler has already performed authentication.

> One thing I'm not sure of is what a list of access handlers would give us since you could just sequentially run the functions in an outer function anyway. It also makes it clearer to the writer that they need to either call a pre-defined access handler supplied by us or create their own, as opposed to specifying a list and digging through our code to see what the endpoint is doing with that list.

Not a list of access handlers, just a list of handlers. Microcluster is essentially a framework - this would allow easier creation of building blocks for composing sets of behaviour into the handlers. Information can be passed from one handler to the next via the request context. 

The core endpoints can follow the same pattern and continue to set `CtxAccess` in their own authentication middleware. Though I would additionally suggest that we expose these to be modified externally. As an example, as a user of microcloud, how do I change the authentication mechanism for core endpoints to use OIDC instead?

> ```go
> // Runs myHandler, and if untrusted errors out. Then runs the core Authenticate, and if untrusted errors out.
> rest.EndpointAction{Handler: endpointPost, AccessHandler: myHandler}
> 
> // Runs myHandler, and if untrusted errors out. Then runs the core Authenticate,  allows the request if untrusted.
> rest.EndpointAction{Handler: endpointPost, AccessHandler: myHandler, AllowUntrusted: true}
> ```

I don't think this is correct. Reading the code it looks like the top-level handler calls the core Authenticate function first, before calling the access handler. If we require that a user of microcluster sets `AllowUntrusted` to true and also set an `AccessHandler`, for all of these endpoint actions the call to `Authenticate` is superfluous.

masnax: > > ```go
> > // Runs myHandler, and if untrusted errors out. Then runs the core Authenticate, and if untrusted errors out.
> > rest.EndpointAction{Handler: endpointPost, AccessHandler: myHandler}
> > 
> > // Runs myHandler, and if untrusted errors out. Then runs the core Authenticate,  allows the request if untrusted.
> > rest.EndpointAction{Handler: endpointPost, AccessHandler: myHandler, AllowUntrusted: true}
> > ```
> 
> I don't think this is correct. Reading the code it looks like the top-level handler calls the core Authenticate function first, before calling the access handler. 

Oops, yes I had it backwards but the idea remains which is that `AllowUntrusted` basically controls whether we care about the core Authenticate call. 

> If we require that a user of microcluster sets `AllowUntrusted` to true and also set an `AccessHandler`, for all of these endpoint actions the call to `Authenticate` is superfluous.

This is the only part I disagree with. I think it's mainly a naming issue. I think of `AllowUntrusted` as `RequireCoreAuthentication` (with the boolean flipped) and the supplied access handler as a second layer of access on top. So it's not superfluous, since they check different things. What's superfluous today is the needless `AllowAuthenticated` which does nothing.

I'm happy to drop the boolean, I'm just suggesting that we still keep the separation between internal/external authentication and always run the internal authentication. The "default" access handler would effectively serve the same purpose as `AllowUntrusted` does today by actually checking the value of what the internal authentication determined.

markylaing: I've added a `hostAddress` argument to the `Authenticate` method so that it can be used in other contexts. The method now returns an error of type `access.ErrInvalidHost` if the host does not match the expected value. In the top-level handler, errors of this type are ignored so that the request can continue (the request is still considered untrusted from a core API perspective because the value of `trusted` in the request context is still false).

The `AllowAuthenticated` access handler is now always called when `AllowUntrusted` is false. This handler now checks the the `trusted` value in the request context.

This means that when running extra listeners via microcloud, downstreams must set `AllowUntrusted` to true on all endpoint actions and implement their own access handlers.

markylaing:
@tomponline @masnax @roosterfish this is ready for review when you have time.

internal/rest/rest.go
```diff
@@ -23,37 +24,23 @@ import (
...
 		return response.NotImplemented(nil)
 	}
 
-	if action.AccessHandler != nil {
-		// Defer access control to custom handler.
-		accessResp := action.AccessHandler(state, r)
-		if accessResp != response.EmptySyncResponse {
-			return accessResp
+	// If allow untrusted is not set, the request must be authenticated via core authentication (e.g. certificate in truststore).
+	if !action.AllowUntrusted {
```
mas-who, line 32: If the downstream app opt to implement only oidc authentication, do we then set every endpoint to AllowUntrusted=true and then pass a custom AccessHandler?

Unrelated to this line, is it okay if we pass w http.ResponseWriter to the handler functions? Looking at implementing a oidc/login endpoint in lxd-site-manager where I need to set some cookies on the response.

markylaing: > If the downstream app opt to implement only oidc authentication, do we then set every endpoint to AllowUntrusted=true and then pass a custom AccessHandler?

Yep that's how it would work currently. The discussion above was around this. The `AllowUntrusted` field now means "allow requests that are not TLS or are TLS but where the public key of he requestor is not in the microcluster truststore".

For this, you can use a `response.ManualResponse`. I'll show you an example when we have a call on Monday :)

mas-who: Amazing, thanks for the clarification! I will have a look at `response.ManualResponse` so long :+1:

internal/daemon/daemon.go
```diff
@@ -433,6 +437,17 @@ func (d *Daemon) StartAPI(bootstrap bool, initConfig map[string]string, newConfi
 			return err
 		}
 
+		// Start all extension servers before post-bootstrap hook.
+		for _, extensionServer := range d.extensionServers {
+			server := d.initServer(extensionServer.Resources...)
+			url := api.NewURL().Scheme(extensionServer.Protocol).Host(extensionServer.Address.String())
+			network := endpoints.NewNetwork(d.shutdownCtx, endpoints.EndpointNetwork, server, *url, extensionServer.Certificate)
+			err = d.endpoints.Add(network)
```
mas-who, line 445: Related to our discussion from Monday, since now we may define the address and path of every endpoint, we can essentially get around the `/1.0` prefix restriction correct? Maybe that's a good enough approach for lxd site manager. Originally we were thinking of grouping some of the endpoints (management API e.g. `/oidc/*`, `/sites`) with the core API but maybe we can do with an additional API?

internal/daemon/daemon.go
```diff
@@ -433,6 +437,17 @@ func (d *Daemon) StartAPI(bootstrap bool, initConfig map[string]string, newConfi
 			return err
 		}
 
+		// Start all extension servers before post-bootstrap hook.
+		for _, extensionServer := range d.extensionServers {
```
mas-who, line 441: this loop looks identical to the one below just before the post join hook. Maybe this can become a function?

rest/rest.go
```diff
@@ -36,3 +39,20 @@ type Endpoint struct {
...
+	Endpoints []Endpoint
+}
+
+// EndpointType is a type specifying the endpoint on with the resource exists.
+type EndpointType client.EndpointType
+
+// Server contains configuration and handlers for additional listeners to be instantiated after app startup.
+type Server struct {
+	Protocol    string
+	Address     types.AddrPort
```
mas-who, line 55: Question about the address, if a downstream app opt in to setup its own API server, this address would need to be a user configured setting correct?

internal/daemon/daemon.go
```diff
@@ -550,6 +565,17 @@ func (d *Daemon) StartAPI(bootstrap bool, initConfig map[string]string, newConfi
 		return err
 	}
 
+	// Start all extension servers before post-join hook.
+	for _, extensionServer := range d.extensionServers {
+		server := d.initServer(extensionServer.Resources...)
+		url := api.NewURL().Scheme(extensionServer.Protocol).Host(extensionServer.Address.String())
+		network := endpoints.NewNetwork(d.shutdownCtx, endpoints.EndpointNetwork, server, *url, extensionServer.Certificate)
+		err = d.endpoints.Add(network)
```
masnax, line 573: FWIW, this function is variadic so you can aggregate each item returned from `NewNetwork` and run `d.endpoints.Add(networks...)`.

rest/access/handlers.go
```diff
@@ -34,7 +44,7 @@ func Authenticate(state *state.State, r *http.Request, trustedCerts map[string]x
 	}
 
 	switch r.Host {
-	case state.Address().URL.Host:
+	case hostAddress:
```
masnax, line 61: Should we do any validation here to check if `hostAddress` is an actual address, and not just an empty string?

internal/daemon/daemon.go
markylaing: Yep but as discussed this would mean having three listeners, one for the microcluster core API, one for the site management API, and one for the heartbeat API.

rest/access/handlers.go
markylaing: Yeah I suppose this will give a clearer error message if a downstream doesn't use the function correctly.

internal/daemon/daemon.go
```diff
@@ -96,7 +98,7 @@ func NewDaemon(project string) *Daemon {
 // - `extensionsAPI` is a list of endpoints to be served over `/1.0`.
 // - `extensionsSchema` is a list of schema updates in the order that they should be applied.
 // - `hooks` are a set of functions that trigger at certain points during cluster communication.
-func (d *Daemon) Run(ctx context.Context, listenPort string, stateDir string, socketGroup string, extensionsAPI []rest.Endpoint, extensionsSchema []schema.Update, apiExtensions []string, hooks *config.Hooks) error {
+func (d *Daemon) Run(ctx context.Context, listenPort string, stateDir string, socketGroup string, extensionsAPI []rest.Endpoint, extensionsSchema []schema.Update, apiExtensions []string, extensionServers []rest.Server, hooks *config.Hooks) error {
```
roosterfish, line 101: The docs are missing a line for `extensionServers`. It looks this was also forgotten for `apiExtensions` but unrelated to this PR :)

internal/daemon/daemon.go
markylaing: No worries :)

internal/daemon/daemon.go
markylaing: Fixed.

rest/access/handlers.go
markylaing: Fixed.

roosterfish (changes requested):
Looking good, just some small comments.

rest/rest.go
```diff
@@ -36,3 +36,6 @@ type Endpoint struct {
 	AllowedDuringShutdown bool // Whether we should return Unavailable Error (503) if daemon is shutting down.
 	AllowedBeforeInit     bool // Whether we should return Unavailabel Error (503) if the daemon has not been initialized (is not yet part of a cluster).
 }
+
+// EndpointType is a type specifying the endpoint on with the resource exists.
```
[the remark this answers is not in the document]
markylaing: Fixed.

internal/daemon/daemon.go
markylaing: Fixed.

roosterfish (approved):

masnax (approved):
LGTM! thanks

PR #131, opened 2024-05-22
Adds `.golanci.yml` config file.

markylaing:
Uses the same rules as in LXD. I wanted to add this now before this repo grows in size any further.

Additionally:
- Disables ignoring lint errors in CI.
- Removes running revive directly (this is handled by `golangci-lint`).
- Removes shellcheck (no scripts to lint).


masnax (approved):
LGTM

internal/rest/resources/sql.go
```diff
@@ -103,7 +103,7 @@ func sqlSelect(ctx context.Context, tx *sql.Tx, query string, result *types.SQLR
 		return fmt.Errorf("Failed to execute query: %w", err)
 	}
 
-	defer rows.Close()
```
roosterfish, line 106: Would it makes sense to also log an error like in https://github.com/canonical/microcluster/pull/131/commits/de53cdc8a7143bad2a468836c272701a79c6a87a? 

There were some more commits before that added `//nolint:errcheck` instead of logging an error.

.golangci.yml
```diff
@@ -0,0 +1,117 @@
...
+      - default
+      - prefix(github.com/canonical/microcluster)
+  errcheck:
+    exclude-functions:
+      - (io.ReadCloser).Close
+      - (io.WriteCloser).Close
+      - (io.ReadWriteCloser).Close
+      - (*os.File).Close
+      - (*github.com/gorilla/websocket.Conn).Close
+      - (*github.com/mdlayher/vsock.Listener).Close
```
roosterfish, line 24: Should we add some comments explaining why those are excluded?

roosterfish (changes requested):
Looks good, only two small suggestions.

internal/rest/resources/sql.go
markylaing: Yep can do.

.golangci.yml
markylaing: Ah yes these were commonly ignored errors in LXD. I'll double check these.

internal/rest/resources/sql.go
markylaing: I've logged all deferred errors except one that is in a test case (where I've added a comment explaining why it is ignored with a lint rule).

.golangci.yml
markylaing: I've removed all of these except for `(io.ReadCloser).Close` which is used for closing response bodies. We can discuss adding more again in future.

markylaing:
@roosterfish ready for another round when you have time. Thanks.

PR #132, opened 2024-05-29
Alternative to #123

masnax:
This takes a slightly simpler approach to solving the issue in #123 which doesn't increase the number of cross-node communications over the network.

Rather than relying on a second notification, we can simply apply the first set of API extensions to the database as soon as we run the update that introduces API extensions. 

This means we won't have to wait 30s for the upgrade notifier to time out because it detects an API extension mismatch on 1 node after applying all schema updates.


masnax:
@tomponline I think this one might be a better approach to solving the problem in #123, since that one adds another back-and-forth connection between each cluster member.

internal/db/update/update.go
```diff
@@ -68,7 +69,18 @@ func (m *SchemaUpdateManager) AppendSchema(schemaExtensions []schema.Update, api
 	m.apiExtensions = apiExtensions
 }
 
+// updateFromV3 auto-applies the initial set of API extensions to the internal_cluster_members table.
+// This is done so that the cluster won't have to notified twice,
```
roosterfish, line 73: `to be notified`?

internal/db/update/update.go
```diff
@@ -58,10 +64,23 @@ func (m *SchemaUpdateManager) Schema() *SchemaUpdate {
 	return schema
 }
 
-func (m *SchemaUpdateManager) AppendSchema(extensions []schema.Update) {
-	m.updates[updateExternal] = extensions
+func (m *SchemaUpdateManager) AppendSchema(schemaExtensions []schema.Update, apiExtensions extensions.Extensions) {
```
roosterfish, line 67: Let's try and make the name of the receiver var consistent. Right now it's sometimes called `m` and sometimes `s`. What do you think about choosing `s`? :)

roosterfish (changes requested):
It's much simpler to follow than #123. Only two small things I found.

However I am wondering if it would be worth merging the `updateFromV3` into `updateFromV2` as they belong together?
I understand it might not apply the patch for any MicroCluster out there that is tracking our `main` branch as `v2` might already be applied. So probably too risky as we have never cut any release yet.

tomponline:
Once this is a rebased and approved by @roosterfish ill take a quick pass @mseralessandri

roosterfish (approved):
Thanks, still looks good to me.

I was about to suggest having an extra test for the `updateFromV3()` func but this is called under to hood in `Test_updateFromV1ClusterMembers()` when calling `_, err = schemaMgr.Schema().Ensure(db)` so I think it's fine like this.

tomponline (approved):
This seems much simpler than the other one. Thanks

PR #137, opened 2024-06-06
Rename `internal` to `core`

### LINKED ISSUE #108: Rename confusing endpoints
In particular, `/cluster/1.0/cluster` is a bit stammering, so we can rename it to `/cluster/1.0/members` instead.

`/cluster` is the start point for built-in microcluster API endpoints
`/cluster/1.0` is the start point for all publicly accessible microcluster API endpoints (issuing tokens, listing cluster members, etc.).
`/cluster/1.0/cluster` is the subset of those endpoints that manages cluster formation.

In particular, these endpoints are

`POST /cluster/1.0/cluster` -- adds a cluster member to dqlite
`GET /cluster/1.0/cluster` -- lists existing cluster member information
`DELETE /cluster/1.0/cluster/{name}` -- deletes the cluster member with the given name.

Since all of these pertain to cluster *members*, we can simply rename the endpoint to `/cluster/1.0/members` to be a bit less confusing. 

This would be an API change and should happen with a new API extension.

masnax:
Closes #108 

* Adds a new schema update that renames the tables `internal_cluster_members` and `internal_token_records` to `core_cluster_members` and `core_token_records`. This is so that it is clear that these tables should be accessed by downstream projects and are not an internal implementation detail.

* Updates all DB functions appropriately

* Since there is a cyclical issue with the schema update applying to the table that we use to check schema updates, we need to dynamically grab the table name in case it hasn't been renamed yet, so there is a new helper `getClusterTableName` to fetch it from `sqlite_master`.

* Renames the `/cluster` endpoints to `/core`. This has the added effect of fixing the `/cluster/1.0/cluster` stammering, now it's `/core/1.0/cluster`. 

* Moves `types.ClusterMember` and `types.ClusterMemberLocal` out of the `internal` package so that they are accessible by downstream projects.

* Since `types.ClusterMemberLocal` is exported, we can include it as an argument to the `OnNewMember` hook so that when that hook triggers, it is immediately available which cluster member got added without needing another transaction.

### DIFF
--- a/cluster/cluster_members.go
+++ b/cluster/cluster_members.go
@@ -17,21 +17,21 @@ import (
 //go:generate -command mapper lxd-generate db mapper -t cluster_members.mapper.go
 //go:generate mapper reset
 //
-//go:generate mapper stmt -e internal_cluster_member objects table=internal_cluster_members
-//go:generate mapper stmt -e internal_cluster_member objects-by-Address table=internal_cluster_members
-//go:generate mapper stmt -e internal_cluster_member objects-by-Name table=internal_cluster_members
-//go:generate mapper stmt -e internal_cluster_member id table=internal_cluster_members
-//go:generate mapper stmt -e internal_cluster_member create table=internal_cluster_members
-//go:generate mapper stmt -e internal_cluster_member delete-by-Address table=internal_cluster_members
-//go:generate mapper stmt -e internal_cluster_member update table=internal_cluster_members
+//go:generate mapper stmt -e core_cluster_member objects table=core_cluster_members
+//go:generate mapper stmt -e core_cluster_member objects-by-Address table=core_cluster_members
+//go:generate mapper stmt -e core_cluster_member objects-by-Name table=core_cluster_members
+//go:generate mapper stmt -e core_cluster_member id table=core_cluster_members
+//go:generate mapper stmt -e core_cluster_member create table=core_cluster_members
+//go:generate mapper stmt -e core_cluster_member delete-by-Address table=core_cluster_members
+//go:generate mapper stmt -e core_cluster_member update table=core_cluster_members
 //
-//go:generate mapper method -i -e internal_cluster_member GetMany table=internal_cluster_members
-//go:generate mapper method -i -e internal_cluster_member GetOne table=internal_cluster_members
-//go:generate mapper method -i -e internal_cluster_member ID table=internal_cluster_members
-//go:generate mapper method -i -e internal_cluster_member Exists table=internal_cluster_members
-//go:generate mapper method -i -e internal_cluster_member Create table=internal_cluster_members
-//go:generate mapper method -i -e internal_cluster_member DeleteOne-by-Address table=internal_cluster_members
-//go:generate mapper method -i -e internal_cluster_member Update table=internal_cluster_members
+//go:generate mapper method -i -e core_cluster_member GetMany table=core_cluster_members
+//go:generate mapper method -i -e core_cluster_member GetOne table=core_cluster_members
+//go:generate mapper method -i -e core_cluster_member ID table=core_cluster_members
+//go:generate mapper method -i -e core_cluster_member Exists table=core_cluster_members
+//go:generate mapper method -i -e core_cluster_member Create table=core_cluster_members
+//go:generate mapper method -i -e core_cluster_member DeleteOne-by-Address table=core_cluster_members
+//go:generate mapper method -i -e core_cluster_member Update table=core_cluster_members
 
 // Role is the role of the dqlite cluster member.
 type Role string
@@ -39,8 +39,8 @@ type Role string
 // Pending indicates that a node is about to be added or removed.
 const Pending Role = "PENDING"
 
-// InternalClusterMember represents the global database entry for a dqlite cluster member.
-type InternalClusterMember struct {
+// CoreClusterMember represents the global database entry for a dqlite cluster member.
+type CoreClusterMember struct {
 	ID             int
 	Name           string `db:"primary=yes"`
 	Address        string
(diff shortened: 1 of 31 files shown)


masnax:
@roosterfish I've fixed the table name query here, so this should also be ready for review, thanks.

cluster/cluster_members.go
```diff
@@ -52,15 +52,15 @@ type InternalClusterMember struct {
...
+// CoreClusterMemberFilter is used for filtering queries using generated methods.
+type CoreClusterMemberFilter struct {
 	Address *string
 	Name    *string
 }
 
 // ToAPI returns the api struct for a ClusterMember database entity.
 // The cluster member's status will be reported as unreachable by default.
-func (c InternalClusterMember) ToAPI() (*internalTypes.ClusterMember, error) {
+func (c CoreClusterMember) ToAPI() (*internalTypes.ClusterMember, error) {
```
roosterfish: As part of this should we also align the name on the Go level and rename `internal/rest/types` to `core/rest/types`?

cluster/cluster_members.mapper.go
```diff
@@ -17,57 +17,57 @@ import (
 var _ = api.ServerEnvironment{}
 
 var internalClusterMemberObjects = RegisterStmt(`
```
roosterfish, line 19: I think those vars require a name update too?

cluster/cluster_members.mapper.go
```diff
@@ -17,57 +17,57 @@ import (
...
-UPDATE internal_cluster_members
+UPDATE core_cluster_members
   SET name = ?, address = ?, certificate = ?, schema_internal = ?, schema_external = ?, api_extensions = ?, heartbeat = ?, role = ?
  WHERE id = ?
 `)
 
 // internalClusterMemberColumns returns a string of column names to be used with a SELECT statement for the entity.
-// Use this function when building statements to retrieve database entries matching the InternalClusterMember entity.
+// Use this function when building statements to retrieve database entries matching the CoreClusterMember entity.
 func internalClusterMemberColumns() string {
```
roosterfish, line 61: `coreClusterMemberColumns()` ?

cluster/cluster_members.go
```diff
@@ -188,3 +208,19 @@ func GetClusterMemberAPIExtensions(ctx context.Context, tx *sql.Tx) ([]extension
 
 	return results, nil
 }
+
+// The cluster members table was renamed with updateFromV4, so we cannot be sure of its name prior to applying that update.
```
roosterfish, line 211: The comment is the reason for adding this helper but isn't precisely describing it's behavior.

cluster/cluster_members.go
```diff
@@ -188,3 +208,19 @@ func GetClusterMemberAPIExtensions(ctx context.Context, tx *sql.Tx) ([]extension
...
+func getClusterTableName(ctx context.Context, tx *sql.Tx) (string, error) {
+	stmt := "select name from sqlite_master where name = 'internal_cluster_members' or name = 'core_cluster_members'"
```
roosterfish, line 213: To be in line with the other queries can you please uppercase the SQL syntax?

roosterfish (changes requested):

cluster/cluster_members.go
roosterfish: Nevermind, you have already done this change in one of the later commits.

cluster/cluster_members.mapper.go
roosterfish: Those vars are still called `internal`. Should they be moved to start with `core` too? There are some functions too.

roosterfish (changes requested):
Thanks, I think there are just some few left over vars/function names that start with `internal` instead of `core`.

cluster/cluster_members.mapper.go
masnax: Ah oops, I put the changes in the wrong commit, fixed now :)

roosterfish (dismissed):
LGTM!

masnax:
I'm going to hold off on merging this until the microcluster schema fixes have been settled.

roosterfish:
@masnax should we include this in the LTS too?

It would make backporting into the new LTS easier if the names match.

masnax:
> @masnax should we include this in the LTS too?
> 
> It would make backporting into the new LTS easier if the names match.

Yes, it's a schema change so we should definitely add it to the LTS. I've just been holding off until we close a few MicroCloud PRs so that we don't have a ton of rebases and behavioural changes to work around for multiple PRs.

masnax:
@roosterfish This one is also ready for review. I've tested that updating from the version of microcluster in stable releases works, as well as from upstream microcluster.

internal/rest/client/client.go
```diff
@@ -26,6 +26,23 @@ import (
 	"github.com/canonical/microcluster/rest/types"
 )
 
+// EndpointType is a type specifying the endpoint on with the resource exists.
```
roosterfish, line 29: on which

internal/extensions/extensions.go
```diff
@@ -29,6 +29,7 @@ type Extensions []string
 // Populate internal extensions here.
 var internalExtensions = Extensions{
 	"internal:runtime_extension_v1",
+	"internal:rename_core_endpoints",
```
roosterfish, line 32: As we are about to release the LTS, having this API extension is optional right?

roosterfish (changes requested):
Still looks good, only one typo and a comment about the API extension which should not be a blocker.

roosterfish (approved):

internal/rest/client/client.go
roosterfish: Feel free to fix this in a follow up PR to get this one unblocked :)

PR #138, opened 2024-06-06
Recover from join failures

masnax:
Reverts the joiner and cluster if there is an error during the join process.

* Instead of iterating over every join address to find which system initially consumed our join token, the address of this system is now included in the response returned by validating the join token.
* Allows reducing a 2 system cluster to a 1 system cluster, as there is no issue with quorum since the roles are turned over first.
* If deleting a node with `?force=true`, then errors related to reaching the node will be ignored as we are forcibly removing it and it may be unreachable.
* If the joiner fails to join the cluster, a revert block is executed which resets the state of the joiner back to pre-initialization, and additionally attempts to send a request to the existing cluster to clear out any remaining state that may persist on the cluster side.

### DIFF
--- a/internal/rest/resources/cluster.go
+++ b/internal/rest/resources/cluster.go
@@ -480,7 +480,7 @@ func clusterMemberDelete(s *state.State, r *http.Request) response.Response {
 
 	// Tell the cluster member to run its PreRemove hook and return.
 	err = internalClient.RunPreRemoveHook(ctx, c.UseTarget(name), internalTypes.HookRemoveMemberOptions{Force: force})
-	if err != nil {
+	if err != nil && !force {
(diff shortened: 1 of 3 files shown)


masnax:
@roosterfish I've added your suggestions:

* The token API type fields now have comments explaining what they are there for, including `TrustedMember`.
* The revert block now re-uses the logic from `ResetClusterMember`. In order to get this to work, I had to split it out into its own function to decouple it from the `response.Response` type. Additionally, it now returns a function that handles the re-execing externally, so that we can block the `clusterDisableMu` only when we need to.

roosterfish (approved):
Thanks, all clear now to me :)

PR #142, opened 2024-06-10
Replace `extendedEndpoints` with `extensionServers` for registering core endpoints [WD-11978]

### LINKED ISSUE #135: Allow core API extensions that do not begin with `/1.0`.
LXD site manager will need to implement OIDC on the core API. To do this, some handlers need to be added to handle the auth flow. Currently it (appears to be) only possible to add handlers to `/1.0`, whereas it would be ideal to add these handlers under `/oidc` (e.g. `/oidc/login` as in LXD).

One way to do this would be to handle setting `Path` field in `rest.Endpoint` to a full path, e.g. with a leading forward slash.

mas-who:
## Done
 - Allow users to define core api endpoints without /1.0 prefix using `extensionServers`
 - Validate user defined endpoints to not conflict with internal endpoints
 - Allow endpoints defined in `ExtensionServer` to be registered before cluster initialisation
 - Removed `extendedEndpoints` in favour of `extensionServer` for registering core listener endpoints

## Context
 - Aimed to resolve #135


mas-who:
@masnax Instead of taking down the network, I decided to extract all the core endpoints from `ExtensionServer` and grouping them with the pre-defined core endpoints. Do you see any problems with this approach?

internal/rest/resources/resources.go
```diff
@@ -46,3 +49,47 @@ var ExtendedEndpoints = rest.Resources{
...
+			return fmt.Errorf("Endpoint %q conflicts with internal endpoint", url)
+		}
+	}
+
+	return nil
+}
+
+// MergeExtendedEndpoints merges the endpoints from extensionServers with endpoints defined in extendedEndpoints.
+// It also ensures that there are no conflicts between endpoints from extensionServers and internal endpoints.
+func MergeExtendedEndpoints(extensionServers []rest.Server, extendedEndpoints rest.Resources) ([]rest.Resources, error) {
```
mas-who, line 77: Is there a better place to keep `checkDuplicateEndpoints` and `MergeExtendedEndpoints`?

rest/rest.go
```diff
@@ -51,6 +51,7 @@ type EndpointType client.EndpointType
 
 // Server contains configuration and handlers for additional listeners to be instantiated after app startup.
 type Server struct {
+	CoreAPI     bool
```
mas-who, line 54: Thought about checking if the `Address` is specified to determine if a `Server` should be grouped with the core API, but that complicates things since the address of a network listener should be set in the `PreBootStrap` or `PreJoin` hooks. Having the `CoreAPI` flag makes the intention explicit that the server and its endpoints should be added to the core API network listener.

internal/daemon/daemon.go
```diff
@@ -69,7 +69,8 @@ type Daemon struct {
 	// stop is a sync.Once which wraps the daemon's stop sequence. Each call will block until the first one completes.
 	stop func() error
 
-	extensionServers []rest.Server
+	extensionServers       []rest.Server
```
masnax: Looks like you also remove this in the following commit.

internal/rest/resources/resources.go
masnax: I think there should be fine. It's under `internal` so that's really the biggest concern :)

rest/rest.go
```diff
@@ -38,6 +38,7 @@ type Endpoint struct {
 
 	AllowedDuringShutdown bool // Whether we should return Unavailable Error (503) if daemon is shutting down.
 	AllowedBeforeInit     bool // Whether we should return Unavailabel Error (503) if the daemon has not been initialized (is not yet part of a cluster).
+	AllowNoPrefix         bool // Whether we should allow the endpoint to be registered without the /1.0 prefix.
```
masnax: Looks like this commit is redundant now since you undo the change in a later commit. Could you please clean it up?

rest/rest.go
masnax, line 54: Could you also add some validation here that ensures `Address`, `Protocol`, and `Certificate` are unset if `CoreAPI` is true?

masnax (changes requested):
Thanks @mas-who, looks great overall :)

Main thing that caught my eye is that there are a couple commits that introduce changes that are then undone, so if you could clean up the commits to remove the redundancy, that would be great.


As a side note, I'm wondering if it even makes sense to have the `extendedEndpoints` list anymore. I feel like having two ways to supply endpoints that behave the same is a bit confusing and unnecessary.

Perhaps we can drop `extendedEndpoints` and just have a single `extensionServers` list, with the core endpoints specified where `CoreAPI=true`.

mas-who:
> Thanks @mas-who, looks great overall :)
> 
> Main thing that caught my eye is that there are a couple commits that introduce changes that are then undone, so if you could clean up the commits to remove the redundancy, that would be great.

Sorry about that, my intention was to present the departure from the original approach that we discussed. I will clean up the commits and moving forward I will make sure that commits are representative of the final PR state.

mas-who:
> As a side note, I'm wondering if it even makes sense to have the `extendedEndpoints` list anymore. I feel like having two ways to supply endpoints that behave the same is a bit confusing and unnecessary.
> 
> Perhaps we can drop `extendedEndpoints` and just have a single `extensionServers` list, with the core endpoints specified where `CoreAPI=true`.

Sure thing, I can make this change too as part of this PR. Just want to confirm that is it okay to remove `extendedEndpoints` since many downstream apps would be using this already?

mas-who:
I have removed `extendedEndpoints` and adjusted the example accordingly. You can check the last two commits on this PR. We can also split out those changes specifically into a separate PR.

internal/daemon/daemon.go
mas-who, line 1: this changes the order of route definitions so that if a handler was defined for `/` in a downstream app, that handler would be used instead.

context:
In site manager, if `/` is accessed from a browser client, we want to redirect that request to `/ui`. This is also implemented in LXD.

internal/daemon/daemon.go
```diff
@@ -328,6 +328,21 @@ func (d *Daemon) initServer(resources ...rest.Resources) *http.Server {
 	mux.SkipClean(true)
 	mux.UseEncodedPath()
 
+	state := d.State()
```
mas-who, line 331: this changes the order of route definitions so that if a handler was defined for / in a downstream app, that handler would be used instead.

context:
In site manager, if / is accessed from a browser client, we want to redirect that request to /ui. This is also implemented in LXD.

rest/rest.go
```diff
@@ -53,7 +54,18 @@ type EndpointType client.EndpointType
...
 	Certificate *shared.CertInfo
 	Resources   []Resources
 }
+
+// ValidateServerConfigs checks that the server configuration is valid.
+func (s Server) ValidateServerConfigs() error {
+	if s.CoreAPI {
+		if s.Address != nil || s.Protocol != "" || s.Certificate != nil {
+			return fmt.Errorf("Core API server cannot have Address, Protocol or Certificate")
+		}
```
masnax, line 67: So it seems this PR will also allow an upstream project to specify their own path prefix for core endpoints, am I right about that?

If so, then I think we need some more validation there. I have a couple questions:
* Should we allow multiple sets of resources for the core API, or verify that the list is exactly 1 entry long?
* If we should allow multiple sets of resources, should they all have the same path prefix?
* Should we enforce that `Resources.Path` must be set? It's ambiguous what happens right now if it's unset.

rest/rest.go
```diff
@@ -53,7 +54,18 @@ type EndpointType client.EndpointType
 type Server struct {
 	CoreAPI     bool
 	Protocol    string
-	Address     types.AddrPort
+	Address     *types.AddrPort
```
masnax, line 57: Does this need to be a pointer? I believe it would just be instantiated as an empty struct, which you can verify with `s.Address == types.AddrPort{}` in the validator.

masnax (changes requested):
Looking much cleaner now, thank you!

I've left a comment about validating the path prefix on the resources for the core API, please let me know your thoughts.

Additionally, it seems the field defined as [ExtendedEndpoint](https://github.com/canonical/microcluster/blob/817979f5a17227cdbed5a40dc19885cd8eeb0aef/internal/rest/client/client.go#L32) is now unused except for [Query](https://github.com/canonical/microcluster/blob/817979f5a17227cdbed5a40dc19885cd8eeb0aef/client/client.go#L25-L30), but since we don't actually validate that the core API must use `1.0`, this function is now incorrect. I think we can remove both and just allow using `QueryStruct` which is called by `Query` anyway.

rest/rest.go
mas-who: Yes, an upstream project can now define its own path prefix for core endpoints. To extend on this point a bit, since `extendedEndpoints` together with `resources.ExtendedEndpoints` is now removed, an upstream project would have to define its own path prefixes. The mental model here is to give more flexibility to the upstream project in shaping its own core routing structure.

Agreed on more validation is required, here are my thoughts let me know if you agree :slightly_smiling_face: 
- We should allow multiple sets of resources since we want to allow an upstream project to also define endpoints that does not have the `/1.0` prefix. Perhaps the restriction here should be only one `Server` with `CoreAPI=true` is allowed? I think we can be opinionated here since every `Server` with `CoreAPI=true` will be assigned to the same network listener anyway.
- I don't think we should enforce same path prefix at the `Resource` level if multiple resources are defined. A `Resource` should represent a logical grouping of endpoints that sit behind a path prefix. A relevant validation here could be to ensure duplicated path prefixes does not exist, this will also ensure no duplicated core endpoints exist.
- If `Resources.Path` is not set, then the path prefix becomes `/` which can be used to setup public endpoints like `/ui`, `/oidc` etc. I do agree that this is not immediately clear, so maybe we can enforce that `Resources.Path` must be set and if root prefix is required, it should be set to `Resources.Path="/"`? We would have to adjust [`internal.rest.HandleEndpoint`](https://github.com/canonical/microcluster/blob/df5dbb210195b8053b4940677350aa5cbe909897/internal/rest/rest.go#L152) to cater for this.

rest/rest.go
mas-who: aah yes, pointer semantics doesn't make a whole lot of sense here, I will update it.

mas-who:
> Additionally, it seems the field defined as [ExtendedEndpoint](https://github.com/canonical/microcluster/blob/817979f5a17227cdbed5a40dc19885cd8eeb0aef/internal/rest/client/client.go#L32) is now unused except for [Query](https://github.com/canonical/microcluster/blob/817979f5a17227cdbed5a40dc19885cd8eeb0aef/client/client.go#L25-L30), but since we don't actually validate that the core API must use `1.0`, this function is now incorrect. I think we can remove both and just allow using `QueryStruct` which is called by `Query` anyway.

I see I have missed those related dependencies of `ExtendedEndpoint`, will remove `ExtendedEndpoint` and `Query` as suggested and adjust the example accordingly.

rest/rest.go
mas-who: - On point 2 I meant resources in a server. Like you mentioned we should allow same endpoints across different network listeners.
- Agreed on point 3, it's clearer to rename the field instead of a special case. I have validated the behaviour of setting `""` for the path prefix in this site manager [PR](https://github.com/canonical/lxd-site-manager/pull/15) (tested locally).

mas-who:
@masnax I've added those additional validations for `extensionServers`.

example/api/servers.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"github.com/canonical/microcluster/rest"
+)
+
+// Servers represents the list of listeners that the daemon will start
+// Each Server has pre-defined endpoints that will be added to the listener
+// If the Server is marked as CoreAPI, its endpoints will be added to the core listener of Microcluster.
+var Servers = []rest.Server{
+	{
+		CoreAPI: true,
```
roosterfish, line 13: What is the use case for having the `CoreAPI` field, if you can also define the endpoints as part of running `(m *MicroCluster) Start()` by providing the `extensionsAPI` slice?
 
Wouldn't both end up on the same listener?

mas-who: We are removing `extensionsAPI` with this PR. The idea is that we will use `extensionServers` for defining core API listener as well as additional listeners.

masnax:
@mas-who #155 is now merged, so you can rebase your changes and replace the hardcoded ["1.0"](https://github.com/canonical/microcluster/blob/main/example/client/client.go) with a constant that is defined and passed to the `Resources` of the `CoreAPI` server.

mas-who:
> @mas-who #155 is now merged, so you can rebase your changes and replace the hardcoded ["1.0"](https://github.com/canonical/microcluster/blob/main/example/client/client.go) with a constant that is defined and passed to the `Resources` of the `CoreAPI` server.

Thanks @masnax , on it :+1:

mas-who:
> @mas-who #155 is now merged, so you can rebase your changes and replace the hardcoded ["1.0"](https://github.com/canonical/microcluster/blob/main/example/client/client.go) with a constant that is defined and passed to the `Resources` of the `CoreAPI` server.

@masnax I've made this change.

masnax (approved):
LGTM, thanks :)

example/api/servers.go
roosterfish: Thanks, clear now. 

When adding this comment I haven't yet seen the commit that removes the `extensionsAPI` argument from the `Start()` func. The overall change wasn't fully clear to me by just reading the description of the PR :)

PR #144, opened 2024-06-11
Use an interface for `State`

### LINKED ISSUE #133: Make the state parameter read-only in handlers and hooks for critical fields.
The state parameter that is passed into all API handlers is currently instantiated [here](https://github.com/canonical/microcluster/blob/d23c4306e691a985c34f6d8b4faa4a9ef5d08dfc/internal/daemon/daemon.go#L333). 

Instantiating it here means that the pointer contains the same memory address in all handlers. The state itself contains fields that are direct pointers from the daemon struct. This means that it is possible to mutate fields on the `Daemon` struct from within handlers. These fields are critical to the functioning of the cluster and should not be modifiable by clients. For example, one could set `(*state.State).Database = nil`:
```golang
var breakThings = rest.Endpoint{
	Path: "break-things",
	Get: rest.EndpointAction{
		Handler: func(s *state.State, r *http.Request) response.Response {
			s.Database = nil
			return response.SyncResponsePlain(true, false, fmt.Sprintf("%p\n", s))
		},
		AllowUntrusted: true,
	},
}
```
```
mark@RUBIX:~/projects/canonical/lxd-site-manager$ curl -k https://127.0.0.1:5000/1.0/break-things
0xc0000f8870
mark@RUBIX:~/projects/canonical/lxd-site-manager$ curl -k https://127.0.0.1:5000/1.0/break-things
curl: (92)2024/06/04 12:34:35 http2: panic serving 127.0.0.1:45384: runtime error: invalid memory address or nil pointer dereference
goroutine 111 [running]:
net/http.(*http2serverConn).runHandler.func1()
        /home/mark/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.3.linux-amd64/src/net/http/h2_bundle.go:6185 +0x145
panic({0xa698a0?, 0xf486e0?})
        /home/mark/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.3.linux-amd64/src/runtime/panic.go:770 +0x132
github.com/canonical/microcluster/internal/rest.HandleEndpoint.func1({0xbe4e20, 0xc0000be088}, 0xc000253560)
        /home/mark/projects/canonical/microcluster/internal/rest/rest.go:165 +0x129
net/http.HandlerFunc.ServeHTTP(0xc000253440?, {0xbe4e20?, 0xc0000be088?}, 0x0?)
        /home/mark/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.3.linux-amd64/src/net/http/server.go:2166 +0x29
github.com/gorilla/mux.(*Router).ServeHTTP(0xc00046a000, {0xbe4e20, 0xc0000be088}, 0xc000252fc0)
        /home/mark/go/pkg/mod/github.com/gorilla/mux@v1.8.1/mux.go:212 +0x1e2
net/http.serverHandler.ServeHTTP({0x0?}, {0xbe4e20?, 0xc0000be088?}, 0x0?)
        /home/mark/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.3.linux-amd64/src/net/http/server.go:3137 +0x8e
net/http.initALPNRequest.ServeHTTP({{0xbe5d98?, 0xc0000bbc50?}, 0xc00048a708?, {0xc000474000?}}, {0xbe4e20, 0xc0000be088}, 0xc000252fc0)
        /home/mark/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.3.linux-amd64/src/net/http/server.go:3745 +0x231
net/http.(*http2serverConn).runHandler(0x0?, 0x0?, 0x0?, 0x0?)
        /home/mark/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.3.linux-amd64/src/net/http/h2_bundle.go:6192 +0xbb
created by net/http.(*http2serverConn).scheduleHandler in goroutine 109
        /home/mark/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.3.linux-amd64/src/net/http/h2_bundle.go:6127 +0x21d
 HTTP/2 stream 0 was not closed cleanly: INTERNAL_ERROR (err 2)
mark@RUBIX:~/projects/canonical/lxd-site-manager$ 
```

That's contrived example, but my point is that this should not be exposed to clients, nor should changes to the state be persistent unless using explicit APIs.

masnax:
Closes #133 

* Replaces `state.State` with a new `state.InternalState` struct which implements a `state.State` interface. This means that the fields of `State` are not overwrite-able by the upstream project, and internal functions are not exposed externally.

* To avoid import cycles, `config.Hooks` is moved into the `state` package as well. This is actually not too bad since the `config` package was a confusing place for it. It also means we don't need to define functions and can just provide the whole type to `InternalState`

### DIFF
--- a/internal/state/state.go
+++ b/internal/state/state.go
@@ -2,6 +2,7 @@ package state
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/canonical/lxd/shared"
@@ -18,8 +19,44 @@ import (
 	"github.com/canonical/microcluster/rest/types"
 )
 
-// State is a gateway to the stateful components of the microcluster daemon.
-type State struct {
+// State exposes the internal daemon state for use with extended API handlers.
+type State interface {
+	// FileSystem structure.
+	FileSystem() *sys.OS
+
+	// Listen Address.
+	Address() *api.URL
+
+	// Name of the cluster member.
+	Name() string
+
+	// Server certificate is used for server-to-server connection.
+	ServerCert() *shared.CertInfo
+
+	// Cluster certificate is used for downstream connections within a cluster.
+	ClusterCert() *shared.CertInfo
+
+	// Database.
+	Database() *db.DB
+
+	// Local truststore access.
+	Remotes() *trust.Remotes
+
+	// Cluster returns a client to every cluster member according to dqlite.
+	Cluster(isNotification bool) (client.Cluster, error)
+
+	// Leader returns a client to the dqlite cluster leader.
+	Leader() (*client.Client, error)
+
+	// HasExtension returns whether the given API extension is supported.
+	HasExtension(ext string) bool
+
+	// ExtensionServers returns an immutable list of the daemon's additional listeners.
+	ExtensionServers() []string
+}
+
+// InternalState is a gateway to the stateful components of the microcluster daemon.
+type InternalState struct {
 	// Context.
 	Context context.Context
 
@@ -29,32 +66,11 @@ type State struct {
 	// ShutdownDoneCh receives the result of the d.Stop() function and tells the daemon to end.
 	ShutdownDoneCh chan error
 
-	// File structure.
-	OS *sys.OS
-
-	// Listen Address.
-	Address func() *api.URL
-
-	// Name of the cluster member.
-	Name func() string
-
-	// Server.
+	// Endpoints manages the network and unix socket listeners.
 	Endpoints *endpoints.Endpoints
-
-	// Server certificate is used for server-to-server connection.
-	ServerCert func() *shared.CertInfo
-
-	// Cluster certificate is used for downstream connections within a cluster.
-	ClusterCert func() *shared.CertInfo
 
 	// Local daemon's config.
 	LocalConfig func() *internalConfig.DaemonConfig
-
-	// Database.
-	Database *db.DB
-
-	// Remotes.
-	Remotes func() *trust.Remotes
 
 	// Initialize APIs and bootstrap/join database.
 	StartAPI func(bootstrap bool, initConfig map[string]string, newConfig *trust.Location, joinAddresses ...string) error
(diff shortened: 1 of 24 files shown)


internal/state/state.go
```diff
@@ -196,3 +197,28 @@ var StopListeners func() error
 
 // ReloadClusterCert reloads the cluster keypair from the state directory.
 var ReloadClusterCert func() error
+
+// Unrestrict returns the underlying InternalState from the exposed State interface.
+func Unrestrict(s State) (*InternalState, error) {
```
roosterfish, line 202: As there really isn't any restriction applied, I guess it would be better to rename this to something like `Resolve` or `ResolveInternalState`? Or `Derive`/`DeriveInternalState`.

internal/state/extended_state.go
```diff
@@ -0,0 +1,14 @@
+package state
+
+// ExtendedState represents an extension to the default state object.
+type ExtendedState interface {
```
roosterfish, line 4: I am bit confused by the naming. Is there a difference between "extended" and "external"? If I understand it right then if someone is using an extended state, from his point of view this is always external. 

So should this better be named `ExternalState` instead? 
Another thing we might think about is just calling it `External` as the package name already includes `state` but this might collide with the parent which has to be named `state.State`.

roosterfish (commented):
Some comments from my side on the naming so that we don't confuse "extended" with "external".

I like that you can now inject your own derivation of the state but I don't find it too intuitive that you also have to implement the `SetState()` and `GetState()` method every time by yourself.

Would it be an approach to do something like we have in LXD for storage drivers where some defaults are set/implemented in the [`common` struct](https://github.com/canonical/lxd/blob/main/lxd/storage/drivers/driver_common.go#L24) which you could then embed into your own state struct?

internal/state/state.go
masnax: I've called it `ToInternal` now.

masnax:
@roosterfish I've removed the extensions to the state for now, an just made this PR replace `State` with an interface.

roosterfish (approved):
Looks great, very clean and nice to see that we now have a clear separation between public facing funcs and the daemon's internal state.

PR #145, opened 2024-06-11
use cluster certificate if additional network listener certificate not provided [WD-11968]

mas-who:
## Done
 - Use cluster certificate for new network listeners when certificates are not provided

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -578,9 +578,14 @@ func (d *Daemon) StartAPI(bootstrap bool, initConfig map[string]string, newConfi
 func (d *Daemon) addExtensionServers() error {
 	var networks []endpoints.Endpoint
 	for _, extensionServer := range d.extensionServers {
+		cert := extensionServer.Certificate
+		if cert == nil {
+			cert = d.ClusterCert()
+		}
+
 		server := d.initServer(extensionServer.Resources...)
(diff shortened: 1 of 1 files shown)


internal/daemon/daemon.go
```diff
@@ -578,9 +578,13 @@ func (d *Daemon) StartAPI(bootstrap bool, initConfig map[string]string, newConfi
 func (d *Daemon) addExtensionServers() error {
 	var networks []endpoints.Endpoint
 	for _, extensionServer := range d.extensionServers {
+		cert := extensionServer.Certificate
+		if cert == nil {
+			cert = d.ClusterCert()
+		}
```
roosterfish, line 584: Should not be a blocker but the linter should normally complain that there is now newline after the `if` block. Something to follow up on our side :)

roosterfish (approved):
Thanks, looks good to me!

internal/daemon/daemon.go
markylaing: This is a [custom linter in the LXD repo](https://github.com/canonical/lxd/blob/main/test/lint/newline-after-block.sh) but it's not present here.

mas-who: I will put in the new line here, going to steal that linter for the site manager repo too :slightly_smiling_face:

roosterfish (approved):

PR #146, opened 2024-06-12
internal: Prevent Daemon startup if control socket is present

MggMuggins:
I'm having a hard time coming up with any situation where it's safe to start the daemon if a control socket still exists in the state dir.

I'm doing this check anyway as a part of cluster recovery.


masnax (approved):
LGTM, thanks :)

PR #147, opened 2024-06-14
internal/daemon: Don't error out if upgrade notification fails

masnax:
If the upgrade notification fails, that does not mean that we have to stop starting the local node, rather it is informing us of an issue on a remote cluster member, so we can ignore it during set up.

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -623,7 +623,7 @@ func (d *Daemon) sendUpgradeNotification(ctx context.Context, c *client.Client) 
 	}
 
 	if resp.StatusCode != http.StatusOK {
-		return fmt.Errorf("Database upgrade notification failed: %s", resp.Status)
+		logger.Errorf("Database upgrade notification failed: %s", resp.Status)
 	}
 
 	return nil


internal/daemon/daemon.go
```diff
@@ -623,7 +623,7 @@ func (d *Daemon) sendUpgradeNotification(ctx context.Context, c *client.Client)
 	}
 
 	if resp.StatusCode != http.StatusOK {
-		return fmt.Errorf("Database upgrade notification failed: %s", resp.Status)
+		logger.Errorf("Database upgrade notification failed: %s", resp.Status)
```
roosterfish, line 626: What is the drawback of having this (if any)?

masnax: So if we do error out here, the daemon will fail to start. So systemd will restart the snap, thus going through this process again.

The benefit of that is whichever error happened the first time might not happen the second time.

Alternatively, if there is an unhealthy cluster member, this means that the cluster will get stuck in a restart loop because we will repeatedly fail to send a notification to the unhealthy cluster member.

masnax: If we just log the error, that means there is a chance that we won't fully send the notification to a cluster member that it can stop waiting. That means this other cluster member has to be manually restarted.

roosterfish: Ok thanks. I was just checking the code behind the `PATCH /cluster/internal/database` endopint that is called by `sendUpgradeNotification` on each member of the cluster.

So a cluster member that won't receive this notification thinks the other member's schemas are out of date so it waits for them to update their schema. I understand restarting this member at a later point in time will resolve the issue so I am fine with this change.

What I am not sure about now (and maybe this is another topic) but what happens in case multiple cluster members require a schema update and the first one sends a notification but the other ones are still updating? The member that is waiting on a notification will continue right after receiving the first notification (See https://github.com/canonical/microcluster/blob/main/internal/db/db.go#L38). 
Does this produce a conflict/race as the waiting member doesn't wait for the others to send the notification?

roosterfish (approved):

internal/daemon/daemon.go
masnax: There is no such thing as other cluster members still updating. The schema is only updated by the dqlite leader, and applies to all cluster members. There are no local updates.

The flow works like this on each cluster member:
* A cluster member starts, and increments its column in the `internal_cluster_members` table with the size of the update that it expects.
* The same cluster member then checks all rows for the schema columns. 
* If they are not equal:
  * It blocks for a notification if it detects its own value is higher.
  * It exits if it detects its own value is lower than some other node.
* If the rows are all equal:
  * It applies the update (via dqlite leader) and sends the notification to all other nodes.

So the only race would be if two nodes arrive at the conclusion that all nodes are equal at the same time. But the first 2 bullets happen within 1 transaction, which means the dqlite leader has to process them sequentially. If there was actually an update to run, only the last transaction processed by the dqlite leader would return a result that tells a node to run the updates.

roosterfish: Thanks, I'll bookmark this exact comment :)

Ok, the transaction is a critical peace here. This ensures that only one member (doesn't have to be the current leader) arrives at a stage where it sees that all rows are equal. Now this member updates the schema (which will get executed by whomever is currently the leader of dqlite)?

PR #148, opened 2024-06-14
Add descriptive errors for database status checks

### LINKED ISSUE #66: Queries return `Error: Daemon not yet initialized` while waiting for cluster member upgrades
If one member of the cluster gets upgraded to a version that introduces new schema upgrade, queries made from upgraded member result in "Daemon not initialized" error which is bit misleading.
Example:
```shell
$ microovn cluster list
Error: Daemon not yet initialized
```
I'm using `microovn` as an example because I have it at hand but the error comes verbatim from calling `microcluster.internal.rest.client.Client.GetClusterMembers()`.

Looking at the logs I get more informative message
```
level=warning msg="Waiting for other cluster members to upgrade their versions" address="https://10.5.3.129:6443"
```
This state persists until all cluster members get upgraded at which point everything goes back to normal. Is there a possibility to get back more precise error message, like the one that's in the logs, when calling API via MicroCluster client? Alternatively are there any checks that I could perform before doing requests that would get me this message? I tried `MicroCluster.Ready()` and `MicroCluster.Status()` but both returned only generic "cluster is not ready" message.

masnax:
Closes #66

Rather than just returning `Daemon not yet initialized` when the database is offline, there are now more nuanced errors:

* `Database is not yet initialized` will be reported if the database is offline (prior to starting dqlite, or when the database is stopped)
* `Database is still starting` will be reported after dqlite has begun starting up, but before it has finished.
* `Database upgrade is in progress: %d cluster members have not yet received the update` will be reported for as long as a system is blocking, waiting for a schema upgrade or API extension to be committed. 

The number returned by the latter error will be fetched from the cluster members table as a count of all systems whose schema versions are smaller than the local schema version. Additionally, `sqlite_master` is checked to see if the current schema supports api extensions. If so, then we ensure that API extensions must also match what the local system expects.

### DIFF
--- a/cluster/cluster_members.go
+++ b/cluster/cluster_members.go
@@ -270,3 +270,57 @@ func GetClusterMemberAPIExtensions(ctx context.Context, tx *sql.Tx) ([]extension
 
 	return results, nil
 }
+
+// GetUpgradingClusterMembers returns the list of all cluster members during an upgrade, as well as a map of members who we consider to be in a waiting state.
+// This function can be used immediately after dqlite is ready, before we have loaded any prepared statements.
+// A cluster member will be in a waiting state if a different cluster member still exists with a smaller API extension count or schema version.
+func GetUpgradingClusterMembers(ctx context.Context, tx *sql.Tx, schemaInternal uint64, schemaExternal uint64, apiExtensions extensions.Extensions) (allMembers []InternalClusterMember, awaitingMembers map[string]bool, err error) {
+	tableName, err := prepareUpdateV1(tx)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// Check for the `api_extensions` column, which may not exist if we haven't actually run the update yet.
+	stmt := fmt.Sprintf(`
+SELECT count(name)
+FROM pragma_table_info('%s')
+WHERE name IN ('api_extensions');
+`, tableName)
+
+	var count int
+	err = tx.QueryRow(stmt).Scan(&count)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// Fetch all cluster members with a smaller schema version than we expect.
+	stmt = `SELECT id, name, address, certificate, schema_internal, schema_external, %s, heartbeat, role
+  FROM %s
+  ORDER BY name
+	`
+
+	// If API extensions are supported, ensure the list for each cluster member also matches what we expect,
+	// and only return cluster members for whom it does not.
+	apiField := "'[]' as api_extensions"
+	if count == 1 {
+		apiField = "api_extensions"
+	}
+
+	stmt = fmt.Sprintf(stmt, apiField, tableName)
+	allMembers, err = getInternalClusterMembersRaw(ctx, tx, stmt)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	awaitingMembers = make(map[string]bool, len(allMembers))
+	for _, member := range allMembers {
+		awaitingMembers[member.Name] = member.SchemaInternal < schemaInternal || member.SchemaExternal < schemaExternal
+
+		// If we have API extension support, also compare against the database API extensions.
+		if count == 1 {
+			awaitingMembers[member.Name] = member.APIExtensions.IsSameVersion(apiExtensions) != nil || awaitingMembers[member.Name]
+		}
+	}
+
+	return allMembers, awaitingMembers, nil
+}
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -247,8 +247,9 @@ func (db *DB) retry(ctx context.Context, f func(context.Context) error) error {
 
 // Update attempts to update the database with the executable at the path specified by the SCHEMA_UPDATE variable.
 func (db *DB) Update() error {
-	if !db.IsOpen() {
-		return fmt.Errorf("Failed to update, database is not yet open")
+	err := db.IsOpen(context.Background())
+	if err != nil {
+		return fmt.Errorf("Failed to update, database is not yet open: %w", err)
 	}
 
 	updateExec := os.Getenv(sys.SchemaUpdate)
@@ -263,7 +264,7 @@ func (db *DB) Update() error {
 	time.Sleep(wait)
 
 	logger.Info("Triggering cluster auto-update now")
-	_, err := shared.RunCommand(updateExec)
+	_, err = shared.RunCommand(updateExec)
 	if err != nil {
 		logger.Error("Triggering cluster update failed", logger.Ctx{"err": err})
 		return err
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -130,7 +130,7 @@ func (db *DB) waitUpgrade(bootstrap bool, ext extensions.Extensions) error {
 	newSchema := db.Schema()
 	if !bootstrap {
 		checkVersions := func(ctx context.Context, current int, tx *sql.Tx) error {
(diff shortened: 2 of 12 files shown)


masnax:
Drafting until #150 is settled.

internal/db/db.go
```diff
@@ -189,6 +207,10 @@ func (db *DB) waitUpgrade(bootstrap bool, ext extensions.Extensions) error {
 
 	// If we are not bootstrapping, wait for an upgrade notification, or wait a minute before checking again.
 	if otherNodesBehind && !bootstrap {
+		db.statusLock.Lock()
```
roosterfish, line 210: What is the reason for moving those lines to here? Wouldn't they be better placed in the `Open()` func right before the call to `db.waitUpgrade(bootstrap, ext)`? This way the status is also set for the `newSchema.Ensure(db.db)` operation inside of `waitUpgrade()`.

roosterfish: And it's easier to reason about the status in case the `reverter` gets called as all the statues are set in the same func.

internal/db/dqlite.go
```diff
@@ -349,7 +366,10 @@ func dqliteNetworkDial(ctx context.Context, addr string, db *DB) (net.Conn, erro
 
 // Stop closes the database and dqlite connection.
 func (db *DB) Stop() error {
+	db.statusLock.Lock()
 	db.cancel()
+	db.status = StatusOffline
```
roosterfish, line 427: After stopping the database I think it's confusing if you get back the error `"Database is not yet initialized"`. This indicates you could still recover from the stop. 
Do we use it like this or is a call to `Stop()` meant as a final operation as part of shutting down the daemon?
If so would it make more sense to have a `StatusShutdown` instead?

cluster/cluster_members.go
```diff
@@ -270,3 +270,56 @@ func GetClusterMemberAPIExtensions(ctx context.Context, tx *sql.Tx) ([]extension
...
+	`
+
+	// If API extensions are supported, ensure the list for each cluster member also matches what we expect,
+	// and only return cluster members for whom it does not.
+	apiField := "'[]' as api_extensions"
+	if count == 1 {
+		apiField = "api_extensions"
+	}
+
+	stmt = fmt.Sprintf(stmt, apiField, tableName)
```
roosterfish, line 308: To not open any doors for injections in the future, should `tableName` be a prepared statement instead? 

Right now that is not an issue as `prepareUpdateV1` returns only defined strings or `internal_cluster_members_new` if it's found in the DB.

cluster/cluster_members.go
```diff
@@ -270,3 +270,56 @@ func GetClusterMemberAPIExtensions(ctx context.Context, tx *sql.Tx) ([]extension
 
 	return results, nil
 }
+
+// GetUpgradingClusterMembers returns the list of all cluster members during an upgrade, as well as a map of members who we consider to be in a waiting state.
+// This function can be used immediately after dqlite is ready, before we have loaded any prepared statements.
```
roosterfish, line 275: Would it be worth adding a comment saying a member is still marked to be in a waiting state if the schemas match but not yet the API extensions (as the latter takes precedence in the returned map)?

internal/db/dqlite.go
```diff
@@ -206,13 +206,56 @@ func (db *DB) Cluster(ctx context.Context, client *dqliteClient.Client) ([]dqlit
...
+				}
+			}
+
+			return nil
+		})
+		if err != nil {
+			return api.StatusErrorf(http.StatusInternalServerError, "Failed to fetch awaiting cluster members: %w", err)
+		}
+
+		return api.StatusErrorf(http.StatusServiceUnavailable, "%s. %d cluster members have not yet received the update", status, awaitingSystems)
```
roosterfish, line 268: `:` instead of `.`?

internal/rest/resources/certificates.go
```diff
@@ -35,7 +35,7 @@ func clusterCertificatesPut(s *state.State, r *http.Request) response.Response {
 	}
 
 	// Forward the request to all other nodes if we are the first.
-	if !client.IsNotification(r) && s.Database.IsOpen() {
+	if !client.IsNotification(r) && s.Database.IsOpen(r.Context()) == nil {
```
roosterfish, line 38: Should the error be logged as sending the notification seems crucial?

roosterfish (changes requested):
Looks good overall, only a few smaller questions and suggestions as usual :)

internal/db/db.go
masnax: My thought was that `StatusUpgrading` is used to indicate when the daemon is blocked on waiting for an upgrade, and this should signify to the user that they need to do something on their other systems. 

`waitUpgrade` always runs and spends the majority of its implementation checking the dqlite database to decide if there is an update. Only if it reaches the `select` block at the bottom does it block on expecting an update.


If we move this status up to `IsOpen`, then it will be the longest sustained status (longer than `StatusStarting`) prior to being ready, and the daemon will always be in this state on every restart, which removes any significance between `StatusUpgrading` and `StatusStarting`.

internal/db/dqlite.go
masnax: Sure, I can add a `StatusShutdown` here.

cluster/cluster_members.go
masnax: Hmm, I didn't know you could do that with the table name as well. I'll give it a try.

internal/db/dqlite.go
masnax: Instead, I've made `StatusOffline` report `Database is offline` and added a `StatusNotReady` to indicate the database is not yet initialized.

cluster/cluster_members.go
masnax: Sadly, this doesn't work in sqlite. You get `Error: near "?": syntax error`

internal/db/db.go
roosterfish: Ok makes sense. From the status message it wasn't clear to me that this should indicate that the member is waiting on a schema update to complete (waiting on the notification). It sounds more like update process in general.

By the way, does the node on which the schema update gets executed also display the same status whilst running the update?

cluster/cluster_members.go
roosterfish: Ah yeah that's true. Don't want to over complicate things now but maybe what we have to do in the future is check against a whitelist of allowed table names?

roosterfish (dismissed):
Thanks, looks good.

internal/db/db.go
masnax: That one will display that `StatusStarting` until the updates complete, since it never reaches the `upgradeCh` block.

masnax: I think the name of that status is confusing. It's status message is "Waiting for other cluster members to update their schema versions" which is definitely not what the node that runs the updates is doing. 

I'll rename it to `StatusWaiting`.

masnax: Alright, I've updated the status name to `StatusWaiting`. Its default message is `Database is waiting for an upgrade` and its `IsOpen` error message is `Database is waiting for an upgrade: %d cluster members have not yet received the update`.

So on bootstrap, or when starting to a fully reconciled set of schema updates & API extensions, the daemon will report `Database is still starting` until the database is available. If it detects that it cannot be available without user interaction due to a schema update not present on every node, it reports the `StatusWaiting` error instead.

roosterfish: Great, this sounds feasible.

roosterfish (approved):

PR #149, opened 2024-06-14
Minimal cluster recovery

MggMuggins:
A few things:
- Updated [the spec](https://discourse.ubuntu.com/t/microcluster-microcloud-quorum-recovery/45560) to reflect what I've implemented here
- This works with microd; ~~tests in the works~~ see below.
- I've not implemented address changes as it requires an update to the database once it's up; @masnax mentioned that patches in LXD were relevant. I'll throw the code I do have pursuant to address changes in a branch soon.
- ~~Also still working on the `TODO` to reach out to the rest of the cluster~~
- Setting all the node roles except one voter to `stand-by` causes dqlite to fail; I'd like to find out why but it's not a high priority as the other configurations I've tried (including stand-by members) work. +1 on a fix for #66 .

~~This PR depends on #146~~ Merged

LXD-1118

### DIFF
--- a/cluster/recover.go
+++ b/cluster/recover.go
@@ -1,0 +1,40 @@
+package cluster
+
+import (
+	"fmt"
+
+	"github.com/canonical/go-dqlite"
+	dqliteClient "github.com/canonical/go-dqlite/client"
+)
+
+// DqliteMember is the information that can be derived locally about a cluster
+// member without access to the dqlite database.
+type DqliteMember struct {
+	// dqlite.NodeInfo fields
+	DqliteID uint64 `json:"id" yaml:"id"`
+	Address  string `json:"address" yaml:"address"`
+	Role     string `json:"role" yaml:"role"`
+
+	Name string `json:"name" yaml:"name"`
+}
+
+// NodeInfo is used for interop with go-dqlite.
+func (m DqliteMember) NodeInfo() (*dqlite.NodeInfo, error) {
+	var role dqliteClient.NodeRole
+	switch m.Role {
+	case "voter":
+		role = dqliteClient.Voter
+	case "stand-by":
+		role = dqliteClient.StandBy
+	case "spare":
+		role = dqliteClient.Spare
+	default:
+		return nil, fmt.Errorf("invalid dqlite role %q", m.Role)
+	}
+
+	return &dqlite.NodeInfo{
+		ID:      m.DqliteID,
+		Role:    role,
+		Address: m.Address,
+	}, nil
+}
--- a/internal/recover/recover.go
+++ b/internal/recover/recover.go
@@ -1,0 +1,332 @@
+package recover
+
+import (
+	"archive/tar"
+	"compress/gzip"
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"io/fs"
+	"os"
+	"path"
+	"strings"
+
+	dqlite "github.com/canonical/go-dqlite/client"
+	"github.com/canonical/lxd/shared/logger"
+	"gopkg.in/yaml.v2"
+
+	"github.com/canonical/microcluster/cluster"
+	"github.com/canonical/microcluster/internal/sys"
+	"github.com/canonical/microcluster/internal/trust"
+)
+
+// GetDqliteClusterMembers parses the trust store and
+// path.Join(filesystem.DatabaseDir, "cluster.yaml").
+func GetDqliteClusterMembers(filesystem *sys.OS) ([]cluster.DqliteMember, error) {
+	storePath := path.Join(filesystem.DatabaseDir, "cluster.yaml")
+	nodeInfo, err := dumpYamlNodeStore(storePath)
+	if err != nil {
+		return nil, err
+	}
+
+	remotes, err := ReadTrustStore(filesystem.TrustDir)
+	if err != nil {
+		return nil, err
+	}
+
+	remotesByName := remotes.RemotesByName()
+
+	var members []cluster.DqliteMember
+	for _, remote := range remotesByName {
+		for _, info := range nodeInfo {
+			if remote.Address.String() == info.Address {
+				members = append(members, cluster.DqliteMember{
+					DqliteID: info.ID,
+					Address:  info.Address,
+					Role:     info.Role.String(),
+					Name:     remote.Name,
+				})
+			}
+		}
+	}
+
+	return members, nil
+}
+
+func dumpYamlNodeStore(path string) ([]dqlite.NodeInfo, error) {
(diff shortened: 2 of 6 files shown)


MggMuggins:
I've added a test to check the trivial case(s) for cluster recovery; at least make sure that the DB comes up and the recovery export is importable on more than just the member that exported it. It would be nice to have tests for more scenarios but it feels like building out a proper integration test suite for microcluster is out of scope for my cluster recovery task.

cluster/recover.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"fmt"
+
+	"github.com/canonical/go-dqlite"
+	dqliteClient "github.com/canonical/go-dqlite/client"
+)
+
+// LocalMember is the information that can be derived locally about a cluster member
+// without access to the dqlite database.
+type LocalMember struct {
```
masnax, line 12: Should this be in a public package? I think this might fit better in the `internal/recover` or `internal/db` packages, no?

internal/recover/recover.go
```diff
@@ -0,0 +1,339 @@
...
+	"gopkg.in/yaml.v2"
+
+	"github.com/canonical/microcluster/cluster"
+	"github.com/canonical/microcluster/internal/sys"
+	"github.com/canonical/microcluster/internal/trust"
+)
+
+// GetLocalClusterMembers parses the trust store and
+// path.Join(filesystem.DatabaseDir, "cluster.yaml").
+func GetLocalClusterMembers(filesystem *sys.OS) ([]cluster.LocalMember, error) {
```
masnax, line 26: Can this be renamed to `GetDqliteClusterMembers` instead? `LocalClusterMembers` might clash with [types.ClusterMemberLocal](https://github.com/canonical/microcluster/blob/f79afecf8b8945cd4500d3d0546e52671cc79aac/internal/rest/types/cluster.go#L23)

internal/recover/recover.go
```diff
@@ -0,0 +1,339 @@
...
+
+	err := unpackTarball(tarballPath, unpackDir)
+	if err != nil {
+		return err
+	}
+
+	// TODO Is this a reasonable sanity check or is a parseable cluster.yaml
+	// containing the local node sufficient?
+	metadataPath := path.Join(unpackDir, "metadata1")
+	if _, err := os.Stat(metadataPath); errors.Is(err, os.ErrNotExist) {
```
masnax, line 160: Is this file always present, both when the daemon is stopped cleanly and uncleanly?

internal/recover/recover.go
```diff
@@ -0,0 +1,339 @@
...
+	if !found {
+		return fmt.Errorf("missing local cluster member in incoming cluster.yaml")
+	}
+
+	err = os.WriteFile(recoveryInfoYamlPath, localInfoYaml, 0o664)
+	if err != nil {
+		return err
+	}
+
+	// FIXME: Take a DB backup
```
masnax, line 205: I think this should definitely get implemented before this PR is merged.

microcluster/app.go
```diff
@@ -255,12 +255,45 @@ func (m *MicroCluster) RecoverFromQuorumLoss(members []cluster.LocalMember) erro
...
 	}
 
-	// Check each cluster member's /1.0 endpoint to ensure that they are unreachable
+	// Check each cluster member's /1.0 to ensure that they are unreachable.
 	// This is a sanity check to ensure that we're not reconfiguring a cluster
 	// that's still partially up.
-	// It may also be possible to check the raft term of each surviving member
-	// and redirect the user to call RecoverFromQuorumLoss on that member instead.
-	// TODO
+	remotes, err := recover.ReadTrustStore(m.FileSystem.TrustDir)
```
masnax, line 263: Could we move these below the control socket check so we don't unnecessarily send out requests to every other cluster member if we are going to fail on a local check.

microcluster/app.go
```diff
@@ -255,12 +255,45 @@ func (m *MicroCluster) RecoverFromQuorumLoss(members []cluster.LocalMember) erro
...
+	}
+
+	cluster, err := remotes.Cluster(false, serverCert, clusterKey)
+	if err != nil {
+		return err
+	}
+
+	err = cluster.Query(context.Background(), true, func(ctx context.Context, client *client.Client) error {
+		var rslt internalTypes.Server
+		err := client.Query(ctx, "GET", "1.0", api.NewURL(), nil, &rslt)
```
masnax, line 290: What is the default context timeout for this query? Should we specify a rather short one?

masnax (changes requested):
Looks good to me overall, just some naming around `LocalClusterMembers`.

Additionally, we should probably back up the existing database, probably in a directory named with a timestamp, and add a log when the daemon starts that a database backup exists.

cluster/recover.go
MggMuggins: `LocalMember` is exposed in `GetLocalClusterMembers` and `RecoverFromQuorumLoss` with the idea that consumers can modify it in ways other than just rendering Yaml (as discussed in our sync).

internal/recover/recover.go
MggMuggins: Yeah, let's do `s/GetLocalClusterMembers/GetDqliteClusterMembers` and `s/LocalMember/DqliteMember/`. Think I will stick with `--local` for the cli though unless you have objections; I think it makes more sense than using `--dqlite` in that context.

internal/recover/recover.go
MggMuggins: Unclear. I initially added this bit before realizing that checking if the local node is in the incoming `cluster.yaml` was a much more robust check. Going to drop this bit.

internal/recover/recover.go
MggMuggins: Will add the commits from #160 here.

microcluster/app.go
MggMuggins: I reduced it to 10s so the user isn't stuck waiting for half a minute if for some reason we don't get a connection refused but the server we're connected to doesn't come back with a response. It doesn't look like `GET /1.0` queries the DB, so I'm a little unsure why we'd be able to establish a connection but not get a response (unless a machine has too much memory or CPU pressure).
This bug is old so it might not be relevant anymore, but I'm also not convinced that we can rely on "Context deadline exceeded" meaning that we've successfully established a connection: https://github.com/golang/go/issues/31863
Happy to make more changes here if needed.

masnax (approved):
Looking good, thanks @MggMuggins 

I suppose next steps would be getting this into MicroCloud, and then adding support for changing dqlite addresses.

PR #155, opened 2024-06-17
Move EndpointType to `types` package.

masnax:
The hardcoded `1.0` value for resources supplied by the project using microcluster will be removed by #142, which means we need a way to supply the endpoint prefix to client functions.

This PR moves the type definition for `EndpointType` into the API types package, and moves all internal implementations into the internal API types package. It is also renamed to `EndpointPrefix` for clarity.

Then, when #142 is merged, this value will be specified in the set of `Resources` supplied to microcluster, and can be passed into the `Query` method.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -8,6 +8,7 @@ import (
 	"github.com/canonical/lxd/shared/api"
 
 	"github.com/canonical/microcluster/internal/rest/client"
+	"github.com/canonical/microcluster/rest/types"
 )
 
 // Client is a rest client for the MicroCluster daemon.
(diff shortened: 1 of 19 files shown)


masnax:
@mas-who Once this PR is merged, it should solve the issue you were facing when removing the now-extraneous `EndpointType`.

client/client.go
```diff
@@ -22,11 +23,11 @@ func IsNotification(r *http.Request) bool {
 
 // Query is a helper for initiating a request on the /1.0 endpoint. This function should be used for all client
```
mas-who, line 24: I think the comment here is now outdated

mas-who (commented):
Thanks for the change! I think we should get this merged before #142

client/client.go
masnax: Good catch, thanks :)

example/client/client.go
```diff
@@ -20,7 +20,7 @@ func ExtendedPostCmd(ctx context.Context, c *client.Client, data *types.Extended
 	defer cancel()
 
 	var outStr string
-	err := c.Query(queryCtx, "POST", api.NewURL().Path("extended"), data, &outStr)
+	err := c.Query(queryCtx, "POST", "1.0", api.NewURL().Path("extended"), data, &outStr)
```
roosterfish, line 23: Can we use the `types.ExtendedEndpoint` const here?

roosterfish (changes requested):
Overall this looks good, just one small comment.

example/client/client.go
masnax: No, since it's from an internal package. This hardcoded value should get removed by #142.

roosterfish: Wasn't it moved away from the internal package as part of https://github.com/canonical/microcluster/pull/155/commits/e7c64e38c8b4de6b9fd7cb902827e4005dc7af04?

masnax: The type was, so external projects can set it in their `Server` structs after #142.

But all internally defined constants are in the [internal package](https://github.com/canonical/microcluster/blob/e7c64e38c8b4de6b9fd7cb902827e4005dc7af04/internal/rest/types/server.go)

roosterfish: Thanks, clear now.

roosterfish (approved):

PR #157, opened 2024-06-20
Streamline extension servers with core API

masnax:
This changes the structure of the extension servers a bit to make them play more nicely with the core API.

* Removes `CoreAPI` flag, instead any `Server` without an address/certificate will be set up with the core listener. 
* Adds `PreInit` flag. This signifies whether the `Server` should start prior to initialization if `ListenPort` is defined on the daemon.

* Adds 2 functions: `addCoreServers` and `addExtensionServers`. The former captures all entries in the `Servers` list that either do not have an address & certificate, or whose fields match the core values, and offers all of their resources over the same server. The latter starts up a new server for any remaining `Server`s whose address is defined.

* Adds a helper `AddServers` for applying the extension server config to microcluster, rather than directly in `microcluster.Args`. This is because at the time we define `microcluster.Args`, we have not yet created the state directory, which means there may not be a place for the project to store the certificates defined for the `Server`. Now, the project must call `microcluster.App` first, and then `AddServer`. 

* Adds global validation of `Servers`:
  * A server's resource path must not conflict with any other defined resource path.
  * A server's address must not conflict with any other defined server.
  * A server must define an address if it defines a protocol.

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -201,7 +201,9 @@ func (d *Daemon) init(listenPort string, schemaExtensions []schema.Update, apiEx
 	}
 
 	for _, server := range d.extensionServers {
-		serverEndpoints = append(serverEndpoints, server.Resources...)
+		if server.ServeUnix {
+			serverEndpoints = append(serverEndpoints, server.Resources...)
+		}
 	}
 
 	ctlServer := d.initServer(serverEndpoints...)
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -184,8 +184,12 @@ func (d *Daemon) init(listenPort string, schemaExtensions []schema.Update, apiEx
 
 	d.db = db.NewDB(d.shutdownCtx, d.serverCert, d.ClusterCert, d.os)
 
-	// Extract user defined endpoints for core listener.
-	coreEndpoints, err := resources.GetAndValidateCoreEndpoints(d.extensionServers)
+	listenAddr := api.NewURL()
+	if listenPort != "" {
+		listenAddr = listenAddr.Host(fmt.Sprintf(":%s", listenPort))
+	}
+
+	err = resources.ValidateEndpoints(d.extensionServers, listenAddr.URL.Host)
 	if err != nil {
 		return err
 	}
@@ -195,7 +199,11 @@ func (d *Daemon) init(listenPort string, schemaExtensions []schema.Update, apiEx
 		resources.InternalEndpoints,
 		resources.PublicEndpoints,
 	}
-	serverEndpoints = append(serverEndpoints, coreEndpoints...)
+
+	for _, server := range d.extensionServers {
+		serverEndpoints = append(serverEndpoints, server.Resources...)
+	}
+
 	ctlServer := d.initServer(serverEndpoints...)
 	ctl := endpoints.NewSocket(d.shutdownCtx, ctlServer, d.os.ControlSocket(), d.os.SocketGroup)
 	d.endpoints = endpoints.NewEndpoints(d.shutdownCtx, ctl)
@@ -206,14 +214,16 @@ func (d *Daemon) init(listenPort string, schemaExtensions []schema.Update, apiEx
 
 	if listenPort != "" {
 		serverEndpoints = []rest.Resources{resources.PublicEndpoints}
-		serverEndpoints = append(serverEndpoints, coreEndpoints...)
-		server := d.initServer(serverEndpoints...)
-		url := api.NewURL().Host(fmt.Sprintf(":%s", listenPort))
-		network := endpoints.NewNetwork(d.shutdownCtx, endpoints.EndpointNetwork, server, *url, d.serverCert)
-		err = d.endpoints.Add(network)
+		err = d.addCoreServers(true, *listenAddr, d.ServerCert(), serverEndpoints)
 		if err != nil {
 			return err
 		}
+	}
+
+	// Add extension servers before post-join hook.
+	err = d.addExtensionServers(true, d.ServerCert(), listenAddr.URL.Host)
+	if err != nil {
+		return err
 	}
 
 	d.db.SetSchema(schemaExtensions, d.Extensions)
@@ -422,22 +432,25 @@ func (d *Daemon) StartAPI(bootstrap bool, initConfig map[string]string, newConfi
 		return err
 	}
 
-	// Extract user defined endpoints for core listener.
-	coreEndpoints, err := resources.GetAndValidateCoreEndpoints(d.extensionServers)
+	// Validate the extension servers again now that we have applied addresses.
+	err = resources.ValidateEndpoints(d.extensionServers, d.address.URL.Host)
 	if err != nil {
 		return err
 	}
 
-	serverEndpoints := []rest.Resources{resources.InternalEndpoints, resources.PublicEndpoints}
-	serverEndpoints = append(serverEndpoints, coreEndpoints...)
-	server := d.initServer(serverEndpoints...)
-	network := endpoints.NewNetwork(d.shutdownCtx, endpoints.EndpointNetwork, server, d.address, d.ClusterCert())
 	err = d.endpoints.Down(endpoints.EndpointNetwork)
 	if err != nil {
 		return err
 	}
 
-	err = d.endpoints.Add(network)
+	serverEndpoints := []rest.Resources{resources.InternalEndpoints, resources.PublicEndpoints}
+	err = d.addCoreServers(false, d.address, d.ClusterCert(), serverEndpoints)
+	if err != nil {
+		return err
+	}
+
+	// Add extension servers before post-join hook.
+	err = d.addExtensionServers(false, d.ClusterCert(), d.address.URL.Host)
 	if err != nil {
 		return err
 	}
@@ -460,12 +473,6 @@ func (d *Daemon) StartAPI(bootstrap bool, initConfig map[string]string, newConfi
 		}
 
 		err = d.trustStore.Refresh()
-		if err != nil {
-			return err
-		}
-
-		// Add extension servers before post-bootstrap hook.
-		err = d.addExtensionServers()
 		if err != nil {
 			return err
 		}
@@ -587,12 +594,6 @@ func (d *Daemon) StartAPI(bootstrap bool, initConfig map[string]string, newConfi
 		return err
 	}
 
-	// Add extension servers before post-join hook.
-	err = d.addExtensionServers()
-	if err != nil {
-		return err
-	}
-
 	if len(joinAddresses) > 0 {
 		return d.hooks.PostJoin(d.State(), initConfig)
 	}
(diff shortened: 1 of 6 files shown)


masnax:
@mas-who does all of this sound reasonable to you?

mas-who:
> @mas-who does all of this sound reasonable to you?

Yes it all sounds pretty good, my main concern is the fact that server address/cert will most likely be set during initialisation. But I see you've dealt with that by using the `PreInit` flag in combination with the presence of server address/cert.

I will do a bit of testing with site manager on this branch tomorrow and provide more detailed feedback if any.

masnax:
> my main concern is the fact that server address/cert will most likely be set during initialisation. 

A general solution for this should come next. @roosterfish and I have discussed a solution to specifying the address when initializing. 

The gist of it will be that each `Server` will need to be defined with a unique `Name` when first created.

Then the [NewCluster](https://github.com/canonical/microcluster/blob/526a3b61826cf0f3d86a96af7a19fe85347c99a3/microcluster/app.go#L179) and [JoinCluster](https://github.com/canonical/microcluster/blob/526a3b61826cf0f3d86a96af7a19fe85347c99a3/microcluster/app.go#L194) functions will have a new `map[string]types.AddrPort` argument. The addresses passed in here will be assigned to the corresponding `Server` with the name keyed in the map.

internal/daemon/daemon.go
```diff
@@ -206,16 +215,18 @@ func (d *Daemon) init(listenPort string, schemaExtensions []schema.Update, apiEx
...
-		network := endpoints.NewNetwork(d.shutdownCtx, endpoints.EndpointNetwork, server, *url, d.serverCert)
-		err = d.endpoints.Add(network)
+		err = d.addCoreServers(true, *listenAddr, d.ServerCert(), serverEndpoints)
 		if err != nil {
 			return err
 		}
 	}
 
+	// Add extension servers before post-join hook.
+	err = d.addExtensionServers(true, d.ServerCert(), listenAddr.URL.Host)
```
mas-who, line 225: Do we need this here? Is the assumption here that we may have project defined network listeners that would be fully configured prior to initialisation?

internal/daemon/daemon.go
```diff
@@ -422,22 +433,25 @@ func (d *Daemon) StartAPI(bootstrap bool, initConfig map[string]string, newConfi
...
 
-	err = d.endpoints.Add(network)
+	serverEndpoints := []rest.Resources{resources.InternalEndpoints, resources.PublicEndpoints}
+	err = d.addCoreServers(false, d.address, d.ClusterCert(), serverEndpoints)
+	if err != nil {
+		return err
+	}
+
+	// Add extension servers before post-join hook.
+	err = d.addExtensionServers(false, d.ClusterCert(), d.address.URL.Host)
```
mas-who, line 454: nice, this makes a lot of sense since additional network listeners should be added regardless if we are bootstrapping or joining the cluster!

internal/rest/resources/resources.go
```diff
@@ -44,68 +45,53 @@ var InternalEndpoints = rest.Resources{
...
-		seen := make(map[string]bool)
-		for _, endpoints := range extensionServer.Resources {
-			if seen[string(endpoints.PathPrefix)] {
-				return nil, fmt.Errorf("Path prefix %q is duplicated in server configuration", endpoints.PathPrefix)
-			}
+		// Ensure no endpoint path conflicts with an internal or external endpoint.
+		for _, resource := range server.Resources {
+			for _, endpoint := range resource.Endpoints {
+				url := filepath.Join(string(resource.PathPrefix), endpoint.Path)
+				if existingEndpointPaths[url] {
```
mas-who, line 87: I think using `existingEndpointPaths` here would prevent different servers to have the same endpoints. A separate map should be created before we iterate resources of a single server to keep track of duplicated endpoints local to that server. We will still need to check against `existingEndpointPaths` for internal endpoint conflicts.

mas-who (commented):
Thanks max, this is a nice change! Left some comments for your reference.

mas-who:
> > my main concern is the fact that server address/cert will most likely be set during initialisation.
> 
> A general solution for this should come next. @roosterfish and I have discussed a solution to specifying the address when initializing.
> 
> The gist of it will be that each `Server` will need to be defined with a unique `Name` when first created.
> 
> Then the [NewCluster](https://github.com/canonical/microcluster/blob/526a3b61826cf0f3d86a96af7a19fe85347c99a3/microcluster/app.go#L179) and [JoinCluster](https://github.com/canonical/microcluster/blob/526a3b61826cf0f3d86a96af7a19fe85347c99a3/microcluster/app.go#L194) functions will have a new `map[string]types.AddrPort` argument. The addresses passed in here will be assigned to the corresponding `Server` with the name keyed in the map.

Cool, we would still need a way to adjust the `extensionServers` that's been set on the Daemon right? Also I assume eventually we would be able to persist the server configs for future updates? (I think this was suggested by @roosterfish to have a core table for this)

internal/daemon/daemon.go
```diff
@@ -195,7 +200,11 @@ func (d *Daemon) init(listenPort string, schemaExtensions []schema.Update, apiEx
 		resources.InternalEndpoints,
 		resources.PublicEndpoints,
 	}
-	serverEndpoints = append(serverEndpoints, coreEndpoints...)
+
+	for _, server := range d.extensionServers {
```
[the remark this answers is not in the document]
masnax: Alright then, perhaps we need a flag to enable/disable unix socket support.

masnax:
> Cool, we would still need a way to adjust the `extensionServers` that's been set on the Daemon right? Also I assume eventually we would be able to persist the server configs for future updates? (I think this was suggested by @roosterfish to have a core table for this)

The only fields that really need to be adjusted are the address and certificate. 

The address can be assigned when the daemon starts, and stored in the `daemon.yaml` file. 

For the certificates, they are currently first assigned (and generated) when the daemon starts. We will have to extend the internal `certificates` endpoint which allows replacing the cluster certificate, to also support replacing the certificate for any of the listeners.

internal/rest/resources/resources.go
mas-who: In site manager, we have a `GET /1.0/sites` endpoint on the core listener which allows the UI to get a list of sites for display. The creation of sites happens on the site control listener with the endpoint `POST /1.0/sites` (a request from LXD would reach this endpoint). The url path would be the same across these different network listeners.

masnax: Alright, so we can allow duplicate paths but we have to verify once we assign addresses that no `Server`s with the same address have duplicate paths. 

Because otherwise you'll get into a situation with 2 servers (`ServerA` and `ServerB`) that were supplied the same address and thus will be merged into one listener, but have different implementations of `/1.0/sites`.

internal/daemon/daemon.go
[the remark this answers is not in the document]
masnax: I've added a `ServeUnix` flag now. If this is set to true, the resources of the `Server` will be served over the unix socket as well.

If `ServeUnix` is true, then we must also validate that all resource paths are unique across all `Server`s with `ServeUnix=true` or `Address=""`, since they will be aggregated on the unix socket.

internal/rest/resources/resources.go
masnax: The validation now works like this:
* We enforce every `Server` must have a unique `Address`, or have no address.
  * If a server has a unique address, resource paths must be unique within that server.
  * If a server has no address (or if `ServeUnix=true`), resource paths must be unique across all servers with the same configuration, or with the internal paths.

mas-who:
> > Cool, we would still need a way to adjust the `extensionServers` that's been set on the Daemon right? Also I assume eventually we would be able to persist the server configs for future updates? (I think this was suggested by @roosterfish to have a core table for this)
> 
> The only fields that really need to be adjusted are the address and certificate.
> 
> The address can be assigned when the daemon starts or during init, and stored in the `daemon.yaml` file.
> 
> For the certificates, they are currently first assigned (and generated) when the daemon starts. We will have to extend the internal `certificates` endpoint which allows replacing the cluster certificate, to also support replacing the certificate for any of the listeners.

I see, sounds good! Would it be easy enough to set up endpoints for reading and updating additional network listener addresses stored in `daemon.yaml` across the cluster? I guess the follow up implementation would also allow an upstream project to restart the network listeners if those configs changed.

mas-who:
@masnax thanks for the changes :slightly_smiling_face: LGTM and also tested in site manager

masnax:
> I see, sounds good! Would it be easy enough to set up endpoints for reading and updating additional network listener addresses stored in `daemon.yaml` across the cluster? I guess the follow up implementation would also allow an upstream project to restart the network listeners if those configs changed.

That sounds odd to me, what is the use case for changing the listen address for a particular listener on node A remotely from node B while node A is still running? 

I would think this should be done while node A is offline, and likely directly from node A. It's risky to reload a listener with a new address because we can't know if it's in the middle of processing a request.

masnax:
@roosterfish Could you please review this one next? Thanks :)

mas-who:
> > I see, sounds good! Would it be easy enough to set up endpoints for reading and updating additional network listener addresses stored in `daemon.yaml` across the cluster? I guess the follow up implementation would also allow an upstream project to restart the network listeners if those configs changed.
> 
> That sounds odd to me, what is the use case for changing the listen address for a particular listener on node A remotely from node B while node A is still running?
> 
> I would think this should be done while node A is offline, and likely directly from node A. It's risky to reload a listener with a new address because we can't know if it's in the middle of processing a request.

Yes I agree. The ability to update listener addresses post initialisation was mentioned in a previous sync meeting, although I can't seem to find in my notes that we discussed the reason why we would want to update listener addresses dynamically post initialisation. Let's say if we don't support on-the-fly update of listener addresses, would the user just have to re-join the node with new addresses? Also I think in lxd, we can set the [core.https_address](https://documentation.ubuntu.com/lxd/en/latest/server/#server-core:core.https_address) config which would update the address for the API and I think this can happen whilst the server is running (I have accidentally locked myself out before by changing this address), not sure if this setting works similar to what we might want to do in microclsuter, thoughts @roosterfish ? 

In any case, this topic isn't directly related to this PR so we can always discuss it again in he SM sync :)

internal/daemon/daemon.go
```diff
@@ -195,7 +199,11 @@ func (d *Daemon) init(listenPort string, schemaExtensions []schema.Update, apiEx
...
+		serverEndpoints = append(serverEndpoints, server.Resources...)
+	}
+
 	ctlServer := d.initServer(serverEndpoints...)
 	ctl := endpoints.NewSocket(d.shutdownCtx, ctlServer, d.os.ControlSocket(), d.os.SocketGroup)
```
roosterfish, line 210: What do you think about having something similar like `addCoreServers` also for the Unix socket server? Or should it be part of `addCoreServers`?

roosterfish (changes requested):
Thanks, only a few remarks.

internal/daemon/daemon.go
```diff
@@ -587,30 +594,70 @@ func (d *Daemon) StartAPI(bootstrap bool, initConfig map[string]string, newConfi
...
+			continue
+		}
+
+		// If the server is not available prior to initialization, then skip it if we are before initialization.
+		if !s.PreInit && preInit {
+			continue
+		}
+
+		// If the server has a defined address that is different from the default URL, we can't bring it up with the core listener.
+		if s.Address.String() != defaultURL.URL.Host && s.Address != (types.AddrPort{}) {
```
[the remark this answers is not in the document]
masnax: It may be empty during init if `args.ListenPort` is unset. I'll add a comment explaining this.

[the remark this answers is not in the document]
masnax: This happens during daemon start, prior to initialization.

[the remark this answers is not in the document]
masnax: Ah, no, excuse me I read this loop wrong. 

This is checking for the following case:
* `defaultURL` is set to some value.
* `server.Address` is unset.

So we check that `server.Address` must be equal to `defaultURL` or be empty.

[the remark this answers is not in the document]
roosterfish: Thanks, clear now.

masnax:
@roosterfish AS discussed, I've added back `CoreAPI`.

So now the Servers work like this.

* If `CoreAPI` is true:
  * `Address` must be empty. The core API listener will import all resources from this Server.
  * `Certificate` must be empty
  * `Protocol` must be empty
  * `ServeUnix` can be set to serve the resources over the unix socket.
  * `PreInit` can be set to serve the resources over the pre-init `listenPort` if defined in `microcluster.Args`.

* If `CoreAPI` is false:
  * If `Address` is defined, a discrete listener will be created. Otherwise the server will be ignored.
  * If `Certificate` is empty, the cluster certificate will be used.
  * If `Protocol` is empty, `https` will be used.
  * `ServeUnix` must be empty
  * `PreInit` can be set.

internal/rest/resources/resources.go
```diff
@@ -44,68 +45,97 @@ var InternalEndpoints = rest.Resources{
...
+			serverAddresses[server.Address.String()] = true
+		} else if server.Protocol != "" {
+			return fmt.Errorf("Server protocol defined without address")
 		}
 
-		err := extensionServer.ValidateServerConfigs()
+		// Ensure no endpoint path conflicts with another endpoint on the same server.
+		// If a server lacks an address, we need to compare it to every other server
+		// that also lacks an address, as well as the internal endpoints.
+		serverPaths, err := resourcesConflict(server, existingEndpointPaths)
```
mas-who, line 99: @masnax just a thought here, if we restrict the number of servers with `CoreAPI=true` to 1, then we wouldn't need to do this validation? From a usage perspective I can't really see a reason why multiple server structs would be defined with `CoreAPI=true`. Although I think maybe we also shouldn't assume what the user might want to do, so feel free to disregard this comment :)

internal/rest/resources/resources.go
```diff
@@ -44,68 +45,97 @@ var InternalEndpoints = rest.Resources{
...
-			if err != nil {
-				return nil, err
-			}
+	return nil
+}
 
-			coreEndpoints = append(coreEndpoints, endpoints)
-			seen[string(endpoints.PathPrefix)] = true
+// resourcesConflict returns an error if the endpoint paths of the given server conflict with any paths in the given map of existing paths.
+func resourcesConflict(server rest.Server, existingPaths map[string]bool) (map[string]bool, error) {
```
mas-who, line 116: @masnax just a thought here, if we restrict the number of servers with `CoreAPI=true` to 1, then we wouldn't need to do cross server validation here? From a usage perspective I can't really see a reason why multiple server structs would be defined with `CoreAPI=true`. Although I think maybe we also shouldn't assume what the user might want to do, so feel free to disregard this comment :)

masnax: I'm thinking with `PreInit` and `ServeUnix` there may be some cases of multiple core API servers with different combinations of those 2 flags.

PR #161, opened 2024-06-26
internal/daemon: Attempt to re-use control.socket if present

masnax:
If the daemon crashed or stopped uncleanly due to an error, then the control socket file might still be present in the state directory. This means the daemon will enter a restart loop until the control socket is manually removed.

Instead, we can just attempt to re-connect to the control socket. If it's already in use then that will error out the daemon anyway.

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -81,14 +81,26 @@ func NewDaemon(project string) *Daemon {
 	}
 
 	d.stop = sync.OnceValue(func() error {
-		d.shutdownCancel()
-
-		err := d.db.Stop()
-		if err != nil {
-			return fmt.Errorf("Failed shutting down database: %w", err)
+		if d.shutdownCancel != nil {
+			d.shutdownCancel()
 		}
 
-		return d.endpoints.Down()
+		var dqliteErr error
+		if d.db != nil {
+			dqliteErr = d.db.Stop()
+			if dqliteErr != nil {
+				logger.Error("Failed shutting down database", logger.Ctx{"error": dqliteErr})
+			}
+		}
+
+		if d.endpoints != nil {
+			err := d.endpoints.Down()
+			if err != nil {
+				return err
+			}
+		}
+
+		return dqliteErr
 	})
 
 	return d
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -118,15 +118,6 @@ func (d *Daemon) Run(ctx context.Context, listenPort string, stateDir string, so
 		return fmt.Errorf("Failed to initialize directory structure: %w", err)
 	}
 
-	isAlreadyRunning, err := d.os.IsControlSocketPresent()
-	if err != nil {
-		return err
-	}
-
-	if isAlreadyRunning {
-		return fmt.Errorf("Control socket already present (%q); is another daemon already running?", d.os.ControlSocketPath())
-	}
-
 	d.extensionServers = extensionServers
 
 	err = d.init(listenPort, extensionsSchema, apiExtensions, hooks)
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -18,6 +18,7 @@ import (
 	"github.com/canonical/lxd/shared"
 	"github.com/canonical/lxd/shared/api"
 	"github.com/canonical/lxd/shared/logger"
+	"github.com/canonical/lxd/shared/revert"
 	"github.com/gorilla/mux"
 	"gopkg.in/yaml.v2"
 
@@ -130,6 +131,16 @@ func (d *Daemon) Run(ctx context.Context, listenPort string, stateDir string, so
 		return fmt.Errorf("Failed to initialize directory structure: %w", err)
 	}
 
+	// Clean up the daemon state on an error during init.
+	reverter := revert.New()
+	defer reverter.Fail()
+	reverter.Add(func() {
+		err := d.stop()
+		if err != nil {
+			logger.Error("Failed to cleanly stop the daemon", logger.Ctx{"error": err})
+		}
+	})
+
 	d.extensionServers = extensionServers
 
 	err = d.init(listenPort, extensionsSchema, apiExtensions, hooks)
@@ -143,6 +154,8 @@ func (d *Daemon) Run(ctx context.Context, listenPort string, stateDir string, so
 	}
 
 	close(d.ReadyChan)
+
+	reverter.Success()
 
 	for {
 		select {
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -6,12 +6,14 @@ import (
 	"errors"
 	"fmt"
 	"math/rand"
+	"net/http"
 	"os"
 	"time"
 
 	"github.com/canonical/lxd/lxd/db/query"
 	"github.com/canonical/lxd/lxd/db/schema"
 	"github.com/canonical/lxd/shared"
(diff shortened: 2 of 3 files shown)


internal/daemon/daemon.go
```diff
@@ -118,15 +118,6 @@ func (d *Daemon) Run(ctx context.Context, listenPort string, stateDir string, so
 		return fmt.Errorf("Failed to initialize directory structure: %w", err)
 	}
 
-	isAlreadyRunning, err := d.os.IsControlSocketPresent()
```
roosterfish, line 121: I understand we don't require this extra check as later on during `init()` we run [`removeStale()`](https://github.com/canonical/microcluster/blob/main/internal/endpoints/socket.go#L54) which will get rid of the socket if it already exists?

Please also remove `IsControlSocketPresent()` as it now looks to be unused.

roosterfish (changes requested):

roosterfish:
I guess this comes from https://github.com/canonical/microovn/pull/143? 

When following the reproducer steps, I am failing to understand what is causing an error in MicroCluster so that the socket file isn't properly cleaned up in the first place?

internal/daemon/daemon.go
masnax: It's not used now but will be used by  #149

masnax:
> I guess this comes from [canonical/microovn#143](https://github.com/canonical/microovn/pull/143)?
> 
> When following the reproducer steps, I am failing to understand what is causing an error in MicroCluster so that the socket file isn't properly cleaned up in the first place?

If the upgrade doesn't happen fast enough, some nodes might hit a context deadline and restart, leaving the control socket dangling.

roosterfish:
> If the upgrade doesn't happen fast enough, some nodes might hit a context deadline and restart, leaving the control socket dangling.

Ok. So I guess this change is reasonable, but should we also make sure in case of context deadline the socket gets cleaned too?

masnax:
> > If the upgrade doesn't happen fast enough, some nodes might hit a context deadline and restart, leaving the control socket dangling.
> 
> Ok. So I guess this change is reasonable, but should we also make sure in case of context deadline the socket gets cleaned too?

I've added a reverter to clean up any potentially dangling daemon state on a start failure.

internal/daemon/daemon.go
```diff
@@ -118,14 +119,23 @@ func (d *Daemon) Run(ctx context.Context, listenPort string, stateDir string, so
...
-	if err != nil {
-		return err
-	}
+	// Clean up the daemon state on an error during init.
+	reverter := revert.New()
+	defer reverter.Fail()
+	reverter.Add(func() {
+		if d.db != nil {
```
roosterfish, line 126: Can't we run `d.State().Stop()` here instead which sends to the `shutdownDoneCh`?

roosterfish: I maybe add the socket cleanup to https://github.com/canonical/microcluster/blob/main/internal/daemon/daemon.go#L83?

mkalcok:
I'm not familiar enough with context deadline, but is any timeout in this situation necessary? Couldn't the daemon just wait indefinitely for the rest of the cluster to get upgraded?

masnax:
> I'm not familiar enough with context deadline, but is any timeout in this situation necessary? Couldn't the daemon just wait indefinitely for the rest of the cluster to get upgraded?

The context deadline is there in case of an upgrade mismatch. Think about this case, with 2 updates available, all systems are at version 0:

System1 - updates to expect version 1, waits for the other systems to get to version 1.
System2  - updates to expect version 1, waits for other systems to get to version 1.
System3 - updates to expect version 2, waits for other systems to get to version 2.

Eventually the systems will restart the from a context timeout and repeat the loop to get a more up-to-date view of the cluster. Then systems 1 and 2 will detect that system 3 is ahead of them and report this to the user.

masnax:
> I'm not familiar enough with context deadline, but is any timeout in this situation necessary? Couldn't the daemon just wait indefinitely for the rest of the cluster to get upgraded?

There is also the factor of the microcluster bug in MicroOVN's stable branch where the systems need to wait for the next heartbeat to trust connections from each other. Since the update is happening too quickly, they are not all able to coordinate the update, thus a context deadline occurs.

masnax:
Alright, so looks like there were 3 bugs that are now addressed:

* the daemon Stop function would return prematurely if it failed to stop dqlite. Now it properly attempts to close the listeners even if dqlite fails to stop. It always returns an error if one is encountered.

* the retry loop for schema updates was trying to restart dqlite while it was already running. This was causing an address bind failure which crashed the daemon. This is now gone as well, so the daemon will not restart, only the database itself will.

* If the daemon encounters a `control.socket` file, it will try to reuse it before erroring out if it's already bound.

masnax:
I've made one other addition now:
* The database actually never needs to stop, we just need to re-run the upgrade checks. The only time we try to restart the database is if there was a problem with applying prepared statements.

* As a result of the above, we can allow transactions over the API if `AllowedBeforeInit=true` as long as the database is either in a ready or waiting state.

internal/db/db.go
```diff
@@ -56,6 +60,14 @@ func (db *DB) Open(ext extensions.Extensions, bootstrap bool, project string) er
 
 	err = cluster.PrepareStmts(db.db, project, false)
 	if err != nil {
+		// If we receive an error at this point, close the database.
+		closeErr := db.db.Close()
+		if closeErr != nil {
```
roosterfish, line 65: Should we also do this as part of the reverter?

internal/db/db.go
```diff
@@ -235,6 +235,11 @@ func (db *DB) waitUpgrade(bootstrap bool, ext extensions.Extensions) error {
 
 // Transaction handles performing a transaction on the dqlite database.
 func (db *DB) Transaction(outerCtx context.Context, f func(context.Context, *sql.Tx) error) error {
+	status := db.Status()
+	if status != StatusWaiting && status != StatusReady {
```
roosterfish, line 239: When the DB is in waiting state, why is it fine to continue with the transaction?

internal/db/db.go
masnax: No, because we expect "skippable" errors from `waitUpgrade`. If that returns an error, we don't want to close the database because we want to retry the upgrade without re-opening the database. If we put it in the reverter, then the database will always close.

internal/db/db.go
masnax: A waiting state means the database is open, but is waiting for the schema update to appear on other systems, which means some functionality is restricted.

The database is technically fully functional at this time, so we should be able to query it on endpoints where `AllowedBeforeInit=true`.

This is useful for querying the current upgrade status, like in #162

Additionally, it allows for the downstream project to still start the underlying services like OVN and Ceph while awaiting a dqlite schema change.

internal/db/db.go
roosterfish: Ok but why not putting the `reverter.Add()` right before calling `PrepareStmts`?

If we ever add new funcs after `PrepareStmts` that might return errors too we won't close the DB in case they error out. Or is closing the DB specific for if `PrepareStmts` returns an error?

masnax: Oh I see what you mean, good point.

masnax: Fixed.

roosterfish (approved):
Thanks, LGTM!

masnax:
@roosterfish Thanks for the quick turn around!

PR #162, opened 2024-06-26
Add upgrade progress to cluster member status.

masnax:
Adds `MemberUpgrading` and `MemberNeedsUpgrade` to the list of possible cluster member statuses.

Consequentially, this means we need to set `AllowedBeforeInit=true` to the `/cluster` API. In the POST case, we still enforce that the database must be fully ready. 

But for listing cluster members, as long as the cluster is at least in a waiting state, we can try to grab the current list of members. In such a case, we will compare those cluster members' schema versions to our own local version, and set `UPGRADING` as the status if they match our's, or `NEEDS UPGRADE` if they do not.

### DIFF
--- a/example/cmd/microctl/cluster_members.go
+++ b/example/cmd/microctl/cluster_members.go
@@ -3,6 +3,7 @@ package main
 import (
 	"sort"
 
+	"github.com/canonical/lxd/shared"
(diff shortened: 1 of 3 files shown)


roosterfish (approved):
LGTM!

PR #163, opened 2024-06-28
Client: remove "/" from path prefix if `endpointType` has empty value

mas-who:
Recently we merged #142 which allows the user to define their own path prefix for a client query. However, if the prefix is `""`, then we may end up with a path like `//some-path` in [QueryStruct](https://github.com/canonical/microcluster/blob/db6075a4cf00db01687e26e61a08fe679a6f567b/internal/rest/client/client.go#L275). This PR should resolve the issue.

### DIFF
--- a/internal/rest/client/client.go
+++ b/internal/rest/client/client.go
@@ -283,8 +283,8 @@ func (c *Client) QueryStruct(ctx context.Context, method string, endpointType ty
 
 	localURL.URL.Host = c.url.URL.Host
 	localURL.URL.Scheme = c.url.URL.Scheme
-	localURL.URL.Path = "/" + string(endpointType) + localURL.URL.Path
-	localURL.URL.RawPath = "/" + string(endpointType) + localURL.URL.RawPath
+	localURL.URL.Path = filepath.Join("/", string(endpointType), localURL.URL.Path)
+	localURL.URL.RawPath = filepath.Join("/", string(endpointType), localURL.URL.RawPath)
 
 	localQuery := localURL.URL.Query()
 	clientQuery := c.url.URL.Query()


internal/rest/client/client.go
```diff
@@ -281,10 +281,15 @@ func (c *Client) QueryStruct(ctx context.Context, method string, endpointType ty
...
+	pathPrefix := "/" + string(endpointType)
+	if endpointType == "" {
+		pathPrefix = ""
+	}
+
 	localURL.URL.Host = c.url.URL.Host
 	localURL.URL.Scheme = c.url.URL.Scheme
-	localURL.URL.Path = "/" + string(endpointType) + localURL.URL.Path
-	localURL.URL.RawPath = "/" + string(endpointType) + localURL.URL.RawPath
+	localURL.URL.Path = pathPrefix + localURL.URL.Path
```
roosterfish, line 291: We can just do `filepath.Join("/", string(endpointType), localURL.URL.Path)` as it will skip path elements which are empty.

roosterfish (changes requested):
Thanks for catching this, pls see my comment below.

internal/rest/client/client.go
roosterfish: No need for the extra `if`.

mas-who: Neat! Thanks will update :slightly_smiling_face:

roosterfish:
Looks good now, can you please also add the package name as prefix to the commit message?

mas-who:
> Looks good now, can you please also add the package name as prefix to the commit message?

done :)

roosterfish (approved):

PR #164, opened 2024-07-01
allow modification of response if access handler returns a callback

mas-who:
## Done
 - Allow `AccessHandler` to return a callback that can modify the `http.ResponseWriter` object.

## Context
 - In an upstream project, endpoints authentication applied by using the custom `AccessHandler` may need to write to the response regardless of success or failure. Currently in microcluster if an `AccessHandler` returns `EmptySyncResponse`, then we will not be able to access the response object even if we need to write to it.
 - The upstream project can get around this issue by writing auth data to the request context and handle it in the actual endpoint handler. However, this approach tend to require the authentication logic to be split up which makes the code difficult to understand.

### DIFF
--- a/rest/rest.go
+++ b/rest/rest.go
@@ -19,7 +19,7 @@ type EndpointAlias struct {
 // EndpointAction represents an action on an API endpoint.
 type EndpointAction struct {
 	Handler        func(state *state.State, r *http.Request) response.Response
-	AccessHandler  func(state *state.State, r *http.Request) response.Response
+	AccessHandler  func(state *state.State, r *http.Request) (trusted bool, resp response.Response)
 	AllowUntrusted bool
 	ProxyTarget    bool // Allow forwarding of the request to a target if ?target=name is specified.
 }
--- a/internal/rest/rest.go
+++ b/internal/rest/rest.go
@@ -30,17 +30,32 @@ func handleAPIRequest(action rest.EndpointAction, state *state.State, w http.Res
 
 	// If allow untrusted is not set, the request must be authenticated via core authentication (e.g. certificate in truststore).
 	if !action.AllowUntrusted {
-		resp := access.AllowAuthenticated(state, r)
-		if resp != response.EmptySyncResponse {
+		trusted, resp := access.AllowAuthenticated(state, r)
+		if !trusted {
+			if resp == nil {
+				return response.Forbidden(nil)
+			}
+
 			return resp
 		}
 	}
 
 	// Run the custom access handler if set.
 	if action.AccessHandler != nil {
-		resp := action.AccessHandler(state, r)
-		if resp != response.EmptySyncResponse {
+		trusted, resp := action.AccessHandler(state, r)
+		if !trusted {
+			if resp == nil {
+				return response.Forbidden(nil)
+			}
+
 			return resp
+		}
+
+		if resp != nil {
+			err := resp.Render(w)
+			if err != nil {
+				return response.InternalError(err)
+			}
 		}
 	}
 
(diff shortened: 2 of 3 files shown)


mas-who:
@roosterfish as discussed I experimented a bunch with this approach in site manager and it works. Although looking at it, I think it's not too different to passing the `w http.ResponseWriter` object to the `AccessHandler`?

masnax:
Is it not possible to introduce a type that implements the [Response interface](https://github.com/canonical/lxd/blob/c14800807e1159748de23a1109a9613d3d2cf44b/lxd/response/response.go#L30) and use the implementation of `Render` to write to the response?

mas-who:
> Is it not possible to introduce a type that implements the [Response interface](https://github.com/canonical/lxd/blob/c14800807e1159748de23a1109a9613d3d2cf44b/lxd/response/response.go#L30) and use the implementation of `Render` to write to the response?

Unfortunately it doesn't quite do the trick, in the case where `AccessHandler` needs to write to the response for a successful auth flow, currently we won't be able to access the `http.ResponseWriter` object.

markylaing:
> > Is it not possible to introduce a type that implements the [Response interface](https://github.com/canonical/lxd/blob/c14800807e1159748de23a1109a9613d3d2cf44b/lxd/response/response.go#L30) and use the implementation of `Render` to write to the response?
> 
> Unfortunately it doesn't quite do the trick, in the case where `AccessHandler` needs to write to the response for a successful auth flow, currently we won't be able to access the `http.ResponseWriter` object.

@masnax OIDC authentication requires writing headers on an unsuccessful login from the LXD CLI. For the browser, it needs to write cookies:
1. On success, when a new user logs in.
2. On success, when a user refreshes their identity token.
3. On failure, to delete invalid tokens.

In microcluster's [`handleAPIRequest`](https://github.com/canonical/microcluster/blob/06f9aaea410a93e75f6e638c3b65efdbabc5a205/internal/rest/rest.go#L26-L52), `Render` is only returned if it is not an empty sync response. In that case the main handler is not called. It is not possible to implement `response.Response` in such a way as to circumvent this.

rest/rest.go
```diff
@@ -19,7 +19,7 @@ type EndpointAlias struct {
 // EndpointAction represents an action on an API endpoint.
 type EndpointAction struct {
 	Handler        func(state *state.State, r *http.Request) response.Response
-	AccessHandler  func(state *state.State, r *http.Request) response.Response
+	AccessHandler  func(state *state.State, r *http.Request) (response.Response, func(w http.ResponseWriter) error)
```
markylaing, line 22: I'm not certain about this callback approach. I think it might be simpler to make this a
```
AccessHandler func(state *state.State, r *http.Request) (bool, response.Response)
```

The request can be allowed to continue if the returned value is true. `response.Response` is always rendered if not nil.

mas-who: the callback is more for writing to the `http.ResponseWriter`, currently if the request moves forward, the `response.Response` returned by `Handler` would be rendered. Which means we can't write to the response from the `AccessHandler`

markylaing: Yep I understand that, but the `response.Response` is already essentially a callback. I'm proposing that we no longer use it to decide whether or not the request should continue. The access handler can return:
- `true, nil` allow the request to continue, don't write anything to the response writer.
- `true, response.ManualResponse(...)`, allow the request to continue, call `Render` on the response.
- `false, nil`, stop handling the request, return a 403 with a generic forbidden message.
- `false, response.ManualResponse(...)`, stop handling the request, call `Render` on the response.

mas-who: Got you, that feels way cleaner :+1:

markylaing: Yeah that seems fair I think

mas-who:
@masnax I've pushed the requested update, please have a look when you get some time :slightly_smiling_face:

masnax (approved):
LGTM

PR #165, opened 2024-07-02
Project version

MggMuggins:
Take a (required) project version via the microcluster app args and include it in the response from `GET /cluster/1.0`. This will allow MicroCloud to enforce version requirements on its sub-snaps.

This also adds a build-time constant version (`git describe`) to the example.

LXD-1279

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -43,6 +43,7 @@ import (
 // Daemon holds information for the microcluster daemon.
 type Daemon struct {
 	project string // The project refers to the name of the go-project that is calling MicroCluster.
+	version string // The version of the go-project that is calling MicroCluster
 
 	config *internalConfig.DaemonConfig // Local daemon's configuration from daemon.yaml file.
 
@@ -75,12 +76,13 @@ type Daemon struct {
 }
 
 // NewDaemon initializes the Daemon context and channels.
-func NewDaemon(project string) *Daemon {
+func NewDaemon(project string, version string) *Daemon {
 	d := &Daemon{
 		shutdownDoneCh:   make(chan error),
 		ReadyChan:        make(chan struct{}),
 		extensionServers: make(map[string]rest.Server),
 		project:          project,
+		version:          version,
 	}
 
 	d.stop = sync.OnceValue(func() error {
@@ -955,14 +957,20 @@ func (d *Daemon) ServerCert() *shared.CertInfo {
 	return d.serverCert
 }
 
-// Address ensures both the daemon and state have the same address.
+// Address is the listen address for the daemon.
 func (d *Daemon) Address() *api.URL {
 	return api.NewURL().Scheme("https").Host(d.config.GetAddress().String())
 }
 
-// Name ensures both the daemon and state have the same name.
+// Name is this daemon's cluster member name.
 func (d *Daemon) Name() string {
 	return d.config.GetName()
+}
+
+// Version is provided by the MicroCluster consumer. The daemon includes it in
+// its /cluster/1.0 response.
+func (d *Daemon) Version() string {
+	return d.version
 }
 
 // LocalConfig returns the daemon's internal config implementation.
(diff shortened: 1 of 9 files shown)


internal/state/state.go
```diff
@@ -36,6 +36,9 @@ type State struct {
 	// Name of the cluster member.
 	Name func() string
 
+	// Version provided by the consumer of MicroCluster
```
roosterfish, line 39: Missing a dot at the end :)

internal/daemon/daemon.go
```diff
@@ -772,6 +774,11 @@ func (d *Daemon) Name() string {
 	return d.name
 }
 
+// Version ensures both the daemon and state have the same version.
```
roosterfish, line 777: The comment feels misleading here as it's just a getter for the internal `.version` value?

microcluster/app.go
```diff
@@ -80,9 +84,13 @@ func (m *MicroCluster) Start(ctx context.Context, extensionsSchema []schema.Upda
 		return err
 	}
 
+	if m.args.Version == "" {
```
roosterfish, line 87: Can we run this check already on `microcluster.App()` where we hand over the args? Then the error message would be much clearer as you could just say the version is missing in the args which were set as part of calling `App()`.

roosterfish (dismissed):
Already looking good, just a few smaller suggestions.

microcluster/app.go
MggMuggins: I implemented it that way initially; however, we need to initialize `microcluster.App` every time a user wants to make a client request. Since we're only using the version value in the daemon, I didn't want to force clients to supply one if it wasn't going to be used.
I'll clarify the error message; let me know if you'd still prefer that it always be required.

internal/daemon/daemon.go
MggMuggins: This follows the pattern used for `Name()`; since it's just a getter it hardly needs a comment (except for the linter...). I can add a commit to "refactor" these comments if that makes more sense here.

microcluster/app.go
masnax: I think what we need is two sets of `microcluster.Args`. Right now a lot of args that are daemon-start specific are conflated with general args like `Proxy`, `Client`, and `StateDir`. 


I think it's fine for this PR but I think we should have a follow-up PR that creates a `microcluster.DaemonArgs` that contains all of the arguments that we specify to `microcluster.Start`. 

In fact, maybe this can be an internally defined type in the `daemon package`:

```go
// in the daemon package
type Args struct

// in the microcluster package
type DaemonArgs = daemon.Args
```

And that way the signature for both `daemon.Run` and `microcluster.Start` can be changed to the following:
```go
func (d *Daemon) Run(ctx context.Context, args daemon.Args)

func (m *MicroCluster) Start(ctx context.Context, daemonArgs microcluster.DaemonArgs)
```

microcluster/app.go
```diff
@@ -42,6 +42,10 @@ type Args struct {
 	StateDir    string
 	SocketGroup string
 
+	// Consumers of MicroCluster should provide a version to serve at /cluster/1.0
+	// This is required at daemon start.
+	Version string
```
masnax, line 47: As discussed today, it may be worth enforcing that this has to be a `semver` by making it a defined type.

MggMuggins: I'm a little hesitant to do this until we know that we can expect a semver version from Micro{Ceph,OVN}. Requiring semver may also be irritating without much benefit in other Micro* projects, since we're requiring that this field be populated before the daemon can start.
Also, the [golang.org implementation](https://pkg.go.dev/golang.org/x/mod/semver) doesn't expose a public type; it just uses `string`.
Let me know what you'd prefer.

microcluster/app.go
```diff
@@ -80,9 +84,13 @@ func (m *MicroCluster) Start(ctx context.Context, extensionsSchema []schema.Upda
 		return err
 	}
 
+	if m.args.Version == "" {
+		return fmt.Errorf("microcluster.Args.Version was missing at daemon start")
```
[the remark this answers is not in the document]
MggMuggins: This error is most likely to happen in downstream tests/during development; it seems unlikely that anyone would run into it in a user-facing context. I was thinking to provide as much context as possible with that scenario in mind.

How about `Version was missing at MicroCluster daemon start`?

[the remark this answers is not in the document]
roosterfish: Looks good.

masnax:
@MggMuggins needs a rebase

MggMuggins:
@masnax Ready to go

masnax (approved):
Thanks @MggMuggins

PR #166, opened 2024-07-05
internal/db: Pass dial context through to dialer.Dial

masnax:
`go-dqlite` sets a context on calls to the `DialFunc` specified by us that may have already been cancelled. In such cases we shouldn't need to continue to execute `Dial`, so this passes the context down to `dial.DialContext`.


MggMuggins (approved):
LGTM; example tests pass

PR #168, opened 2024-07-08
microcluster: Add the apiExtensions to the MicroCluster app args

gabrielmougard:
needed by: https://github.com/canonical/microcloud/pull/245/commits/e6eadc1981614ae9d1a909239243dde93a901841

___

This way, a MicroCluster app started in an particular snap application (like `microovn`), when its recovered as a MicroCluster app in `microcloud`, can expose its API extensions.

### DIFF
--- a/internal/rest/types/server.go
+++ b/internal/rest/types/server.go
@@ -1,14 +1,16 @@
 package types
 
 import (
+	"github.com/canonical/microcluster/internal/extensions"
 	"github.com/canonical/microcluster/rest/types"
 )
 
 // Server represents server status information.
 type Server struct {
-	Name    string         `json:"name"    yaml:"name"`
-	Address types.AddrPort `json:"address" yaml:"address"`
-	Ready   bool           `json:"ready"   yaml:"ready"`
+	Name       string                `json:"name"    yaml:"name"`
+	Address    types.AddrPort        `json:"address" yaml:"address"`
+	Ready      bool                  `json:"ready"   yaml:"ready"`
+	Extensions extensions.Extensions `json:"extensions" yaml:"extensions"`
 }
 
 const (
--- a/internal/extensions/extensions.go
+++ b/internal/extensions/extensions.go
@@ -7,6 +7,7 @@ import (
 	"regexp"
 
 	"github.com/canonical/lxd/shared"
+	"gopkg.in/yaml.v3"
 )
 
 var internalExtensionRegex = regexp.MustCompile(`^internal:[a-z0-9]+(_[a-z0-9]+)*$`)
@@ -189,6 +190,75 @@ func (e Extensions) IsSameVersion(t Extensions) error {
 		if (e)[i] != (t)[i] {
 			return fmt.Errorf("The source and target registries have different extensions")
 		}
+	}
+
+	return nil
+}
+
+// MarshalJSON implements json.Marshaler for the Extensions type.
+func (e Extensions) MarshalJSON() ([]byte, error) {
+	if len(e) == 1 {
+		return json.Marshal(e[0])
+	}
+
+	return json.Marshal([]string(e))
+}
+
+// MarshalYAML implements yaml.Marshaler for the Extensions type.
+func (e Extensions) MarshalYAML() (any, error) {
+	if len(e) == 1 {
+		return e[0], nil
+	}
+
+	return []string(e), nil
+}
+
+// UnmarshalJSON implements the json.Unmarshaler interface.
+func (e *Extensions) UnmarshalJSON(data []byte) error {
+	var v any
+	if err := json.Unmarshal(data, &v); err != nil {
+		return err
+	}
+
+	switch value := v.(type) {
+	case string:
+		*e = Extensions{value}
+	case []any:
+		*e = make(Extensions, len(value))
+		for i, item := range value {
+			str, ok := item.(string)
+			if !ok {
+				return fmt.Errorf("invalid type for Extension item at index %d", i)
+			}
+
+			(*e)[i] = str
+		}
+	default:
+		return fmt.Errorf("invalid type for Extension")
+	}
+
+	return nil
+}
+
+// UnmarshalYAML implements the yaml.Unmarshaler interface.
+func (e *Extensions) UnmarshalYAML(value *yaml.Node) error {
+	switch value.Kind {
+	case yaml.ScalarNode:
+		var s string
+		if err := value.Decode(&s); err != nil {
+			return err
+		}
+
+		*e = Extensions{s}
+	case yaml.SequenceNode:
+		var slice []string
+		if err := value.Decode(&slice); err != nil {
+			return err
+		}
+
+		*e = Extensions(slice)
+	default:
+		return fmt.Errorf("invalid node kind for Extension: %v", value.Kind)
 	}
 
 	return nil
--- a/internal/rest/resources/api_1.0.go
+++ b/internal/rest/resources/api_1.0.go
@@ -24,8 +24,9 @@ func api10Get(s *state.State, r *http.Request) response.Response {
 	}
 
 	return response.SyncResponse(true, internalTypes.Server{
-		Name:    s.Name(),
-		Address: addrPort,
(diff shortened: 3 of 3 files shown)


roosterfish (changes requested):
Instead of caching it a second time in the `MicroCluster` struct, can we retrieve if from the daemon's `Extensions` field by adding another `func (e *Extensions) List() []string`?
Then the `MicroCluster` struct can grow support for another func e.g. `APIExtensions()` and forward this list.

This way we store them at a single place and don't overload the microcluster app struct.

gabrielmougard:
I like this solution better :) let's do that! Thanks!

gabrielmougard:
@roosterfish updated. I had to give an internal handle to the daemon pointer as part of the MicroCluster app though. I don't know if there is a way to 'query' the daemon instance other than registering it during the `Start()` hook.

roosterfish:
> @roosterfish updated. I had to give an internal handle to the daemon pointer as part of the MicroCluster app though. I don't know if there is a way to 'query' the daemon instance other than registering it during the `Start()` hook.

roosterfish:
Sorry, wrong button, didn't want to close the PR 🤦🏼

microcluster/app.go
```diff
@@ -34,7 +34,8 @@ import (
 type MicroCluster struct {
 	FileSystem *sys.OS
 
-	args Args
+	args   Args
+	daemon *daemon.Daemon
```
roosterfish, line 38: There looks the be only one place where we access the daemon directly and that is during the call to `Start()` as this setups the local microcluster deamon.

All the other client funcs (`Ready()`, `SQL()`, `JoinCluster()` to name a few) use API calls to retrieve the information.
I am almost certain we have to go the same route with the list of API extensions.

Inside the API handler you can access the daemon via the its `State()`. See [here](https://github.com/canonical/microcluster/blob/main/internal/daemon/daemon.go#L808) how to pass something from the daemon which otherwise cannot be reached.

Maybe it would make sense to add the list to the `/cluster/1.0` endpoint like we have in LXD when querying `/1.0`. What do you think @masnax ?
You can then query this endoint in the `APIExtensions()` func and return the respective value.

gabrielmougard: Actually, there is no need for doing that... The extension list is already available when calling `GetClusterMembers` . I just forgot it was exposed there :)

gabrielmougard: @roosterfish updated :)

microcluster/app.go
```diff
@@ -480,3 +480,23 @@ func (m *MicroCluster) SQL(ctx context.Context, query string) (string, *internal
...
+	if err != nil {
+		return nil, err
+	}
+
+	if len(clusterMembers) == 0 {
+		return nil, fmt.Errorf("No cluster members found when fetching API extensions")
+	}
+
+	// Each cluster member will have the same extensions, so we only need to check one.
+	return clusterMembers[0].Extensions, nil
```
roosterfish, line 501: If the members have different API extensions I guess this will report only the extensions of a single member? Would it be worth returning a map instead that contains the list of extensions for every member?

roosterfish: This would presumably happen during an update but I am not sure if during this time you could actually query the endpoint and get back a result?

microcluster/app.go
roosterfish: Great thanks, looks much cleaner :)

microcluster/app.go
gabrielmougard: yes, let's do that it seems much more secure (though it looks like a very rare scenario). Then I'll add a validation logic on the MicroCloud side of things to ensure that all the members have the same extensions. :)

microcluster/app.go
gabrielmougard: fixed

microcluster/app.go
gabrielmougard: @roosterfish updated

microcluster/app.go
```diff
@@ -480,3 +480,27 @@ func (m *MicroCluster) SQL(ctx context.Context, query string) (string, *internal
...
+	if err != nil {
+		return nil, err
+	}
+
+	clusterMembers, err := c.GetClusterMembers(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(clusterMembers) == 0 {
```
roosterfish, line 496: This check isn't anymore required. But instead use the `len()` of cluster members when initializing the map below.

microcluster/app.go
```diff
@@ -480,3 +480,27 @@ func (m *MicroCluster) SQL(ctx context.Context, query string) (string, *internal
 
 	return "", batch, err
 }
+
+// APIExtensions returns the list of API extensions registered with the daemon.
```
roosterfish, line 484: "registered with every member of the cluster."?

microcluster/app.go
gabrielmougard: fixed

microcluster/app.go
gabrielmougard: fixed

gabrielmougard:
@roosterfish updated

masnax (changes requested):
In MicroCloud, at the time you would need to use this function, the database would not yet be up so you can't fetch the list of cluster members.

How about instead, we add the in-memory API Extensions to the `Server` API payload which is returned from `GET /cluster/1.0`?
https://github.com/canonical/microcluster/blob/main/internal/rest/types/server.go


There is already a [Status client function](https://github.com/canonical/microcluster/blob/ec0a2d6e0f31c63851ec416068f8e338cd910e09/microcluster/app.go#L108-L122) which we use in MicroCloud.

gabrielmougard:
@roosterfish @masnax updated

roosterfish (dismissed):
LGTM!

internal/rest/types/server.go
```diff
@@ -6,9 +6,10 @@ import (
 
 // Server represents server status information.
 type Server struct {
-	Name    string         `json:"name"    yaml:"name"`
-	Address types.AddrPort `json:"address" yaml:"address"`
-	Ready   bool           `json:"ready"   yaml:"ready"`
+	Name          string         `json:"name"    yaml:"name"`
+	Address       types.AddrPort `json:"address" yaml:"address"`
+	Ready         bool           `json:"ready"   yaml:"ready"`
+	APIExtensions []string       `json:"api_extensions" yaml:"api_extensions"`
```
masnax, line 12: Hmm, so does this mean that on the other end, checking for a particular extension means iterating through this list like any other slice?

Could we implement the JSON/YAML marshal/unmarshal commands and then pass this through as a real `Extensions` type so that the other end of the API request can use the Extension methods to check if this set contains a particular extension?

masnax (changes requested):
Just one question

internal/rest/types/server.go
masnax: Here's a reference for the `AddrPort` type: 
https://github.com/canonical/microcluster/blob/a1a192380cd5a3f534ba0d48bf7a838ef484d359/rest/types/addrport.go#L41-L82

gabrielmougard: the `Extensions` type is an `internal` type so we'll have to move `Extensions` out of `internal/` if we want to do that (which I'm happy to do if you think its best)

masnax: How come? The user won't have to actually declare the type to use methods from it.

gabrielmougard: hmmm.. you're right.. sorry for that. I'm reverting the change I just pushed

masnax:
```go
package main

import (
	"context"
	"fmt"

	"github.com/canonical/microcluster/microcluster"
)

func main() {
	m, _ := microcluster.App(microcluster.Args{})

	server, _ := m.Status(context.Background())

        // Call .HasExtension without importing anything.
	fmt.Println(server.Extensions.HasExtension("some-ext"))
}
```

gabrielmougard: @masnax updated :)

masnax (approved):
LGTM!

PR #173, opened 2024-07-08
internal/recover: Improve tarball handling during recovery

MggMuggins:
- createTarball can create tarballs with subdirs
- unpackTarball can extract tarballs with subdirs
- Separate (and un-nested) DB and State dirs are now supported

### DIFF
--- a/internal/recover/recover.go
+++ b/internal/recover/recover.go
@@ -10,6 +10,8 @@ import (
 	"io/fs"
 	"os"
 	"path"
+	"path/filepath"
+	"slices"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


internal/recover/recover.go
```diff
@@ -235,23 +224,32 @@ func CreateDatabaseBackup(filesystem *sys.OS) error {
 
 	logger.Info("Creating database backup", logger.Ctx{"archive": backupFilePath})
 
-	dbFS := os.DirFS(filesystem.StateDir)
-	dbFiles, err := fs.Glob(dbFS, "database/*")
+	// For DB backups the tarball should contain `database/` so that the user
+	// can easily untar the backup from the state dir.
+	// Forget ^ if DatabaseDir isn't inside StateDir
+	rootDir := filesystem.StateDir
+	walkDir, err := filepath.Rel(filesystem.StateDir, filesystem.DatabaseDir)
```
roosterfish, line 231: Not sure if more performant but why not using `filepath.Base(filesystem.DatabaseDir)`? It should give the same result and maybe is more intuitive?

Had to first check in the upstream docs what `Rel` is doing :) Not a blocker though.

internal/recover/recover.go
```diff
@@ -348,16 +362,25 @@ func unpackTarball(tarballPath string, destRoot string) error {
...
+		switch header.Typeflag {
+		case tar.TypeReg:
+			file, err := os.Create(filepath)
+			if err != nil {
+				return err
+			}
+
+			countWritten, err := io.Copy(file, tarReader)
+			if countWritten != header.Size {
+				return fmt.Errorf("mismatched written (%d) and size (%d) for entry %q in %q", countWritten, header.Size, header.Name, tarballPath)
```
roosterfish, line 375: To stay consistent the error message should start with uppercase.

roosterfish (changes requested):
Looks good, only a few small comments.

internal/recover/recover.go
MggMuggins: `filepath.Base` would assume that `DatabaseDir` is a directory inside `StateDir`. While I guess this is already sorta making that assumption, I wanted to capture any corner cases like:
```go
DatabaseDir = path.Join(StateDir, "somedir/database")
```

I clarified the comments a bit.

masnax (approved):
lgtm!

internal/recover/recover.go
roosterfish: I see, thanks.

roosterfish (approved):

PR #174, opened 2024-07-09
Extend additional network listener concept

roosterfish:
This PR adds missing features to the additional network listeners feature:

1. **Add additional servers by name**
   This allows referencing them later when updating their configuration and loading the server certificates
2. **API for setting arbitrary certificates**
   You can now set arbitrary certificates (besides the `cluster` certificate) which allows downstream projects to provide custom certificates. They are distributed across active cluster members and shared between new members joining the cluster .
3. **API to modify additional network listeners**
   The local daemon's config got moved into its own type which allows easy access to downstream components. This is used to allow updating the non-immutable fields of additional network listeners.
4. **Local daemon's config update hook**
   As soon as the local daemon's config receives an update, a new hook is triggered on all other cluster members which allows the entire cluster to be aware of configuration changes on individual members.

Furthermore the following changes have been applied:
* Use locking when accessing the list of the daemon's `extensionServers` field as this one is now accessed during startup and when modifying the additional network listeners.
* Store any additional certificate (besides `cluster`) in it's own `{state-dir}/certificates` directory. 
* Consistently use client structs (for request bodies) from the `rest/types` package. Another type `KeyPair` has been added for certificate management.
* Add helpers to the `Endpoint` type to allow updating and stopping individual endpoints
* A fix that ensures the `/cluster/1.0` endpoint can be called without errors in a pre-init state.

LXD-1174

### DIFF
--- a/example/api/servers.go
+++ b/example/api/servers.go
@@ -9,8 +9,8 @@ import (
 // Servers represents the list of listeners that the daemon will start
 // Each Server has pre-defined endpoints that will be added to the listener
 // If the Server is marked as CoreAPI, its endpoints will be added to the core listener of Microcluster.
-var Servers = []rest.Server{
-	{
+var Servers = map[string]rest.Server{
+	"extended": {
 		CoreAPI:   true,
 		ServeUnix: true,
 		Resources: []rest.Resources{
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -810,7 +810,7 @@ func (d *Daemon) addExtensionServers(preInit bool, fallbackCert *shared.CertInfo
 			continue
 		}
 
-		url := api.NewURL().Scheme(extensionServer.Protocol).Host(extensionServer.Address.String())
+		url := api.NewURL().Scheme("https").Host(extensionServer.Address.String())
 		alreadyRunning := false
 		for name := range d.endpoints.List(endpoints.EndpointNetwork) {
 			if name == serverName {
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -44,8 +44,7 @@ import (
 type Daemon struct {
 	project string // The project refers to the name of the go-project that is calling MicroCluster.
 
-	address api.URL                      // Listen Address.
-	config  *internalConfig.DaemonConfig // Local daemon's configuration from daemon.yaml file.
+	config *internalConfig.DaemonConfig // Local daemon's configuration from daemon.yaml file.
 
 	os         *sys.OS
 	serverCert *shared.CertInfo
@@ -453,9 +452,6 @@ func (d *Daemon) StartAPI(bootstrap bool, initConfig map[string]string, newConfi
 		}
 	}
 
-	// For convenience set the local daemon's server address together with its port.
-	d.address = *api.NewURL().Scheme("https").Host(d.config.GetAddress().String())
-
 	if bootstrap {
 		err := d.hooks.PreBootstrap(d.State(), initConfig)
 		if err != nil {
@@ -463,7 +459,7 @@ func (d *Daemon) StartAPI(bootstrap bool, initConfig map[string]string, newConfi
 		}
 	}
 
-	if d.address.URL.Host == "" || d.config.GetName() == "" {
+	if d.Address().URL.Host == "" || d.config.GetName() == "" {
 		return fmt.Errorf("Cannot start network API without valid daemon configuration")
 	}
 
@@ -472,7 +468,7 @@ func (d *Daemon) StartAPI(bootstrap bool, initConfig map[string]string, newConfi
 		return fmt.Errorf("Failed to parse server certificate when bootstrapping API: %w", err)
 	}
 
-	addrPort, err := types.ParseAddrPort(d.address.URL.Host)
+	addrPort, err := types.ParseAddrPort(d.Address().URL.Host)
 	if err != nil {
 		return fmt.Errorf("Failed to parse listen address when bootstrapping API: %w", err)
 	}
@@ -496,7 +492,7 @@ func (d *Daemon) StartAPI(bootstrap bool, initConfig map[string]string, newConfi
 
 	// Validate the extension servers again now that we have applied addresses.
 	d.extensionServersMu.RLock()
-	err = resources.ValidateEndpoints(d.extensionServers, d.address.URL.Host)
+	err = resources.ValidateEndpoints(d.extensionServers, d.Address().URL.Host)
 	if err != nil {
 		return err
 	}
@@ -509,13 +505,13 @@ func (d *Daemon) StartAPI(bootstrap bool, initConfig map[string]string, newConfi
 	}
 
 	serverEndpoints := []rest.Resources{resources.InternalEndpoints, resources.PublicEndpoints}
-	err = d.addCoreServers(false, d.address, d.ClusterCert(), serverEndpoints)
+	err = d.addCoreServers(false, *d.Address(), d.ClusterCert(), serverEndpoints)
 	if err != nil {
 		return err
 	}
 
 	// Add extension servers before post-join hook.
-	err = d.addExtensionServers(false, d.ClusterCert(), d.address.URL.Host)
+	err = d.addExtensionServers(false, d.ClusterCert(), d.Address().URL.Host)
 	if err != nil {
 		return err
 	}
@@ -532,7 +528,7 @@ func (d *Daemon) StartAPI(bootstrap bool, initConfig map[string]string, newConfi
 
 		clusterMember.SchemaInternal, clusterMember.SchemaExternal, _ = d.db.Schema().Version()
 
-		err = d.db.Bootstrap(d.Extensions, d.project, d.address, clusterMember)
+		err = d.db.Bootstrap(d.Extensions, d.project, *d.Address(), clusterMember)
 		if err != nil {
 			return err
 		}
@@ -552,12 +548,12 @@ func (d *Daemon) StartAPI(bootstrap bool, initConfig map[string]string, newConfi
 	}
 
 	if len(joinAddresses) != 0 {
-		err = d.db.Join(d.Extensions, d.project, d.address, joinAddresses...)
+		err = d.db.Join(d.Extensions, d.project, *d.Address(), joinAddresses...)
 		if err != nil {
 			return fmt.Errorf("Failed to join cluster: %w", err)
 		}
 	} else {
-		err = d.db.StartWithCluster(d.Extensions, d.project, d.address, d.trustStore.Remotes().Addresses())
+		err = d.db.StartWithCluster(d.Extensions, d.project, *d.Address(), d.trustStore.Remotes().Addresses())
 		if err != nil {
 			return fmt.Errorf("Failed to re-establish cluster connection: %w", err)
 		}
@@ -592,7 +588,7 @@ func (d *Daemon) StartAPI(bootstrap bool, initConfig map[string]string, newConfi
 		var clusterConfirmation bool
 		err = cluster.Query(d.shutdownCtx, true, func(ctx context.Context, c *client.Client) error {
 			// No need to send a request to ourselves.
-			if d.address.URL.Host == c.URL().URL.Host {
+			if d.Address().URL.Host == c.URL().URL.Host {
 				return nil
 			}
 
@@ -624,7 +620,7 @@ func (d *Daemon) StartAPI(bootstrap bool, initConfig map[string]string, newConfi
 		c.SetClusterNotification()
 
 		// No need to send a request to ourselves.
-		if d.address.URL.Host == c.URL().URL.Host {
+		if d.Address().URL.Host == c.URL().URL.Host {
 			return nil
 		}
 
@@ -735,7 +731,7 @@ func (d *Daemon) UpdateServers() error {
 
 	// Start any additional listener.
 	// This operation is idempotent.
-	err := d.addExtensionServers(false, d.ClusterCert(), d.address.URL.Host)
+	err := d.addExtensionServers(false, d.ClusterCert(), d.Address().URL.Host)
 	if err != nil {
 		return err
 	}
@@ -954,8 +950,7 @@ func (d *Daemon) ServerCert() *shared.CertInfo {
 
 // Address ensures both the daemon and state have the same address.
 func (d *Daemon) Address() *api.URL {
-	copyURL := d.address
-	return &copyURL
+	return api.NewURL().Scheme("https").Host(d.config.GetAddress().String())
 }
 
 // Name ensures both the daemon and state have the same name.
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -72,7 +72,7 @@ type Daemon struct {
 	stop func() error
 
 	extensionServersMu sync.RWMutex
-	extensionServers   []rest.Server
+	extensionServers   map[string]rest.Server
 }
 
 // NewDaemon initializes the Daemon context and channels.
@@ -114,7 +114,7 @@ func NewDaemon(project string) *Daemon {
 // - `extensionsSchema` is a list of schema updates in the order that they should be applied.
 // - `extensionServers` is a list of rest.Server that will be initialized and managed by microcluster.
 // - `hooks` are a set of functions that trigger at certain points during cluster communication.
-func (d *Daemon) Run(ctx context.Context, listenPort string, stateDir string, socketGroup string, extensionsSchema []schema.Update, apiExtensions []string, extensionServers []rest.Server, hooks *config.Hooks) error {
+func (d *Daemon) Run(ctx context.Context, listenPort string, stateDir string, socketGroup string, extensionsSchema []schema.Update, apiExtensions []string, extensionServers map[string]rest.Server, hooks *config.Hooks) error {
 	d.shutdownCtx, d.shutdownCancel = context.WithCancel(ctx)
 	if stateDir == "" {
 		stateDir = os.Getenv(sys.StateDir)
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -484,7 +484,7 @@ func (d *Daemon) StartAPI(bootstrap bool, initConfig map[string]string, newConfi
 		}
 	}
 
-	err = d.ReloadClusterCert()
+	err = d.ReloadCert("cluster")
 	if err != nil {
 		return err
 	}
@@ -896,18 +896,48 @@ func (d *Daemon) ClusterCert() *shared.CertInfo {
 	return shared.NewCertInfo(d.clusterCert.KeyPair(), d.clusterCert.CA(), d.clusterCert.CRL())
 }
 
-// ReloadClusterCert reloads the cluster keypair from the state directory.
-func (d *Daemon) ReloadClusterCert() error {
+// ReloadCert reloads a specific certificate from the filesytem.
+func (d *Daemon) ReloadCert(name string) error {
 	d.clusterMu.Lock()
 	defer d.clusterMu.Unlock()
 
-	clusterCert, err := util.LoadClusterCert(d.os.StateDir)
-	if err != nil {
-		return err
+	var dir string
+	if name == "cluster" {
+		dir = d.os.StateDir
+	} else {
+		dir = d.os.CertificatesDir
 	}
 
-	d.clusterCert = clusterCert
-	d.endpoints.UpdateTLS(clusterCert)
+	cert, err := shared.KeyPairAndCA(dir, name, shared.CertServer, true)
+	if err != nil {
+		return fmt.Errorf("Failed to load TLS certificate %q: %w", name, err)
+	}
+
+	// In case the cluster certificate gets reloaded also populate its value.
+	if name == "cluster" {
+		d.clusterCert = cert
+
+		// The core API endpoints are labeled with core.
+		// When the cluster certificate gets updated reload those.
+		d.endpoints.UpdateTLSByName(endpoints.EndpointsCore, cert)
+
+		// Reload all the other additional listeners that also use the cluster certificate.
+		// This might be the case if they
+		// - don't use a dedicated certificate
+		// - aren't part of the core API
+		// - and cannot load a custom certificate which shares their name
+		d.extensionServersMu.RLock()
+		for name, server := range d.extensionServers {
+			certExists := shared.PathExists(filepath.Join(d.os.CertificatesDir, fmt.Sprintf("%s.crt", name)))
+			if !server.CoreAPI && !server.DedicatedCertificate && !certExists {
+				d.endpoints.UpdateTLSByName(name, cert)
+			}
+		}
+
+		d.extensionServersMu.RUnlock()
+	} else {
+		d.endpoints.UpdateTLSByName(name, cert)
+	}
 
 	return nil
 }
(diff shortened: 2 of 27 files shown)


internal/rest/resources/certificates.go
```diff
@@ -72,19 +72,19 @@ func clusterCertificatesPut(s *state.State, r *http.Request) response.Response {
 			return response.BadRequest(fmt.Errorf("CA must be base64 encoded PEM key"))
 		}
 
-		err = os.WriteFile(filepath.Join(s.OS.StateDir, "cluster.ca"), []byte(req.CA), 0650)
+		err = os.WriteFile(filepath.Join(s.OS.StateDir, "cluster.ca"), []byte(req.CA), 0664)
```
markylaing: Why does microcluster need to handle CA certificates?

roosterfish: I can only speculate why this was added back in time. It got added in with https://github.com/canonical/microcluster/commit/561729a0441477dc4d0b7e8f718d16b452348404.

@masnax any ideas?

masnax: I believe it was Sunbeam that wanted this.

masnax (changes requested):
Looking good, just a handful of thoughts and some nits :)

internal/rest/resources/daemon.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	cluster, err := s.Cluster(false)
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	// Run the OnDaemonConfigUpdate hook on all other members.
+	remotes := s.Remotes()
+	err = cluster.Query(s.Context, true, func(ctx context.Context, c *client.Client) error {
```
mas-who, line 62: My understanding might be off here so please bear with me. AFAIK, `(microcluster.Microcluster).UpdateServers` is the only way we can update server configs during/after cluster initialisation. Once we updated the server configs then the `OnDaemonConfigUpdateHook` would trigger for every member daemon in the cluster. This is not an issue, I'm just trying to validate my understanding since this hook would be a convenient way for me to cache specific server addresses for all cluster members (a requirement in site manager)

mas-who (commented):
thanks for this! I left some comments, nothing major. Going to test it out locally as well. Will provide further feedback if anything else comes up

internal/daemon/daemon.go
```diff
@@ -211,23 +219,29 @@ func (d *Daemon) init(listenPort string, schemaExtensions []schema.Update, apiEx
 		listenAddr = listenAddr.Host(fmt.Sprintf(":%s", listenPort))
 	}
 
+	d.extensionServersMu.RLock()
```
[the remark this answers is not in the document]
mas-who: right I see that `extensionServers` may be updated at any point by a downstream project, so this is necessary correct?

internal/endpoints/endpoint.go
```diff
@@ -15,17 +15,17 @@ const (
...
 )
 
 // String labels EndpointTypes for logging purposes.
 func (et EndpointType) String() string {
 	switch et {
 	case EndpointControl:
 		return "control socket"
-	case EndpointNetwork:
-		return "https socket"
+	case EndpointCore:
```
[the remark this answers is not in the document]
roosterfish: I was just testing with an array of additional listeners and this is actually required. 

First I had the idea that an endpoints (networks) address is already enough to identify it. However when changing the address, you cannot anymore compare it against the others. I'll change it and pass a name instead.

internal/rest/resources/daemon.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/canonical/lxd/lxd/response"
+
+	"github.com/canonical/microcluster/rest"
+	"github.com/canonical/microcluster/rest/access"
+	"github.com/canonical/microcluster/rest/types"
+	"github.com/canonical/microcluster/state"
+)
+
+var daemonCmd = rest.Endpoint{
+	Path: "daemon/servers",
```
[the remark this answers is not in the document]
roosterfish: As we don't allow changing the entire daemon's config file my thinking was there will be other endpoints in the future which allow changing a specific subset of the local daemon's config.

example/api/servers.go
```diff
@@ -9,8 +9,8 @@ import (
 // Servers represents the list of listeners that the daemon will start
 // Each Server has pre-defined endpoints that will be added to the listener
 // If the Server is marked as CoreAPI, its endpoints will be added to the core listener of Microcluster.
-var Servers = []rest.Server{
-	{
+var Servers = map[string]rest.Server{
+	"extended": {
```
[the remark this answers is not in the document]
roosterfish: Valid point, I'll add a check.

internal/rest/resources/daemon.go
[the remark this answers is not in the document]
masnax: Isn't every state change effectively a change to the daemon's configuration? What is special about the `daemon` component here other than the internal implementation detail of changing something in the `daemon` package?

internal/daemon/daemon.go
```diff
@@ -781,14 +781,16 @@ func (d *Daemon) addExtensionServers(preInit bool, fallbackCert *shared.CertInfo
 			continue
 		}
 
+		customCertExists := shared.PathExists(filepath.Join(d.os.CertificatesDir, fmt.Sprintf("%s.crt", name)))
```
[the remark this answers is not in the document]
roosterfish: If the `*.crt` exists then `shared.KeyPairAndCA()` will try to load it. It will error out if any of the files is missing.

internal/daemon/daemon.go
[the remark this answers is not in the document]
roosterfish: Earlier to this change only the daemon accessed `d.extensionServers` as part of bootstrapping the local microcluster. 
Now there is an API endpoint which allows you to modify the mutable fields of the additional listeners so we need to make sure that write access to this map isn't concurrent.

internal/rest/resources/daemon.go
roosterfish: Yes that is right!

mas-who:
I have a question about a particular use case. During cluster initialisation or member join, let's say we are asking a user to supply the address for a listener, is there a way to group that listener's endpoints with the core API listener if an address is not supplied?

internal/rest/resources/certificates.go
```diff
@@ -65,34 +73,41 @@ func clusterCertificatesPut(s *state.State, r *http.Request) response.Response {
 		return response.BadRequest(fmt.Errorf("Private key must be base64 encoded PEM key"))
 	}
 
+	// Validate the certificate's name.
```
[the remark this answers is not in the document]
roosterfish: Fixed.

example/api/servers.go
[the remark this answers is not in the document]
roosterfish: Fixed

internal/daemon/daemon.go
```diff
@@ -44,8 +44,8 @@ import (
 type Daemon struct {
 	project string // The project refers to the name of the go-project that is calling MicroCluster.
 
-	address api.URL // Listen Address.
-	name    string  // Name of the cluster member.
+	address api.URL                      // Listen Address.
```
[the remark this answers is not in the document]
roosterfish: Fixed.

internal/rest/resources/daemon.go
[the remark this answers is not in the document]
roosterfish: Ah sorry I read it the other way around naming it `/daemon` not `/servers`.

As the path is `/cluster/internal/servers` I had the feeling it makes more sense to also put `daemon` so that it's clear we are modifying the local daemon instead of any internal cluster setting.

internal/endpoints/endpoint.go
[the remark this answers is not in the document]
roosterfish: Fixed, renamed back to `EndpointNetwork` and got rid of the third type.

rest/rest.go
```diff
@@ -57,9 +56,9 @@ type Server struct {
 	// ServeUnix sets whether the resources of this endpoint should also be served over the unix socket.
 	ServeUnix bool
 
-	// Certificate is used for setting up TLS on the server.
-	// x509 PEM Certificate
-	Certificate *shared.CertInfo
+	// DedicatedCertificate sets whether the additional lister should use it's own self signed certificate.
```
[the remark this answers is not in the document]
roosterfish: Fixed

rest/types/server.go
```diff
@@ -0,0 +1,12 @@
+package types
+
+// ServerConfig represents the non-immutable fields of an additional network listener.
```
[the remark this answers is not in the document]
roosterfish: Fixed.

internal/daemon/daemon.go
```diff
@@ -409,20 +428,32 @@ func (d *Daemon) initServer(resources ...rest.Resources) *http.Server {
...
+	} else {
+		err := d.config.Load()
 		if err != nil {
-			return fmt.Errorf("Failed to apply and save new daemon configuration: %w", err)
+			return err
 		}
 	}
 
+	// For convenience set the local daemon's server address together with its port.
+	d.address = *api.NewURL().Scheme("https").Host(d.config.GetAddress().String())
```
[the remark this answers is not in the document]
roosterfish: Fixed.

roosterfish:
@mas-who no that would not be possible. When you define the additional listener you have to decide if it should be part of the core API or should get spawned with it's own server.

mas-who:
> > @roosterfish I have a question about a particular use case.
> 
> @mas-who no that would not be possible. When you define the additional listener you have to decide if it should be part of the core API or should get spawned with it's own server.

Okay not a problem, I guess I can just default the address to the same core listener host but a different port if it's not supplied by the user :+1:

masnax:
> > > @roosterfish I have a question about a particular use case.
> > 
> > 
> > @mas-who no that would not be possible. When you define the additional listener you have to decide if it should be part of the core API or should get spawned with it's own server.
> 
> Okay not a problem, I guess I can just default the address to the same core listener host but a different port if it's not supplied by the user 👍

I think using another port is better design in general because it would be very weird for endpoints to be conflated with the management network just because they were ignored during initialization.

internal/rest/resources/certificates.go
```diff
@@ -78,26 +78,33 @@ func clusterCertificatesPut(s *state.State, r *http.Request) response.Response {
 		return response.BadRequest(fmt.Errorf("Certificate name cannot be a path"))
 	}
 
+	var certificateDir string
+	if certificateName == "cluster" {
```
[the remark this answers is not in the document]
roosterfish: Added.

internal/rest/resources/daemon.go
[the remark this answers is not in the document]
roosterfish: With `/cluster/internal` I am missing the bit that tells me I am modifying the actual daemon, not something internally that affects the whole cluster.

Let's take `/cluster/internal/sql` as an example. It allows you to run a query against the cluster DB but it's not daemon specific.

roosterfish:
@masnax Ready for another review now :)

internal/rest/resources/daemon.go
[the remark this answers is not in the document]
masnax: That's actually a good point but I would argue `/sql` should be moved to `/cluster/sql` instead. 

I guess what I'm confused about is why is a "server" considered a daemon component, but the database is not?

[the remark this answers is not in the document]
masnax: Are you saying that because it's a local-only action then it's a "daemon" component? But since we execute a cluster-wide hook when we run this endpoint then its not truly local-only.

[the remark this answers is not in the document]
roosterfish: > Are you saying that because it's a local-only action then it's a "daemon" component? 

This.

As the hooks are triggered afterwards by calling `/cluster/internal/hooks/{type}` (which is not daemon specific) I guess that is fine.

I agree moving `/sql` one level up.

internal/daemon/daemon.go
```diff
@@ -622,11 +652,87 @@ func (d *Daemon) StartAPI(bootstrap bool, initConfig map[string]string, newConfi
...
+
+			err := d.endpoints.DownByName(serverName)
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	// Restart all additional listeners which received an update.
+	for serverName, extensionServerConfig := range d.config.GetServers() {
```
masnax, line 689: If you instantiate `GetServers()` once, you won't have to do the map copying twice.

internal/daemon/daemon.go
```diff
@@ -147,7 +150,9 @@ func (d *Daemon) Run(ctx context.Context, listenPort string, stateDir string, so
 		return fmt.Errorf("Database recovery failed: %w", err)
 	}
 
+	d.extensionServersMu.Lock()
 	d.extensionServers = extensionServers
```
masnax, line 154: Just to be safe, you should probably copy the supplied `extensionServers` into the daemon's struct because you modify the map down in `UpdateServers` which might have unintended consequences on the downstream project who supplied the map.

internal/daemon/daemon.go
```diff
@@ -622,11 +652,87 @@ func (d *Daemon) StartAPI(bootstrap bool, initConfig map[string]string, newConfi
...
+	return nil
+}
+
 // startUnixServer starts up the core unix listener with the given resources.
 func (d *Daemon) startUnixServer(serverEndpoints []rest.Resources) error {
 	ctlServer := d.initServer(serverEndpoints...)
 	ctl := endpoints.NewSocket(d.shutdownCtx, ctlServer, d.os.ControlSocket(), d.os.SocketGroup)
-	d.endpoints = endpoints.NewEndpoints(d.shutdownCtx, ctl)
+	d.endpoints = endpoints.NewEndpoints(d.shutdownCtx, map[string]endpoints.Endpoint{
+		"unix": ctl,
```
masnax, line 734: Does this and `core` below deserve a constant?

internal/daemon/daemon.go
```diff
@@ -746,18 +884,48 @@ func (d *Daemon) ClusterCert() *shared.CertInfo {
...
 
-	d.clusterCert = clusterCert
-	d.endpoints.UpdateTLS(clusterCert)
+	// In case the cluster certificate gets reloaded also populate its value.
+	if name == internalTypes.ClusterCertificate {
+		d.clusterCert = cert
+
+		// The core API endpoints are labeled with core.
+		// When the cluster certificate gets updated reload those.
+		d.endpoints.UpdateTLSByName("core", cert)
```
masnax, line 910: I thought you were going to use `cluster` as the name for the core servers so that you could reuse `internalTypes.ClusterCertificate`?

internal/rest/client/cluster.go
```diff
@@ -62,10 +62,10 @@ func (c *Client) ResetClusterMember(ctx context.Context, name string, force bool
 }
 
 // UpdateClusterCertificate sets a new cluster keypair and CA.
-func (c *Client) UpdateClusterCertificate(ctx context.Context, args apiTypes.ClusterCertificatePut) error {
+func (c *Client) UpdateClusterCertificate(ctx context.Context, name string, args apiTypes.KeyPair) error {
```
masnax, line 65: I'm thinking from an external project's use case, they probably won't know to call `UpdateClusterCertificate` with `"cluster"` as an argument if they want to change their core certificate. 

So either we should document this in the client functions, and move `types.ClusterCertificate` into the public types package, or we should add a second client function `UpdateCoreClusterCertificate()` which doesn't take the name and internally uses `types.ClusterCertificate`.

internal/rest/types/cluster.go
```diff
@@ -48,3 +48,8 @@ const (
 	// MemberNeedsUpgrade should be the MemberStatus if the system needs to receive a schema upgrade to be compatible with other cluster members.
 	MemberNeedsUpgrade MemberStatus = "NEEDS UPGRADE"
 )
+
+const (
+	// ClusterCertificate represents the name of the cluster certificate used by the core API.
+	ClusterCertificate string = "cluster"
```
masnax, line 54: This should probably be called `ClusterCertificateName` to avoid any confusion.

rest/rest.go
```diff
@@ -56,17 +57,9 @@ type Server struct {
...
-	// Address is the server listen address.
-	// Example: 127.0.0.1:9000
-	Address types.AddrPort
-
-	// Certificate is used for setting up TLS on the server.
-	// x509 PEM Certificate
-	Certificate *shared.CertInfo
+	// DedicatedCertificate sets whether the additional listener should use its own self signed certificate.
+	// If false the default cluster certificate from the core API is used.
+	DedicatedCertificate bool
```
masnax, line 62: Could you please add to the comment here that the certificate will be searched for in the `/certificates` directory of the state directory, with the name given to the server when supplied to `microcluster.Start`

rest/types/config.go
```diff
@@ -0,0 +1,8 @@
+package types
+
+// Config is the in memory version of the local daemon.yaml file.
+type Config struct {
```
masnax, line 4: Perhaps this should be called `DaemonConfig`? I know it stammers with the `config` package's type but since this is a public type, if we ever decide that we need more types of config and this name is too generic, then every project will have to refactor their code. (Like with `Server` for the past few weeks :p)

masnax (changes requested):
Looking very good, just a few questions & nits and we should be good to go!

internal/daemon/daemon.go
roosterfish: Good catch thanks!

internal/daemon/daemon.go
roosterfish: Jup but I decided against as we tend to label everything with core/core API so to make it more clear I didn't use `cluster` at the end.

Now by having a constant for both `core` and `unix` (as you suggested) it makes even more sense :)

internal/rest/client/cluster.go
roosterfish: I am a fan of the latter as having another client func doesn't hurt and it's presumably easier to follow than reading the doc string of the function. Will add another one.

roosterfish: After thinking a bit more about it I went with using a specific type as it feels more intuitive:

```go
func (m *MicroCluster) UpdateCertificate(ctx context.Context, name types.CertificateName, keypair types.KeyPair) error {}

func (c *Client) UpdateCertificate(ctx context.Context, name apiTypes.CertificateName, args apiTypes.KeyPair) error {}
```

I have also removed `cluster` from the functions name as this might be confusing now.
If you have custom cert you can now derive a new type from `type CertificateName string` and supply it to the function.
Or use `const ClusterCertificateName CertificateName = "cluster"` in case you want to update the cluster one.

rest/types/config.go
roosterfish: Thanks will fix! Previously I had this type together with the daemon's `DaemonConfig` internal type but when implementing the hook I have moved it into the public types package but haven't changed it's name.

internal/daemon/daemon.go
roosterfish: Fixed.

internal/daemon/daemon.go
roosterfish: Fixed

internal/daemon/daemon.go
roosterfish: Fixed

internal/daemon/daemon.go
roosterfish: Fixed.

internal/rest/client/cluster.go
roosterfish: Fixed

internal/rest/types/cluster.go
roosterfish: Fixed

rest/types/config.go
roosterfish: Fixed

microcluster/app.go
```diff
@@ -44,9 +44,9 @@ type Args struct {
 	StateDir    string
 	SocketGroup string
 
-	ListenPort string
-	Client     *client.Client
-	Proxy      func(*http.Request) (*url.URL, error)
+	ListenAddress string
```
masnax, line 47: I'm fine with making this a whole address, but can you please rename it to `PreInitListenAddress` to make it clear what it's for?

Also, this is going to break the daemon's authentication because we detect if we are in a pre-init listening mode by checking if the address is blank: https://github.com/canonical/microcluster/blob/fe0e5ecba28a2898ff81bbe71b60e87300b610da/rest/access/handlers.go#L56-L59

masnax (changes requested):
Few changes need to be made to the pre-init listen address now being supplied to the daemon directly.

microcluster/app.go
masnax: It's not sufficient to check for the database status because the core API comes up before the database does.

I suppose you can add a helper to `Endpoints` to fetch the certificate used by the core server and check if it's the `server.crt` here.

roosterfish: Good catch, thx. Should we check for `state.Database.Status() == StatusNotReady` instead?

roosterfish: Nvm, I saw your other comment too late :)

roosterfish: But in between the core API being started and the DB becoming ready, is there any requirement to already authenticate requests?
Just wondering because the check for `StatusNotReady` is "cheap" as it will run every time we authenticate a request.

masnax: I'm mainly concerned about races and ongoing connections leading to security issues. Think about this:
* You use `PreInitListenAddress` to allow all incoming connections without authentication
* You run `init` to set up your system
* A bad actor queries your cluster on an endpoint with `AllowedBeforeInit=true`, database is offline so it's allowed
* It's an expensive endpoint and takes time to run
* The database comes online
* The request is still ongoing and at this point checks if the database is online, it is so it makes database changes.



Besides, just fetching the cert from `Endpoints` is just as cheap, it's [stored here already](https://github.com/canonical/microcluster/blob/fe0e5ecba28a2898ff81bbe71b60e87300b610da/internal/endpoints/network.go#L20)

roosterfish: I see makes sense. I have just pushed the check for the server's certificate. That feels to be the most bullet proof.

I have tested this with the `/cluster/1.0/ready` endpoint as it is allowed before init and doesn't require the DB (which would post the "error not initialized" prior to init).

rest/access/handlers.go
```diff
@@ -53,9 +54,18 @@ func Authenticate(state *state.State, r *http.Request, hostAddress string, trust
 		return true, nil
 	}
 
-	if state.Address().URL.Host == "" {
-		logger.Info("Allowing unauthenticated request to un-initialized system")
-		return true, nil
+	// Check if the listener is using the server.crt.
+	// This indicates that the daemon is in a pre-init state and is listening on the PreInitListenAddress.
+	serverCertificate := state.ServerCert()
+	networkEndpoints := state.Endpoints.List(endpoints.EndpointNetwork)
```
masnax, line 60: Hmm, I wonder if it's safe to drop all authentication if _any_ listener uses the `server.crt`. I think we should only check the `core` listener. 

It's a small edge case, but if you have an extension server that perpetually uses the `server.crt`, then every request to every endpoint would always skip authentication.

roosterfish: True, if somebody reads the `server` keypair and pushes it to the `/certificates` dir this will have the side effect that every request can run unauthenticated.

Fixed!

roosterfish:
I was wondering one more thing.
We have to reject any request with `m.AddServers()` that uses `core` and `unix` in its name as this will conflict with the core API and Unix listeners.

roosterfish:
> I was wondering one more thing. We have to reject any request with `m.AddServers()` that uses `core` and `unix` in its name as this will conflict with the core API and Unix listeners.

Fixed.

masnax (approved):
@roosterfish Thanks, looks good to me. 

One thing, if possible could you add some tests in a separate PR to cover a some basic setup of multiple listeners?

masnax:
@roosterfish looks like some commits aren't signed

roosterfish:
> One thing, if possible could you add some tests in a separate PR to cover a some basic setup of multiple listeners?

Whoops, they are all signed now. 
Sure I'll add a few tests in a follow up PR.

PR #178, opened 2024-07-10
Integrate dqlite heartbeats

masnax:
Uses the latest `go-dqlite`, allowing us to integrate microcluster's heartbeat with `go-dqlite`'s own heartbeats. 

This means there won't be redundant heartbeats, and also that microcluster's status of nodes will be more up to date than before.

Additionally, due to the reduced overhead from reaching out to the dqlite leader fewer times, we can make the heartbeats occur more frequently with only a minimal performance hit. The new heartbeat interval is every 10 seconds, compared to 30 seconds before. At about 20s, the performance would be identical to before. The heartbeat interval can now also be configured by the downstream project if 10s is too aggressive.

The final commit of this PR sets the maximum concurrent connections we will maintain when probing for the dqlite leader. Upstream go-dqlite has set this to 10 by default, whereas the old behaviour was sequential only. This commit explicitly sets it back to 1. In a future PR, we can try to increment this value if the heartbeats we receive are unhealthy.


go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/canonical/microcluster
 
 go 1.22.4
 
+replace github.com/canonical/go-dqlite => github.com/canonical/go-dqlite v0.0.0-20240710170928-54a06bec1336
```
roosterfish, line 5: Why does this have to be in a `replace` statement? Can't we just update the regular list of dependencies and replace the `github.com/canonical/go-dqlite v1.21.0` line?

internal/db/dqlite.go
```diff
@@ -51,14 +51,20 @@ type DB struct {
...
 
 	schema *update.SchemaUpdate
 
 	statusLock sync.RWMutex
 	status     Status
 }
 
+const (
+	// DefaultHeartbeartInterval is the default interval used for heartbeats and dqlite role probes.
+	DefaultHeartbeartInterval time.Duration = time.Second * 10
```
roosterfish, line 66: `Heartbeart`, same in the comment above

roosterfish (changes requested):
Looks nice, just a small typo and question about the dependency update.

go.mod
masnax: Seems to get overwritten with `go mod tidy` if we do that.

internal/db/dqlite.go
masnax: Oops :)

go.mod
roosterfish: You can run `go get github.com/canonical/go-dqlite@54a06bec1336` to get this exact commit so it doesn't fall back to the latest tag.

roosterfish:
Requires a rebase now.

go.mod
masnax: They have a tagged release with these features now, so problem solved :)

internal/rest/resources/heartbeat.go
```diff
@@ -162,27 +141,11 @@ func beginHeartbeat(s state.State, r *http.Request) response.Response {
...
-		}
-
-		// Sleep at least 2 seconds to sync up with other nodes.
-		if sleepInterval < 2*time.Second {
-			sleepInterval = 2 * time.Second
-		}
-
-		logger.Debugf("Heartbeat was sent %v ago, sleep %v seconds before retrying", timeSinceLast, sleepInterval)
-		<-time.After(sleepInterval)
+		logger.Debugf("Heartbeat was already sent %s ago, skipping heartbeat round", timeSinceLast.String())
```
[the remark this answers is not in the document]
masnax: Fixed.

roosterfish:
Requires a rebase now.

masnax:
Rebased now.

roosterfish:
Just one more thing, there is a duplicate import now https://github.com/canonical/microcluster/pull/178/files#diff-d0560775f9ab1fefca803a7f4106abec57f83bb55f0eef246fc21c0e5461776aR31

masnax:
> Just one more thing, there is a duplicate import now https://github.com/canonical/microcluster/pull/178/files#diff-d0560775f9ab1fefca803a7f4106abec57f83bb55f0eef246fc21c0e5461776aR31

Whoops, fixed now

roosterfish (approved):
LGTM!

PR #179, opened 2024-07-11
Allow empty address for `GET /cluster/1.0`

masnax:
This API endpoint should be available at all times to query the daemons' state information. However, if it is not yet initialized, its address may not be set if the project is not using `PreInitListenAddress`. 

In these cases, there is technically no address yet so we can just return an empty `types.AddrPort` instead.

Calling `String()` on this type would return `invalid AddrPort` which I found to be a bit unintuitive so instead the wrapper type has been given its own `String` method which returns an empty string if the AddrPort is empty, and otherwise falls back to the internal `String` implementation.

### DIFF
--- a/internal/rest/resources/api_1.0.go
+++ b/internal/rest/resources/api_1.0.go
@@ -20,7 +20,7 @@ var api10Cmd = rest.Endpoint{
 func api10Get(s *state.State, r *http.Request) response.Response {
 	addrPort, err := types.ParseAddrPort(s.Address().URL.Host)
 	if err != nil {
-		return response.SmartError(err)
+		addrPort = types.AddrPort{}
 	}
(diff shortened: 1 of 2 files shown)


masnax:
@gabrielmougard This should fix the issue you were experiencing when trying to fetch the API extensions.

gabrielmougard (approved):
LGTM

roosterfish (approved):
Looks good!

I thought this was fixed with https://github.com/canonical/microcluster/pull/174/commits/a0c3e6e4acea2c6106432b30732dc5a049eb8bd2 but I didn't think about the case where you don't set the `PreInitListenAddress` but still try to access `/cluster/1.0` on the Unix socket.

PR #180, opened 2024-07-12
internal/rest/resources: fix conflicting endpoint error for same paths on different listeners

mas-who:
## Done
Currently we update the `existingPaths` map with both coreAPI and extension server endpoint paths. This will result in conflicting endpoint error in `resourcesConflict` if any server was evaluated before the coreAPI server. The error would also occur non-deterministically because `extensionServers` is a map and is iterated in `ValidateEndpoints`. 

To fix this, we can introduce a map `extensionServerPaths` in `resourcesConflict` to specifically check for conflicts within an extension server, this prevents the `existingPaths` map getting updated with non-coreAPI server paths.

### DIFF
--- a/internal/rest/resources/resources.go
+++ b/internal/rest/resources/resources.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"path/filepath"
 
+	"github.com/canonical/microcluster/internal/endpoints"
 	internalTypes "github.com/canonical/microcluster/internal/rest/types"
 	"github.com/canonical/microcluster/rest"
 	"github.com/canonical/microcluster/rest/types"
@@ -51,21 +52,21 @@ var InternalEndpoints = rest.Resources{
 // - The PathPrefix+Path of an endpoint conflicts with another endpoint in the same server.
 // - The address of the server clashes with another server.
 // - The server does not have defined resources.
-// If the Server is a core API server, its resources must not conflict with any other server, and it must not have a defined address or certificate.
+// If the Server is a core API server, its resources must not conflict with any other core API server, and it must not have a defined address or certificate.
 func ValidateEndpoints(extensionServers map[string]rest.Server, coreAddress string) error {
-	allExistingEndpoints := []rest.Resources{UnixEndpoints, PublicEndpoints, InternalEndpoints}
-	existingEndpointPaths := make(map[string]bool)
 	serverAddresses := map[string]bool{coreAddress: true}
+	baseCoreEndpoints := []rest.Resources{UnixEndpoints, PublicEndpoints, InternalEndpoints}
+	existingEndpointPaths := map[string]map[string]bool{endpoints.EndpointsCore: {}}
 
-	// Record the paths for all internal endpoints.
-	for _, endpoints := range allExistingEndpoints {
-		for _, e := range endpoints.Endpoints {
-			url := filepath.Join(string(endpoints.PathPrefix), e.Path)
-			existingEndpointPaths[url] = true
+	// Record the paths for all internal endpoints on the core listener.
+	for _, existing := range baseCoreEndpoints {
+		for _, e := range existing.Endpoints {
+			url := filepath.Join(string(existing.PathPrefix), e.Path)
+			existingEndpointPaths[endpoints.EndpointsCore][url] = true
 		}
 	}
 
-	for _, server := range extensionServers {
+	for serverName, server := range extensionServers {
 		// Ensure all servers have resources.
 		if len(server.Resources) == 0 {
 			return fmt.Errorf("Server must have defined resources")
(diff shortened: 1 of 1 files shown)


markylaing:
I could very much be wrong about this but if extension servers all have names (since @roosterfish's recent PR I think?), then why not have a map of extension server name to a map of resource paths? Otherwise you risk running into this problem again when paths of two extension servers collide.

mas-who:
> I could very much be wrong about this but if extension servers all have names (since @roosterfish's recent PR I think?), then why not have a map of extension server name to a map of resource paths? Otherwise you risk running into this problem again when paths of two extension servers collide.

I'm not sure if I understand, the paths are evaluated for each extension server independently?

roosterfish:
> I'm not sure if I understand, the paths are evaluated for each extension server independently?

Yes, I also think there can't be a conflict between two non core API extension servers.

roosterfish (dismissed):
Looks good to me but will hold back until @markylaing's comment is resolved.

masnax (changes requested):
Yeah I concur with @markylaing, I think in general this will be easier to reason about if we form the whole map ahead of time and then validate the map without changing it again.

Right now it's a bit confusing to figure out what all the maps are doing. Also it seems the behaviour in this comment was dropped:
```
// If the server uses the core API, its resources must not conflict with any existing listener.
```
But with the changes here we only compare a `Core` server to other `Core` servers, which means an extension server can now have a conflicting endpoint with a `Core` server.

roosterfish:
> But with the changes here we only compare a `Core` server to other `Core` servers, which means an extension server can now have a conflicting endpoint with a `Core` server.

How would that be possible? If the core server is listening on `:9001` and the extension server on `:9002`, how could the endpoints conflict? 
I think it's reasonable to allow having e.g. a `/foo` endpoint on both listeners `:9001` and `:9002`.

masnax:
> > But with the changes here we only compare a `Core` server to other `Core` servers, which means an extension server can now have a conflicting endpoint with a `Core` server.
> 
> How would that be possible? If the core server is listening on `:9001` and the extension server on `:9002`, how could the endpoints conflict? I think it's reasonable to allow having e.g. a `/foo` endpoint on both listeners `:9001` and `:9002`.

Yes, I meant the comment needed updating.

mas-who:
> Yeah I concur with @markylaing, I think in general this will be easier to reason about if we form the whole map ahead of time and then validate the map without changing it again.
> 
> Right now it's a bit confusing to figure out what all the maps are doing. Also it seems the behaviour in this comment was dropped:
> 
> ```
> // If the server uses the core API, its resources must not conflict with any existing listener.
> ```
> 
> But with the changes here we only compare a `Core` server to other `Core` servers, which means an extension server can now have a conflicting endpoint with a `Core` server.

Yeah that makes sense :+1: will setup the map ahead of time and also update comments

mas-who:
> > Yeah I concur with @markylaing, I think in general this will be easier to reason about if we form the whole map ahead of time and then validate the map without changing it again.
> > Right now it's a bit confusing to figure out what all the maps are doing. Also it seems the behaviour in this comment was dropped:
> > ```
> > // If the server uses the core API, its resources must not conflict with any existing listener.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > But with the changes here we only compare a `Core` server to other `Core` servers, which means an extension server can now have a conflicting endpoint with a `Core` server.
> 
> Yeah that makes sense 👍 will setup the map ahead of time and also update comments

edit

mas-who:
@masnax I've made the update to generate map of core server paths and extension server paths (grouped by server name) before validating endpoints. It is a little strange since we have to iterate `extensionServers` twice but I do think it's easier to figure out the logic. Let me know what you think.

internal/rest/resources/resources.go
```diff
@@ -92,49 +83,66 @@ func ValidateEndpoints(extensionServers map[string]rest.Server, coreAddress stri
...
-		serverPaths, err := resourcesConflict(server, existingEndpointPaths)
+		existingPaths := extensionPaths[serverName]
+		if server.CoreAPI {
+			existingPaths = corePaths
+		}
+
+		err := resourcesConflict(server, existingPaths)
 		if err != nil {
 			return err
 		}
```
masnax, line 94: Hmm, right now I think this would always return an error for all valid Servers. 

`resourcesConflict` doesn't make sense anymore because it's checking for existence in the map and erroring out, but the map it's supplied with is the set of all paths that exist on the server, so it will always error for that server.

internal/rest/resources/resources.go
```diff
@@ -92,49 +83,66 @@ func ValidateEndpoints(extensionServers map[string]rest.Server, coreAddress stri
...
+	}
+
+	for name, server := range extensionServers {
+		for _, resource := range server.Resources {
+			for _, endpoint := range resource.Endpoints {
+				url := filepath.Join(string(resource.PathPrefix), endpoint.Path)
+				if server.CoreAPI {
+					corePaths[url] = true
+					continue
+				}
```
masnax, line 121: If `corePaths[url]` is already `true` here then that means this is a duplicate path.

internal/rest/resources/resources.go
```diff
@@ -92,49 +83,66 @@ func ValidateEndpoints(extensionServers map[string]rest.Server, coreAddress stri
...
 
-		// Update the existing paths now that we have checked a new Server.
-		for k, v := range serverPaths {
-			if !existingEndpointPaths[k] {
-				existingEndpointPaths[k] = v
+				if extensionPaths[name] == nil {
+					extensionPaths[name] = make(map[string]bool)
+				}
+
+				extensionPaths[name][url] = true
```
masnax, line 127: Same here, `extensionPaths[name][url]` might already be `true` which is what we are looking for, but that won't be captured if we overwrite the `true` with another `true`.

masnax (changes requested):
Doesn't work quite right, and it's a bit longer than necessary. Since we are checking for duplicates, it might not be so intuitive to create the map beforehand, we would have to use a `map[bool]int`. Sorry for leading you down that path. Instead how about something like this to keep just 1 map:

```go
func ValidateEndpoints(extensionServers map[string]rest.Server, coreAddress string) error {
	allExistingEndpoints := []rest.Resources{UnixEndpoints, PublicEndpoints, InternalEndpoints}
	existingEndpointPaths := map[string]map[string]bool{endpoints.EndpointsCore: {}}

	// Record the paths for all internal endpoints.
	for _, existing := range allExistingEndpoints {
		for _, e := range existing.Endpoints {
			url := filepath.Join(string(existing.PathPrefix), e.Path)
			existingEndpointPaths[endpoints.EndpointsCore][url] = true
		}
	}

	for serverName, server := range extensionServers {
		// FIXME: Other Server checks go here.

		// Ensure no endpoint path conflicts with another endpoint on the same server.
		// If a server is an extension to the core API, it will be compared against all core API paths.
		for _, resource := range server.Resources {
			for _, e := range resource.Endpoints {
				url := filepath.Join(string(resource.PathPrefix), e.Path)

				// Internally, core API servers will be lumped together with the "core" server so
				// record them in the map under that name. Otherwise, use the given server name.
				internalName := serverName
				if server.CoreAPI {
					internalName = endpoints.EndpointsCore
				}

				_, ok := existingEndpointPaths[internalName]
				if !ok {
					existingEndpointPaths[internalName] = map[string]bool{}
				}

				if existingEndpointPaths[internalName][url] {
					return fmt.Errorf("Endpoint %q from server %q conflicts with another endpoint from server %q", url, serverName, internalName)
				}

				existingEndpointPaths[internalName][url] = true
			}
		}
	}

	return nil
}
```

mas-who:
> Doesn't work quite right, and it's a bit longer than necessary. Since we are checking for duplicates, it might not be so intuitive to create the map beforehand, we would have to use a `map[bool]int`. Sorry for leading you down that path. Instead how about something like this to keep just 1 map:
> 
> ```go
> func ValidateEndpoints(extensionServers map[string]rest.Server, coreAddress string) error {
> 	allExistingEndpoints := []rest.Resources{UnixEndpoints, PublicEndpoints, InternalEndpoints}
> 	existingEndpointPaths := map[string]map[string]bool{endpoints.EndpointsCore: {}}
> 
> 	// Record the paths for all internal endpoints.
> 	for _, existing := range allExistingEndpoints {
> 		for _, e := range existing.Endpoints {
> 			url := filepath.Join(string(existing.PathPrefix), e.Path)
> 			existingEndpointPaths[endpoints.EndpointsCore][url] = true
> 		}
> 	}
> 
> 	for serverName, server := range extensionServers {
> 		// FIXME: Other Server checks go here.
> 
> 		// Ensure no endpoint path conflicts with another endpoint on the same server.
> 		// If a server is an extension to the core API, it will be compared against all core API paths.
> 		for _, resource := range server.Resources {
> 			for _, e := range resource.Endpoints {
> 				url := filepath.Join(string(resource.PathPrefix), e.Path)
> 
> 				// Internally, core API servers will be lumped together with the "core" server so
> 				// record them in the map under that name. Otherwise, use the given server name.
> 				internalName := serverName
> 				if server.CoreAPI {
> 					internalName = endpoints.EndpointsCore
> 				}
> 
> 				_, ok := existingEndpointPaths[internalName]
> 				if !ok {
> 					existingEndpointPaths[internalName] = map[string]bool{}
> 				}
> 
> 				if existingEndpointPaths[internalName][url] {
> 					return fmt.Errorf("Endpoint %q from server %q conflicts with another endpoint from server %q", url, serverName, internalName)
> 				}
> 
> 				existingEndpointPaths[internalName][url] = true
> 			}
> 		}
> 	}
> 
> 	return nil
> }
> ```

Yes I agree, sorry the logic is off. Checking conflicts with map of existing paths built ahead of time means we won't be checking duplicates. I've made the update as suggested and tested it with site manager since there are different listeners there with conflicitng paths (which shouldn't error).

masnax (approved):
LGTM, thanks @mas-who

PR #186, opened 2024-07-16
internal/db: Load `StateDir/patch.global.sql` during schema Ensure

MggMuggins:
This is modelled after the same feature in LXD:
  https://github.com/canonical/lxd/blob/34a108012a4b47c045503c9651a37091ddb8bacb/lxd/db/schema/schema.go#L122

This provides a way for an administrator to execute sql during database startup before any schema updates are applied.

It also allows us to queue up changes to the global DB before it has been started (useful during recovery).

LXD-1316

### DIFF
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -8,6 +8,7 @@ import (
 	"math/rand"
 	"net/http"
 	"os"
+	"path"
 	"time"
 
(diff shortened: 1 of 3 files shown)


masnax (approved):
lgtm

PR #187, opened 2024-07-16
.github: Run example system tests in CI

MggMuggins:
I don't see any reason not to run those tests since they exist. Happy to make some modifications to them if that would be desirable.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -61,7 +61,10 @@ jobs:
 
           make -C example check-unit
 
-      - name: Make example
+      - name: Example system tests
         run: |
           set -eux
           make -C example
+
+          cd example/test
+          CLUSTER_VERBOSE=1 ./main.sh


simondeziel (commented):
I agree, we should run those tests. I wonder if that [`kill 0`](https://github.com/canonical/microcluster/blob/main/example/test/main.sh#L107) wasn't meant to be a `exit 0`?

I've never seen `kill 0` used alone, only to check if a PID was present.

MggMuggins:
See [Operands](https://manpages.ubuntu.com/manpages/noble/en/man1/kill.1posix.html). The thought was to prevent the test script from creating zombie processes that had to be killed manually (ofc if something fails before the end there will still be zombies...). I think I remember `microctl shutdown` not always doing the trick; dqlite tries to hang on. I removed it in #188 and haven't had issues recently.

masnax (approved):

PR #188, opened 2024-07-18
Address changes

MggMuggins:
Allows address changes during cluster recovery.

The PR:
- Moves the rest of the recovery code into `internal/recover` instead of being awkwardly split between `microcluster/app.go` and the recover package
- Adds address updates for:
  - `daemon.yaml`
  - The trust store
  - `database/info.yaml`
  - `database/cluster.yaml`
  - The global database (via `patch.global.sql`)

I'm not a massive fan of  rewriting both `info.yaml` and `cluster.yaml` (owned by go-dqlite) but `dqlite.App` won't start otherwise (see 741e6939a3b0927fa87717f72628198092191707 for details). We also need to update `info.yaml` on every member, so I can't PR against go-dqlite unless we move most of the recovery tarball handling too.

@masnax Not sure what your review process looks like but my feeling is Tom would prefer smaller commits than these. Let me know if you'd like a more granular refactor.

### DIFF
--- a/internal/recover/recover.go
+++ b/internal/recover/recover.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"io/fs"
+	"net/netip"
 	"os"
 	"path"
 	"path/filepath"
@@ -23,9 +24,11 @@ import (
 
 	"github.com/canonical/microcluster/client"
 	"github.com/canonical/microcluster/cluster"
+	"github.com/canonical/microcluster/internal/config"
 	internalTypes "github.com/canonical/microcluster/internal/rest/types"
 	"github.com/canonical/microcluster/internal/sys"
 	"github.com/canonical/microcluster/internal/trust"
+	"github.com/canonical/microcluster/rest/types"
 )
 
 // GetDqliteClusterMembers parses the trust store and
@@ -153,12 +156,50 @@ func RecoverFromQuorumLoss(filesystem *sys.OS, members []cluster.DqliteMember) (
 	}
 
 	// Tar up the m.FileSystem.DatabaseDir and write to `dbExportPath`
-	recoveryTarballPath, err := createRecoveryTarball(filesystem)
+	recoveryTarballPath, err := createRecoveryTarball(filesystem, members)
 	if err != nil {
 		return "", err
 	}
 
+	err = updateTrustStore(filesystem.TrustDir, members)
+	if err != nil {
+		return recoveryTarballPath, fmt.Errorf("Failed to update trust store: %w", err)
+	}
+
+	err = writeGlobalMembersPatch(filesystem, members)
+	if err != nil {
+		return recoveryTarballPath, fmt.Errorf("Failed to write global DB update: %w", err)
+	}
+
 	return recoveryTarballPath, nil
+}
+
+func readYaml(path string, v any) error {
+	yml, err := os.ReadFile(path)
+	if err != nil {
+		return err
+	}
+
+	err = yaml.Unmarshal(yml, v)
+	if err != nil {
+		return fmt.Errorf("Unmarshal %q: %w", path, err)
+	}
+
+	return nil
+}
+
+func writeYaml(path string, v any) error {
+	yml, err := yaml.Marshal(v)
+	if err != nil {
+		return err
+	}
+
+	err = os.WriteFile(path, yml, os.FileMode(0o644))
+	if err != nil {
+		return err
+	}
+
+	return nil
 }
 
 // ReadTrustStore parses the trust store. This is not thread safe!
@@ -167,6 +208,40 @@ func readTrustStore(dir string) (*trust.Remotes, error) {
 	err := remotes.Load(dir)
 
 	return remotes, err
+}
+
+// Update the trust store with the new member addresses.
+func updateTrustStore(dir string, members []cluster.DqliteMember) error {
+	remotes, err := readTrustStore(dir)
+	if err != nil {
+		return err
+	}
+
+	remotesByName := remotes.RemotesByName()
+
+	trustMembers := make([]types.ClusterMember, 0, len(members))
+	for _, member := range members {
+		cert := remotesByName[member.Name].Certificate
+		addr, err := netip.ParseAddrPort(member.Address)
+		if err != nil {
+			return fmt.Errorf("Invalid address %q: %w", member.Address, err)
+		}
+
+		trustMembers = append(trustMembers, types.ClusterMember{
+			ClusterMemberLocal: types.ClusterMemberLocal{
+				Name:        member.Name,
+				Address:     types.AddrPort{AddrPort: addr},
+				Certificate: cert,
+			},
+		})
+	}
+
+	err = remotes.Replace(dir, trustMembers...)
+	if err != nil {
+		return fmt.Errorf("Update trust store at %q: %w", dir, err)
+	}
+
+	return nil
 }
 
 // ValidateMemberChanges compares two arrays of members to ensure:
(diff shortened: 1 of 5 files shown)


internal/recover/recover.go
```diff
@@ -57,28 +65,237 @@ func GetDqliteClusterMembers(filesystem *sys.OS) ([]cluster.DqliteMember, error)
...
+	}
+
+	// Ensure that the daemon is not running
+	isSocketPresent, err := filesystem.IsControlSocketPresent()
+	if err != nil {
+		return "", err
+	}
+
+	if isSocketPresent {
+		return "", fmt.Errorf("daemon is running (socket path exists: %q)", filesystem.ControlSocketPath())
```
masnax, line 90: Errors should start with an uppercase letter

internal/recover/recover.go
```diff
@@ -57,28 +65,237 @@ func GetDqliteClusterMembers(filesystem *sys.OS) ([]cluster.DqliteMember, error)
...
+	if err != nil {
+		return "", err
+	}
+
+	cancelCtx, cancel := context.WithTimeout(context.Background(), time.Second*10)
+	err = cluster.Query(cancelCtx, true, func(ctx context.Context, client *client.Client) error {
+		var rslt internalTypes.Server
+		err := client.Query(ctx, "GET", "1.0", api.NewURL(), nil, &rslt)
+		if err == nil {
+			return fmt.Errorf("contacted cluster member at %q; please shut down all cluster members", rslt.Name)
```
masnax, line 126: Error should start with an uppercase letter

internal/recover/recover.go
```diff
@@ -57,28 +65,237 @@ func GetDqliteClusterMembers(filesystem *sys.OS) ([]cluster.DqliteMember, error)
...
+	}
+
+	err = CreateDatabaseBackup(filesystem)
+	if err != nil {
+		return "", err
+	}
+
+	err = dqlite.ReconfigureMembershipExt(filesystem.DatabaseDir, nodeInfo)
+	if err != nil {
+		return "", fmt.Errorf("dqlite recovery: %w", err)
```
masnax, line 142: Errors should start with an uppercase letter

internal/recover/recover.go
```diff
@@ -144,43 +396,44 @@ func MaybeUnpackRecoveryTarball(filesystem *sys.OS) error {
...
 
 		if found {
+			localInfo.Address = incomingInfo.Address
 			break
 		}
 	}
 
 	if !found {
-		return fmt.Errorf("missing local cluster member in incoming cluster.yaml")
+		return fmt.Errorf("missing local cluster member in incoming recovery.yaml")
```
masnax, line 427: Errors should start with an uppercase letter

internal/recover/recover.go
```diff
@@ -90,10 +307,8 @@ func ValidateMemberChanges(oldMembers []cluster.DqliteMember, newMembers []clust
 	for _, newMember := range newMembers {
 		memberValid := false
 		for _, oldMember := range oldMembers {
-			// FIXME: Allow changing member addresses as part of cluster recovery
 			membersMatch := newMember.DqliteID == oldMember.DqliteID &&
-				newMember.Name == oldMember.Name &&
-				newMember.Address == oldMember.Address
+				newMember.Name == oldMember.Name
```
masnax, line 311: Should we do any checks here that addresses are unique, or are real addresses, before we proceed?

Also, should we ensure the new set of roles are valid? Like for example if someone sets all the roles to `spare`, we should error out.

masnax (changes requested):
Looking good, just a handful of nits about error formatting and a question about validation.

internal/recover/recover.go
MggMuggins: We should implement a lint check for this in lxd/microcloud/microcluster (if not already done)

internal/recover/recover.go
MggMuggins: Both good ideas; now it checks for parseable addrs and at least one voter

masnax: I thought that we needed a majority of nodes to be voters. Are there valid cases for most members not being voters?

masnax: That makes me wonder, what happens when we leave the max voter count at 3, but then make all the nodes voters with the recovery command. Do they all immediately re-adjust to 3 voters?

MggMuggins: That's what it [looks](https://github.com/canonical/go-dqlite/blob/master/app/app.go#L600) like [to me](https://github.com/canonical/go-dqlite/blob/master/app/roles.go#L189); can test if you'd like.

internal/recover/recover.go
```diff
@@ -57,43 +65,272 @@ func GetDqliteClusterMembers(filesystem *sys.OS) ([]cluster.DqliteMember, error)
...
+		if addrs[addr] {
+			return fmt.Errorf("Duplicate address %q", addr)
+		}
+
+		addrs[addr] = true
+
 		for _, oldMember := range oldMembers {
-			// FIXME: Allow changing member addresses as part of cluster recovery
 			membersMatch := newMember.DqliteID == oldMember.DqliteID &&
-				newMember.Name == oldMember.Name &&
```
masnax, line 95: Small nit here, `memberValid` only exists to be assigned the value of `membersMatch` if it's true. 

You can simplify it a bit like this:

```go
memberValid := false
for _, oldMember := range oldMembers {
  memberValid = newMember.DqliteID == oldMember.DqliteID && newMember.Name == oldMember.Name
  if memberValid {
    break
  }
}
```

internal/recover/recover.go
```diff
@@ -131,6 +408,7 @@ func CreateRecoveryTarball(filesystem *sys.OS) (string, error) {
 func MaybeUnpackRecoveryTarball(filesystem *sys.OS) error {
 	tarballPath := path.Join(filesystem.StateDir, "recovery_db.tar.gz")
 	unpackDir := path.Join(filesystem.StateDir, "recovery_db")
+	recoveryYamlPath := path.Join(unpackDir, "recovery.yaml")
```
masnax, line 411: Any reason not to remove the `recovery.yaml` when we're done?

internal/recover/recover.go
```diff
@@ -112,14 +382,21 @@ func ValidateMemberChanges(oldMembers []cluster.DqliteMember, newMembers []clust
...
 // filesystem.StateDir.
 // go-dqlite's info.yaml is excluded from the tarball.
+// The new cluster configuration is included as `recovery.yaml`.
 // This function returns the path to the tarball.
-func CreateRecoveryTarball(filesystem *sys.OS) (string, error) {
+func createRecoveryTarball(filesystem *sys.OS, members []cluster.DqliteMember) (string, error) {
 	tarballPath := path.Join(filesystem.StateDir, "recovery_db.tar.gz")
+	recoveryYamlPath := path.Join(filesystem.DatabaseDir, "recovery.yaml")
+
+	err := writeYaml(recoveryYamlPath, members)
```
masnax, line 391: Right before this write to `recovery.yaml`, we write the same information to `cluster.yaml`, so why do we need both files?

masnax (changes requested):
Couple questions about the `recovery.yaml`: Do we really need it, and if so should we remove it after we replace the database dir?

internal/recover/recover.go
MggMuggins: The schema are different:
```yaml
# cluster.yaml
- ID: 3297041220608546238
  Address: 127.0.0.1:8001
  Role: 0
...
```
```yaml
# recovery.yaml
- id: 3297041220608546238
  address: 127.0.0.1:8001
  role: voter
  name: c1
...
```
Since the incoming addresses may have changed, we have no way of knowing which truststore entry to update with it's new address unless we have a mapping of `name -> address`.

internal/recover/recover.go
MggMuggins: No :)

masnax (approved):
Looks good, thanks!

PR #189, opened 2024-07-19
internal/rest/resources: Improve join logic to handle unreachable nodes

bschimke95:
Fixes an issue where the join process would fail if a node in the join_address list is unreachable.

When a token is created, it includes the IP addresses of all nodes in the cluster. If a node is removed between the token creation and joining, the join process previously failed because it tried to connect to the non-existent node.

Now, the client discovery logic will continue attempting to connect to other nodes in the join_address list instead of failing early. This ensures that the join process can succeed as long as at least one node is reachable.

### DIFF
--- a/internal/rest/resources/control.go
+++ b/internal/rest/resources/control.go
@@ -129,12 +129,14 @@ func joinWithToken(state state.State, r *http.Request, req *internalTypes.Contro
 
 		cert, err := shared.GetRemoteCertificate(url.String(), "")
 		if err != nil {
-			return response.SmartError(fmt.Errorf("Failed to get certificate of cluster member %q: %w", url.URL.Host, err))
+			logger.Warn("Failed to get certificate of cluster member", logger.Ctx{"address": url.String(), "error": err})
+			continue
 		}
 
 		fingerprint := shared.CertFingerprint(cert)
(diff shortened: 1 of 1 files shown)


internal/rest/resources/control.go
```diff
@@ -129,12 +129,12 @@ func joinWithToken(state state.State, r *http.Request, req *internalTypes.Contro
 
 		cert, err := shared.GetRemoteCertificate(url.String(), "")
 		if err != nil {
-			return response.SmartError(fmt.Errorf("Failed to get certificate of cluster member %q: %w", url.URL.Host, err))
+			logger.Debug("Failed to get certificate of cluster member", logger.Ctx{"address": url.String(), "error": err})
```
tomponline, line 132: Should this be logger.Warn or logger.Info? Debug seems too low.

bschimke95: True, increased to `Warn`

masnax (approved):
Thanks for this!

masnax:
@bschimke95 Can you please sign your commits with your GPG or SSH key?

bschimke95:
Done

roosterfish (approved):
LGTM!

PR #190, opened 2024-07-19
Manage hook lifecycles with context

masnax:
Since we are no longer exposing `state.Context`, we need to pass a context as argument to each hook to manage its lifetime.

For most hooks, this means their parent context is the request context for the API that triggered the hook. The actual context passed to the hook will be cancelled immediately after the hook returns. This is to discourage long-lived goroutines running in these hooks.

The exception is `OnStart`. For `OnStart`, which is intended to start services that work with microcluster, the lifetime should be equivalent to the daemon itself. So the shutdown context of the daemon is supplied to this hook. Its comment has been updated to reference this.

### DIFF
--- a/example/cmd/microd/main.go
+++ b/example/cmd/microd/main.go
@@ -2,6 +2,7 @@
 package main
 
 import (
+	"context"
 	"os"
 
 	"github.com/canonical/lxd/shared/logger"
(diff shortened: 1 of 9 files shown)


roosterfish (approved):
LGTM!

PR #191, opened 2024-07-22
Doc: Clarify `apiExtensions` parameter

roosterfish:
Make sure all non self-explanatory parameters of the microcluster `Start()` func are documented properly.


masnax (approved):
Thanks!

PR #192, opened 2024-07-22
Go: Consistently use `gopkg.in/yaml.v3`

roosterfish:
In the code we are using different versions (v2 and v3) of the yaml dependency.

This should fix future automatic PRs like https://github.com/canonical/microcluster/pull/183 as `v3` was also listed as indirect dependency.

Fixes https://github.com/canonical/microcluster/pull/183.


MggMuggins (approved):
Example tests pass; good to keep things consistent. LGTM!

PR #193, opened 2024-07-22
Use daemon context for bootstrap hooks

bschimke95:
Follow-up on https://github.com/canonical/microcluster/pull/190

We require the daemon context to perform cleanup tasks which are performed in a separate Go-routine. See https://github.com/canonical/k8s-snap/blob/ede6335f40969b8d1846e528b8376b20190b5921/src/k8s/pkg/k8sd/app/hooks_bootstrap.go#L62-L91

While I understand that long-running tasks are discouraged in those hooks once the daemon is set up, I think it makes sense to have the daemons lifecycle context for setup hooks

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -463,9 +463,7 @@ func (d *Daemon) StartAPI(ctx context.Context, bootstrap bool, initConfig map[st
 	}
 
 	if bootstrap {
-		ctx, cancel := context.WithCancel(ctx)
-		err := d.hooks.PreBootstrap(ctx, d.State(), initConfig)
-		cancel()
+		err := d.hooks.PreBootstrap(d.shutdownCtx, d.State(), initConfig)
 		if err != nil {
 			return fmt.Errorf("Failed to run pre-bootstrap hook before starting the API: %w", err)
 		}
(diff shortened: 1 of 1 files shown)


masnax:
It seems the [Join hooks](https://github.com/canonical/k8s-snap/blob/ede6335f40969b8d1846e528b8376b20190b5921/src/k8s/pkg/k8sd/app/hooks_join.go#L53) also use goroutines like this.

Looking at the cleanup actions performed by these goroutines, I think they mostly clean up files and directories, except for calls to [stopControlPlaneServices](https://github.com/canonical/k8s-snap/blob/73358d8c77e6efc721726ef57958e54bca8829c3/src/k8s/pkg/k8sd/app/cluster_util.go#L54-L70) and [ResetClusterMember](https://github.com/canonical/k8s-snap/blob/ede6335f40969b8d1846e528b8376b20190b5921/src/k8s/pkg/k8sd/app/hooks_bootstrap.go#L87), which I am assuming are the two bottlenecks that necessitate the goroutines. 

`ResetClusterMember` in particular uses an internal microcluster API endpoint that was not intended be accessed directly. It's used internally to re-exec the daemon process when removing a node from the cluster, and also when a node fails the cluster join process. It currently doesn't run during a failure when bootstrapping the first node, but that was an oversight on my part as we should be cleaning up our internal state if we encounter an error so that the bootstrap can be re-attempted.

Since we execute `ResetClusterMember` internally, it means that spawning goroutines in the these hooks actually racy and may behave inconsistently because microcluster may re-exec the daemon process before the goroutine has finished. I think this is actually a good example of why it can be problematic to allow long-lived goroutines in hooks that fire in the middle of internal actions. 

So if you were to drop `ResetClusterMember` from those hooks, would that sufficiently speed up the cleanup that it could be synchronous?

roosterfish:
@bschimke95 can this one be closed?

bschimke95:
Yes, obsolete.
Resolved offline

PR #194, opened 2024-07-22
internal/rest/resources: Reset the daemon if bootstrapping fails

masnax:
We already do this cleanup when a node fails to join the cluster, but we should also try to reset the daemon if bootstrapping fails, so that the node can be easily re-initialized.

### DIFF
--- a/internal/rest/resources/control.go
+++ b/internal/rest/resources/control.go
@@ -91,6 +91,14 @@ func controlPost(state state.State, r *http.Request) response.Response {
 
 		// Re-exec the daemon to clear any remaining state.
 		go reExec()
+
+		// Run the pre-remove hook like we do for cluster node removals.
+		err = intState.Hooks.PreRemove(r.Context(), state, true)
+		if err != nil {
+			logger.Error("Failed to run pre-remove hook on bootstrap error", logger.Ctx{"error": err})
+
+			return
+		}
 	})
 
 	daemonConfig := &trust.Location{Address: req.Address, Name: req.Name}
--- a/internal/rest/resources/control.go
+++ b/internal/rest/resources/control.go
@@ -75,11 +75,25 @@ func controlPost(state state.State, r *http.Request) response.Response {
 		return joinWithToken(state, r, req)
 	}
 
+	reverter := revert.New()
+	defer reverter.Fail()
+	reverter.Add(func() {
+		reExec, err := resetClusterMember(r.Context(), state, true)
+		if err != nil {
+			return
+		}
+
+		// Re-exec the daemon to clear any remaining state.
+		go reExec()
+	})
+
 	daemonConfig := &trust.Location{Address: req.Address, Name: req.Name}
 	err = intState.StartAPI(r.Context(), req.Bootstrap, req.InitConfig, daemonConfig)
 	if err != nil {
 		return response.SmartError(err)
 	}
+
+	reverter.Success()
 
 	return response.EmptySyncResponse
 }
--- a/internal/rest/resources/control.go
+++ b/internal/rest/resources/control.go
@@ -15,6 +15,7 @@ import (
 	"github.com/canonical/lxd/shared/revert"
 	"github.com/canonical/lxd/shared/validate"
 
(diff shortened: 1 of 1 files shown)


internal/rest/resources/control.go
```diff
@@ -75,12 +75,26 @@ func controlPost(state state.State, r *http.Request) response.Response {
 		return joinWithToken(state, r, req)
 	}
 
+	reverter := revert.New()
```
roosterfish, line 84: Do you set the reverter before calling `StartAPI()` so that if the call fails the revert gets executed?

To keep the common pattern of first performing an action and then registering the reverter afterwards, I am wondering if the reverter would be better placed inside `StartAPI()` like it's done in `joinWithToken()`?

internal/rest/resources/control.go
```diff
@@ -75,12 +75,26 @@ func controlPost(state state.State, r *http.Request) response.Response {
...
+	defer reverter.Fail()
+	reverter.Add(func() {
+		reExec, err := resetClusterMember(r.Context(), state, true)
+		if err != nil {
+			return
+		}
+
+		// Re-exec the daemon to clear any remaining state.
+		go reExec()
```
roosterfish, line 93: The revert in `joinWithToken()` also tries to delete the cluster member from the DB after doing the `go reExec()`. Do we need this step here too? 

```go
err = client.DeleteClusterMember(context.Background(), req.Name, true)
```

roosterfish (changes requested):

internal/rest/resources/control.go
masnax: `StartAPI` is also called when the daemon restarts, where a failure doesn't require us to reset the whole daemon. It's only in the bootstrap and join cases where we have to reset the daemon back to its initial state.

internal/rest/resources/control.go
masnax: In the join case, we also have to clear the state of other nodes.

roosterfish: Ok thanks makes sense.

internal/rest/resources/control.go
roosterfish: In this case would it make sense to return a `cleanup` hook from `StartAPI()`? So the caller can decide whether or not to run the revert.

masnax: I'm not sure if that would make things unclear/dangerous. The re-exec logic does not "belong" to `StartAPI`, which may eventually be used in more  situations and require situational cleanup. We don't necessarily want to elicit the question of "do you want to erase all data in the state directory and forcibly kill the daemon" for each call to `StartAPI`.

I think this revert is in line with the pattern you mentioned, it is tied to the control API endpoint used for bootstrap/join. The "action" in this case is handling that API endpoint.

masnax:
You can ignore what I had said about the tokens API in the meeting, it appears at some point I had removed that endpoint I was referring to :)

internal/rest/resources/control.go
roosterfish: Thinking a bit more about it now I agree that we shouldn't make it more complex. 

It's just that you register the revert (`resetClusterMember()`) before the actual actions have been taken as part of `StartAPI()`.
Is `resetClusterMember()` idempotent in a way that it doesn't matter how far into `StartAPI()` an error is returned?

roosterfish (approved):
Looks good!

internal/rest/resources/control.go
masnax: > Is `resetClusterMember()` idempotent in a way that it doesn't matter how far into `StartAPI()` an error is returned?

I'm not fully following you, can you give an example of what you mean here?

`resetClusterMember` attempts to stop the database and delete some files in case they were created. We run it with `force=true` so it will always try to fully clean up and ignore errors. As such I don't think the timing with `StartAPI` is of much concern.

roosterfish: Yeah that's what I meant. I missed the fact that we have a `force` flag for this func which is set to `true` in this case.
Fine for me then.

PR #195, opened 2024-07-23
Harmonize public facing endpoints

roosterfish:
This PR ensures that all the API endpoints which are used by public facing client functions are also listed under the `/core/1.0` path instead of `/core/internal`.

In addition the PR does the actual rename of `/cluster/(1.0|internal)` to `/core/(1.0|internal)` as https://github.com/canonical/microcluster/pull/137 looks to have modified unused constants. Looks we somehow had duplicate constants in the code base.
I have also moved the public facing `types.ExtendedEndpoint` that represents the `/1.0` prefix into the top level `/rest/types` package.

### DIFF
--- a/internal/db/dqlite.go
+++ b/internal/db/dqlite.go
@@ -28,7 +28,7 @@ import (
 	"github.com/canonical/microcluster/cluster"
 	"github.com/canonical/microcluster/internal/db/update"
 	"github.com/canonical/microcluster/internal/extensions"
-	"github.com/canonical/microcluster/internal/rest/client"
+	internalClient "github.com/canonical/microcluster/internal/rest/client"
 	internalTypes "github.com/canonical/microcluster/internal/rest/types"
 	"github.com/canonical/microcluster/internal/sys"
 	"github.com/canonical/microcluster/rest/types"
@@ -310,14 +310,14 @@ func (db *DB) heartbeat(ctx context.Context) {
 	db.heartbeatLock.Lock()
 	defer db.heartbeatLock.Unlock()
 
-	client, err := client.New(db.os.ControlSocket(), nil, nil, false)
+	client, err := internalClient.New(db.os.ControlSocket(), nil, nil, false)
 	if err != nil {
 		logger.Error("Failed to get local client", logger.Ctx{"address": db.listenAddr.String(), "error": err})
 		return
 	}
 
 	// Initiate a heartbeat from this node.
-	err = client.Heartbeat(ctx, internalTypes.HeartbeatInfo{BeginRound: true})
+	err = internalClient.Heartbeat(ctx, client, internalTypes.HeartbeatInfo{BeginRound: true})
 	if err != nil && err.Error() != "Attempt to initiate heartbeat from non-leader" {
 		logger.Error("Failed to initiate heartbeat round", logger.Ctx{"address": db.dqlite.Address(), "error": err})
 		return
@@ -331,7 +331,7 @@ func dqliteNetworkDial(ctx context.Context, addr string, db *DB) (net.Conn, erro
 		return nil, err
 	}
 
-	config, err := client.TLSClientConfig(db.serverCert, peerCert)
+	config, err := internalClient.TLSClientConfig(db.serverCert, peerCert)
 	if err != nil {
 		return nil, fmt.Errorf("Failed to parse TLS config: %w", err)
 	}
--- a/internal/rest/client/client.go
+++ b/internal/rest/client/client.go
@@ -26,23 +26,6 @@ import (
 	"github.com/canonical/microcluster/rest/types"
 )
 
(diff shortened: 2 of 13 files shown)


masnax:
Thanks for catching the duplicate endpoint types, but I'm not sure I agree with putting all these endpoints on `/1.0`, as a lot of these are not supposed to be accessed directly. Could we instead remove the receiver from the client functions so that they cannot be used by other projects?

masnax:
Looking at this again, I think it's all fine except for two points:

1. the `SQL` endpoint I think should remain internal because we don't really want to encourage usage of it as part of the API, that one is mainly offered as an app-level tool to integrate into CLIs.

2. The `daemonCmd` endpoint has a method on `microcluster.App` that seems a bit redundant now, should that be removed?

masnax:
Same with the cluster certificate endpoint's corresponding helper on `microcluster.App`, should that be removed as well?

masnax:
You can ignore what I said about the tokens API in the meeting, it seems I removed that endpoint I was referring to :)

roosterfish:
> Could we instead remove the receiver from the client functions so that they cannot be used by other projects?

Yeah good point. So I think we should separate the endpoints the following way:

* `/core/1.0`
  * api10Cmd
  * clusterCertificatesCmd
  * daemonCmd
  * tokenCmd
  * tokensCmd
  * readyCmd
* `/core/internal`
  * clusterCmd
  * clusterMemberCmd
  * databaseCmd
  * sqlCmd
  * heartbeatCmd
  * trustCmd
  * trustEntryCmd
  * hooksCmd

This means moving `clusterCmd` and `clusterMemberCmd` to `/core/interal` and moving `clusterCertificatesCmd`, `tokenCmd` and `daemonCmd` to `/core/1.0`. 

If someone uses any of the `/core/1.0` endpoints directly we ensure they can all be used as standalone. Whereas the `/core/internal` ones should not be used directly. 
Then from the `Client` type we remove all the ones that should not be called outside of microcluster.

To delete cluster members I would suggest to add another function `MicroCluster.ShrinkCluster(ctx, name)` which allows removing a member from the cluster? This way we don't have to export the `Client.DeleteClusterMember()` func.

To list cluster members we can either keep the `Client.GetClusterMembers()` or also add a function `MicroCluster.ClusterMembers()`.

masnax:
Agreed on all points, though getting rid of `DeleteClusterMember` will probably mean all projects are going to have to perform a somewhat non-trivial refactor.

masnax:
Maybe for simplicity's sake (and also to keep things similar between the `tokens` and `cluster` endpoint which are highly coupled, we can keep `GET `/core/1.0/cluster` and `DELETE `/core/1.0/cluster/{name}` on the public API with client functions.

roosterfish:
I was thinking about this and came up with the following to keep things consistent and together:

* All the cluster member endpoints are now under `/core/internal`. This discourages external use of those endpoints but the `Client` type still has the two `GetClusterMembers` and `DeleteClusterMember` funcs.
This allows us introducing changes in all of the `/core/internal` endpoints on our behalf whilst keeping the `Client` interface stable (we can decide freely how to call the `/core/internal` endpoints under the hood as long as we keep the client funcs stable).
* All the endpoints under `/core/1.0` can be used externally and also still have their func on `Client`.
* All the other endpoints under `/core/internal` got their `Client` func revoked
* All the control endpoints still have their client func as we can again freely modify the endpoints itself (as they are internal) whilst maintaining a stable client interface

roosterfish:
Going into the future we can say `/core/1.0` is kept stable as well as the `Client` interface. 

This should also not cause any refactor on the downstream projects as the two cluster member funcs (list, delete) stay in place.

roosterfish:
@masnax update based on our discussion.

`/core/1.0` now contains the endpoints behind `GetClusterMembers` and `DeleteClusterMembers` whereas `AddClusterMember` and `ResetClusterMember` aren't anymore exposed to the `Client` struct and internally call endpoints on `/core/internal`.

Additionally the `UpdateServers` and `UpdateCertificate` funcs are removed from the app as those where just simple wrappers around the client funcs.

rest/types/endpoint.go
```diff
@@ -2,3 +2,8 @@ package types
 
 // EndpointPrefix is a type specifying the endpoint on which the resource exists.
 type EndpointPrefix string
+
+const (
+	// ExtendedEndpoint - All endpoints added through additional servers which are part of the core API.
+	ExtendedEndpoint EndpointPrefix = "1.0"
```
masnax, line 8: This doesn't need to exist anymore since we allow projects to specify their own prefixes for the core API, as long as they don't conflict with one of the existing ones.

roosterfish: Fair point, removed.

masnax (approved):
lgtm, thanks!

PR #196, opened 2024-07-23
example: Fix bootstrap failure due to unregistered extension

bschimke95:
The example code in `example/README.md` fails to bootstrap with the error: 
"The 'missing_extension' is not registered."
To address this, the example state access code has been commented out to make the example functional again while keeping the code for reference.


roosterfish:
The code in question explicitly checks for the presence of the non existing extension `missing_extension`. This is just logging a warning message and should not lead to any failures during bootstrap?

bschimke95:
Ah, my bad. I was just reading the failure but did not see that it is just a warning.

PR #197, opened 2024-07-24
Bootstrap: Run the PreRemove hook before re-exec

roosterfish:
This allows accessing the database from within the hook before
closing the database and re-executing the daemon.

### DIFF
--- a/internal/rest/resources/control.go
+++ b/internal/rest/resources/control.go
@@ -92,6 +92,7 @@ func controlPost(state state.State, r *http.Request) response.Response {
 
 		reExec, err := resetClusterMember(r.Context(), state, true)
 		if err != nil {
+			logger.Error("Failed to reset cluster member on bootstrap error", logger.Ctx{"error": err})
 			return
 		}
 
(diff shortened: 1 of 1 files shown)


internal/rest/resources/control.go
```diff
@@ -84,21 +84,21 @@ func controlPost(state state.State, r *http.Request) response.Response {
 	reverter := revert.New()
 	defer reverter.Fail()
 	reverter.Add(func() {
-		reExec, err := resetClusterMember(r.Context(), state, true)
+		// Run the pre-remove hook like we do for cluster node removals.
+		err := intState.Hooks.PreRemove(r.Context(), state, true)
 		if err != nil {
+			logger.Error("Failed to run pre-remove hook on bootstrap error", logger.Ctx{"error": err})
 			return
```
roosterfish, line 91: @masnax Should an error in the hook cancel the revert or should we also remove the `return` statement? I would vote for removing it.

bschimke95: my 2ct: If something fails we need to call `resetClusterMember` otherwise a follow-up bootstrap try will fail.

Cleanups should always be on a best-effort basis. If something fails, move on to the next cleanup step.

masnax (approved):

PR #200, opened 2024-07-24
Join: Run the PreRemove hook on join failures

bschimke95:
Execute the PreRemove hook on the node that tried to join. This aligns a failed join with the behavior of a failed bootstrap. With that, PreRemove can generally be used to add custom cleanup logic.


masnax:
Ah, I didn't see this and made my own PR. Whoops! Thanks for catching this :)

masnax (approved):

Issue #199, opened 2024-07-24
Add join token expiration


masnax:
Join tokens exist forever after creation. Instead, we should have some mechanism to expire tokens after a user-specified period of time.

This means updating the join token configuration [here](https://github.com/canonical/microcluster/blob/e7fce6564cbe708e198a6e644696dc5ab7f72b2c/microcluster/app.go#L277-L289) to take a timeout. 

There are a few nuances here like changing the schema to expect an initial creation date for tokens.

masnax:
I'm not sure how LXD handles image expiry, but I'm thinking in microcluster a good place to handle this would be when issuing a heartbeat.

masnax:
There's also the matter of deciding if we want a sensible default expiration, like a minute per token by default.

MggMuggins:
https://documentation.ubuntu.com/lxd/en/latest/server/#server-core:core.remote_token_expiry

PR #202, opened 2024-07-25
Add expiry date column to `core_token_records`

masnax:
CC @MggMuggins @boltmark 

This adds the necessary schema/client/api type changes necessary for join token expiration, following a similar pattern to LXD's expiration date handling. 

As of now, the field is unused when specified to the API, so the remaining steps would be:
* On `POST` to `/core/internal/tokens`, if `req.ExpireAfter` is non-zero, set the expiration of the token to that duration after the current time, when adding the token record to the database.
* On each heartbeat handled by the leader, check all token expirations and delete any expired ones. There's lots of transactions here that can be re-used.
* Update example CLI to take a flag for setting the expiration, this can default to 3 hours like LXD's join tokens.
* Update example CLI to show the expiration date when listing tokens.

### DIFF
--- a/cluster/token_records.go
+++ b/cluster/token_records.go
@@ -57,7 +57,8 @@ func (t *CoreTokenRecord) ToAPI(clusterCert *x509.Certificate, joinAddresses []t
 	}
 
 	return &internalTypes.TokenRecord{
-		Token: tokenString,
-		Name:  t.Name,
+		Token:     tokenString,
+		Name:      t.Name,
+		ExpiresAt: t.ExpiryDate.Time,
 	}, nil
 }
--- a/cluster/token_records.go
+++ b/cluster/token_records.go
@@ -2,6 +2,7 @@ package cluster
 
 import (
 	"crypto/x509"
+	"database/sql"
 
 	"github.com/canonical/lxd/shared"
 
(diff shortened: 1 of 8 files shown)


roosterfish (dismissed):
LGTM!

roosterfish:
There is a conflict now.

roosterfish (approved):

MggMuggins (approved):
Looks great.

Regarding the implementation, do we really need to delete expired tokens every heartbeat? Since we'll behave the same way regardless of whether an expired record exists in the table or not, it makes more sense to me to only update the table when we're also touching it for some other reason. (If only to keep the token code isolated to the token-relevant files; performance not considered)

EDIT: #199 for ref

masnax:
> Looks great.
> 
> Regarding the implementation, do we really need to delete expired tokens every heartbeat? Since we'll behave the same way regardless of whether an expired record exists in the table or not, it makes more sense to me to only update the table when we're also touching it for some other reason. (If only to keep the token code isolated to the token-relevant files; performance not considered)
> 
> EDIT: #199 for ref

Handling token expiry in the token API handlers would keep the token code localized to the token files but it does introduce some ambiguity, like would a `GET` to fetch tokens also delete expired ones? That would be a bit odd. It also means the expired token cleanup would happen in more than one place, and would be tied to user actions. 

I'm assuming we'll have other components that we will want to have expire so localizing all database cleanup to heartbeats might make things cleaner in the long run. We also have the `OnHeartbeat` hook so downstream projects can track this behaviour.

MggMuggins:
That's reasonable; I'll stick it with the heartbeat.

PR #203, opened 2024-07-25
internal/rest/resources: Don't allow token issuing over the network

masnax:
This makes it so that simply trusting the cluster is not enough to issue a join token, it has to be issued from an actual cluster member.

### DIFF
--- a/internal/rest/client/tokens.go
+++ b/internal/rest/client/tokens.go
@@ -16,7 +16,7 @@ func (c *Client) RequestToken(ctx context.Context, name string) (string, error) 
 
 	var token string
 	tokenRecord := types.TokenRecord{Name: name}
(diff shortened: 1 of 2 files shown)


roosterfish (approved):
Looks reasonable!

In case of MicroCloud when trying to reuse an existing service, I guess we can use the MicroCloud proxy to still allow issuing tokens remotely?

masnax:
> Looks reasonable!
> 
> In case of MicroCloud when trying to reuse an existing service, I guess we can use the MicroCloud proxy to still allow issuing tokens remotely?

Actually, I was under the impression that microcluster _already_ behaved this way, so yes MicroCloud always used the proxy to forward the network request to the unix socket of the corresponding service.

PR #205, opened 2024-07-26
Join token expiration

### LINKED ISSUE #199: Add join token expiration

MggMuggins:
Fixes #199.

I got sick of feeling like I couldn't test effectively with the system tests, so this includes a refactor to separate them out a bit. Should make reviewing changes easier in future. They're far from perfect but still better than they were. I can split them into a different PR if that would be preferred.

Regarding tokens:
- Don't return an expired token from `GET tokens/`
- Insert an expiry date if one is provided during token creation
- Delete expired tokens on heartbeat
- Token support in `microctl`

LXD-1349

### DIFF
--- a/example/test/main.sh
+++ b/example/test/main.sh
@@ -93,6 +93,38 @@ test_misc() {
   shutdown_systems
 }
 
+test_tokens() {
+  new_systems 3
+  bootstrap_systems
+
+  microctl --state-dir "${test_dir}/c1" tokens add default_expiry
+
+  microctl --state-dir "${test_dir}/c1" tokens add short_expiry --expire-after 5s
+
+  microctl --state-dir "${test_dir}/c1" tokens add long_expiry --expire-after 400h
+
+  # Need at least one heartbeat (every 10s)
+  sleep 15
+
+  ! microctl --state-dir "${test_dir}/c1" tokens list --format csv | grep -q short_expiry || false
+  microctl --state-dir "${test_dir}/c1" tokens list --format csv | grep -q default_expiry
+  microctl --state-dir "${test_dir}/c1" tokens list --format csv | grep -q long_expiry
+
+  # Ensure expired tokens cannot be used to join the cluster
+  mkdir -p "${test_dir}/c4"
+  microd --state-dir "${test_dir}/c4" "${cluster_flags[@]}" &
+  microctl --state-dir "${test_dir}/c4" waitready
+
+  # Assumes that heartbeats are 10s apart
+  token=$(microctl --state-dir "${test_dir}/c1" tokens add "c4" --expire-after 12s)
+
+  sleep 12
+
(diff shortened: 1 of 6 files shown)


internal/rest/resources/heartbeat.go
```diff
@@ -259,3 +259,21 @@ func beginHeartbeat(ctx context.Context, s state.State, hbReq internalTypes.Hear
...
 }
+
+func deleteExpiredTokens(ctx context.Context, tx *sql.Tx) error {
+	tokens, err := cluster.GetCoreTokenRecords(ctx, tx)
+	if err != nil {
+		return err
+	}
+
+	for _, token := range tokens {
+		if token.ExpiryDate.Valid && token.ExpiryDate.Time.Compare(time.Now()) > 0 {
```
masnax: Can you use `token.Expired()` here?

masnax (changes requested):
Looks good, there are also 2 more cases to consider:

* Check that a token is not expired at the time we receive it from a joining member.

* If someone tries to create a token with an expired token's name in between it expiring and the next heartbeat, a database record will still exist so the creation will error out with a conflict. We should remove the old record and create a new one instead of failing, especially since a `GET` request will omit expired tokens, which would appear confusing.

MggMuggins:
I hadn't marked it ready for review yet :)

This was one of the cases I was concerned about regarding removing expired tokens on heartbeat instead of whenever we touch the tokens table. It makes more sense to me to clean up all the expired tokens when adding a new token rather than trying to special-case it.

masnax:
I think this doesn't have to be either-or. We can use that helper you added to remove tokens both in the `POST` handler for tokens as well as on heartbeats. I still think we should make sure we don't leave noise in the database that won't be cleaned up until the next time someone wants to grow the cluster.

masnax (approved):
LGTM, thanks!

example/test/main.sh
```diff
@@ -16,100 +16,167 @@ fi
...
+  ! microctl --state-dir "${test_dir}/c2" init "c/2" 127.0.0.1:9002 --token "${token_node2}" || false
+
+  shutdown_systems
+}
+
+test_tokens() {
+  new_systems 3
+  bootstrap_systems
+
+  microctl --state-dir "${test_dir}/c1" tokens add default_expiry
```
roosterfish, line 100: Whilst reviewing this part I was wondering if we should add a comment to the microcluster `NewJoinToken()` func telling that by default there is no expiry?

roosterfish (approved):
LGTM, only one comment that can be addressed in a follow up PR.

PR #206, opened 2024-07-29
Extension servers: Validate name to be FQDN

roosterfish:
If any of the extension servers has `DedicatedCertificate` set to `true`, we should take the name of the server instead of the microcluster member name. This way when the certificate is distributed across the cluster it contains the name of the extension server not the one from the member that has initially created the certificate.

When doing so we have to validate if the extension server name is a valid FQDN.
This is a follow up on https://github.com/canonical/microcluster/pull/204#discussion_r1692719572.


masnax (dismissed):
Sounds reasonable, thanks!

roosterfish:
Rebased with the arg changes.

masnax (approved):

PR #207, opened 2024-07-29
Refactor `microcluster.Args` -> `microcluster.DaemonArgs`

MggMuggins:
This performs the refactor discussed in https://github.com/canonical/microcluster/pull/165#discussion_r1664514341.

LXD-1347

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -48,6 +48,9 @@ type Args struct {
 	// Consumers of MicroCluster are required to provide a version to serve at /cluster/1.0.
 	Version string
 
+	// Name of the Unix group of the control socket
+	SocketGroup string
+
 	// Address/port to offer the core API and extension servers over before initializing the daemon
 	PreInitListenAddress string
 
@@ -139,7 +142,7 @@ func NewDaemon(project string) *Daemon {
 
 // Run initializes the Daemon with the given configuration, starts the database,
 // and blocks until the daemon is cancelled.
-func (d *Daemon) Run(ctx context.Context, stateDir string, socketGroup string, args Args) error {
+func (d *Daemon) Run(ctx context.Context, stateDir string, args Args) error {
 	d.shutdownCtx, d.shutdownCancel = context.WithCancel(ctx)
 	if stateDir == "" {
 		stateDir = os.Getenv(sys.StateDir)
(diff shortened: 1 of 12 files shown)


internal/daemon/daemon.go
```diff
@@ -40,6 +40,30 @@ import (
...
+// Args are the data needed to start a MicroCluster daemon.
+type Args struct {
+	Verbose bool
+	Debug   bool
+
+	// Consumers of MicroCluster are required to provide a version to serve at /cluster/1.0.
+	Version string
+
+	PreInitListenAddress string
+	HeartbeatInterval    time.Duration
```
masnax, line 52: Can these have comments too?

`PreInitListenAddress` offers the core API and extension servers with `PreInit=true` over the given address/port prior to initializing the daemon. 

`HeaetbeatInterval` configures how often we attempt to send a heartbeat.

internal/daemon/daemon.go
```diff
@@ -111,11 +134,9 @@ func NewDaemon(project string, version string) *Daemon {
...
 }
 
-// Run initializes the Daemon with the given configuration, starts the database, and blocks until the daemon is cancelled.
-// - `extensionsSchema` is a list of schema updates in the order that they should be applied.
-// - `extensionServers` is a list of rest.Server that will be initialized and managed by microcluster.
-// - `hooks` are a set of functions that trigger at certain points during cluster communication.
-func (d *Daemon) Run(ctx context.Context, listenAddress string, heartbeatInterval time.Duration, stateDir string, socketGroup string, extensionsSchema []schema.Update, apiExtensions []string, extensionServers map[string]rest.Server, hooks *state.Hooks) error {
+// Run initializes the Daemon with the given configuration, starts the database,
+// and blocks until the daemon is cancelled.
+func (d *Daemon) Run(ctx context.Context, stateDir string, socketGroup string, args Args) error {
```
masnax, line 139: How come `SocketGroup` isn't a daemon arg?

MggMuggins: Because we use it to initialize the default OS [here](https://github.com/canonical/microcluster/blob/main/microcluster/app.go#L68). Looking a little closer, I'm not convinced that we should be exposing `SocketGroup` via the OS struct at all; it's not relevant to the permissions of anything other than the socket.

I agree; my vote is we remove `SocketGroup` from `sys.OS` and include it in `DaemonArgs`. For now I'll pass it straight through to `startUnixServer`.

MggMuggins:
I'm getting the test failures on `main` too; I'm hoping to reduce the length of the sleeps in the test suite and it seems like cranking up the heartbeat interval makes this happen more often.

masnax (approved):
Thanks @MggMuggins!

PR #208, opened 2024-07-29
internal/recover: Write the global DB patch on all members

MggMuggins:
Follow-up on the test failures in https://github.com/canonical/microcluster/pull/207. This is almost surely the same race condition I [was seeing in LXD](https://github.com/canonical/lxd/pull/13754#issuecomment-2231001761).

TL;DR we're [selecting](https://github.com/canonical/microcluster/blob/main/cluster/cluster_members.go#L168) on the member's new address before `core_cluster_members` has been updated to include the new addresses. I haven't traced through this as finely as I did for LXD, but like over there, this is likely to be non-trivial/breaking to fix "correctly". Since the query is idempotent, creating the patch file on all nodes is the straightforward fix.

### DIFF
--- a/internal/recover/recover.go
+++ b/internal/recover/recover.go
@@ -496,6 +496,11 @@ func MaybeUnpackRecoveryTarball(filesystem *sys.OS) error {
 		return err
 	}
 
+	err = writeGlobalMembersPatch(filesystem, incomingMembers)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


masnax:
Looking at that query, I think we could pass the daemon name to the `DB` struct and use it as the query filter to avoid dealing with the address entirely.

PR #211, opened 2024-07-30
Query migrations by member name instead of address

MggMuggins:
Alternative to #208


internal/db/db.go
```diff
@@ -25,7 +25,7 @@ import (
 
 // Open opens the dqlite database and loads the schema.
 // Returns true if we need to wait for other nodes to catch up to our version.
-func (db *DB) Open(ext extensions.Extensions, bootstrap bool, project string) error {
+func (db *DB) Open(ext extensions.Extensions, bootstrap bool, project string, memberName string) error {
```
masnax, line 28: Instead of specifying the name here, can we pass the `daemon.Name` function to `NewDB` like we do for the certificates? That way all the daemon configuration is passed in at once.

masnax (changes requested):
Just one nit about the function signature.

cluster/cluster_members.go
```diff
@@ -159,14 +159,14 @@ END;
 
 // UpdateClusterMemberSchemaVersion sets the schema version for the cluster member with the given address.
 // This helper is non-generated to work before generated statements are loaded, as we update the schema.
-func UpdateClusterMemberSchemaVersion(ctx context.Context, tx *sql.Tx, internalVersion uint64, externalVersion uint64, address string) error {
+func UpdateClusterMemberSchemaVersion(ctx context.Context, tx *sql.Tx, internalVersion uint64, externalVersion uint64, memberName string) error {
```
MggMuggins, line 162: I didn't like this solution partially for this reason; if any downstream users are using this function they won't be alerted to the change of semantics since the type doesn't change. That said neither Ceph nor OVN use this function.

masnax: These functions shouldn't be used externally anyway. After this one gets merged I'm planning to send up 1 more to lock down the database a bit with an interface.

internal/db/db.go
MggMuggins: I started out with the name itself as a field of `DB`; passing the function is a far more elegant solution...

masnax (approved):
Thanks!

PR #214, opened 2024-07-31
gomod: Update Go module to v2

masnax:


roosterfish (approved):
LGTM!

roosterfish:
The `Example system tests` job was timing out after 5+h so I have restarted it.

roosterfish:
Found the problem, the PR is missing an update in the `example/Makefile` as `LDFLAGS` still points to the old path:

`LDFLAGS="-X github.com/canonical/microcluster/example/version.version=$(VERSION)"`

That is currently causing the pipeline to fail with `Version was missing at MicroCluster daemon start`. Interestingly it doesn't cause a failure.

MggMuggins:
I'll toss a timeout in there as is done in the LXD tests.

MggMuggins (approved):

PR #217, opened 2024-07-31
example/test: Reduce total system test run time

MggMuggins:
Take advantage of configurable heartbeat intervals to reduce the CI duration by a little less than 2 minutes.

### DIFF
--- a/example/cmd/microd/main.go
+++ b/example/cmd/microd/main.go
@@ -4,6 +4,7 @@ package main
 import (
 	"context"
 	"os"
+	"time"
 
 	"github.com/canonical/lxd/shared/logger"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


example/cmd/microd/main.go
```diff
@@ -205,6 +209,8 @@ func main() {
 	app.PersistentFlags().StringVar(&daemonCmd.flagStateDir, "state-dir", "", "Path to store state information"+"``")
 	app.PersistentFlags().StringVar(&daemonCmd.flagSocketGroup, "socket-group", "", "Group to set socket's group ownership to")
 
+	app.PersistentFlags().DurationVar(&daemonCmd.flagHeartbeatInterval, "heartbeat", time.Second*10, "Group to set socket's group ownership to")
```
roosterfish: The description needs and update.

roosterfish (changes requested):
Looks good to me, only the CLI command desc requires an update.

roosterfish (approved):

PR #218, opened 2024-08-01
GitHub: Update dependabot and renovate configs

roosterfish:
This ensures that dependabot is monitoring both the `v2` and `v3` branches and that renovate is using the same config as in LXD. 

In addition `:gitSignOff` is added to the renovate `extends` directive to make sure commits made by renovate also contain the `Signed-off-by:` line.
Furthermore `baseBranches` now point to both `v2` and `v3` to receive updates for both of them.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,9 +5,25 @@ updates:
     labels: []
     schedule:
       interval: "weekly"
+    target-branch: "v3"
 
   - package-ecosystem: "gomod"
     directory: "/"
     labels: []
     schedule:
       interval: "weekly"
+    target-branch: "v3"
+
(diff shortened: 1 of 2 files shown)


roosterfish:
@masnax @simondeziel From the past I understood renovate is enabled by default for the repo, but we could potentially add this to the `renovate.json` to disable it:

```json
{
  "enabled": false
}
```

I don't see why we should keep/maintain both (also see https://docs.renovatebot.com/bot-comparison/).

simondeziel:
@roosterfish yeah, let's go with disabling renovate as it's enabled org-wide.

simondeziel (approved):

PR #219, opened 2024-08-02
Parse go module version out of project name

masnax:
This helper finds microcluster among the cached build files to register internal statements. Since microcluster's module path has been appended with v2, this function now resolves to `v2` as the project name instead of `microcluster`, so this commit properly checks if the path is of the form v[number] and if so, looks one directory back instead.

This should fix external projects failing to register internal microcluster prepared statements.

### DIFF
--- a/cluster/stmt.go
+++ b/cluster/stmt.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"regexp"
 	"runtime"
 	"strings"
 
@@ -107,8 +108,15 @@ func GetCallerProject() string {
 
 	// If not a snap build path, the project may be in a go module path of the form .../project@version....
 	before, _, ok := strings.Cut(file, "@")
-	if ok && filepath.Base(before) != "" {
-		return filepath.Base(before)
+	base := filepath.Base(before)
+	if ok && base != "" {
+		// If the base path is a go module version like v2, the project name will be one level down.
+		exp := regexp.MustCompile(`^v\d+$`)
+		if exp.MatchString(base) {
+			return filepath.Base(filepath.Dir(before))
+		}
+
+		return base
 	}
 
 	// If not a go module path,	assume a GOPATH of the form example.com/author/project/packages....


masnax:
This will need to get backported to `v2` and we'll have to tag `v2.0.1` as well.

masnax:
Wasn't caught by the example package because it's located in the microcluster package itself. Therefore, the path resolves to `GOPATH/src/...` instead of `GOPATH/pkg/mod/...`, and thus avoids the version path entirely.

cluster/stmt.go
```diff
@@ -107,8 +108,15 @@ func GetCallerProject() string {
 
 	// If not a snap build path, the project may be in a go module path of the form .../project@version....
 	before, _, ok := strings.Cut(file, "@")
-	if ok && filepath.Base(before) != "" {
-		return filepath.Base(before)
+	base := filepath.Base(before)
```
roosterfish, line 111: I am trying to come up with a scenario where you would actually have different callers/paths in the same program?
Wouldn't the return value of `GetCallerProject()` always be identical during runtime? 

If that is the case this func feels overly complex given we could just return a string which isn't `microcluster` for the `stmtsByProject` map.

roosterfish: Or is it required for the case of MicroCloud where we import the Micro* projects and each of them calls microcluster independently? 

Why not just choosing a random UUID during startup of the microcluster daemon and using this instead?

roosterfish (changes requested):
Wasn't aware of this concept until now but maybe we can make it more bulletproof by using an UUID for each daemon instead.

cluster/stmt.go
masnax: I think moving away from this is a potential idea for v3 but I don't want to worry about unintended consequences of moving in a different direction for the LTS.

If I remember right, the goal of this is to be able to distinguish which registered statement came from which project. So we need to still parse the caller function of `RegisterStmt` to determine which package it belongs to. It may be called from microcluster, from a package importing microcluster (like MicroCloud), or via another microcluster project (like MicroCloud importing MicroOVN). 

The returned identifier doesn't have to be the name but I'm not sure if moving to a UUID would necessarily reduce the complexity, since we would still need to parse the project name out to determine if we need to re-use a UUID or generate a new one.

masnax: I should mention, in the latter case (MicroCloud importing MicroOVN), we actually don't want to prepare those statements because there will never be a use-case for them in MicroCloud, since we can't directly access the database of MicroOVN. In fact it would result in build failures because the schema wouldn't match the prepared statement. 

So part of this helper is ensuring we only import from the current project and microcluster, and ignore all other `RegisterStmt` calls in the import tree.

roosterfish: I was thinking this could be an additional field of the `type Daemon struct`.

But yes `RegisterStmt` seems problematic here as we cannot easily access the daemon and it stores the stmt's in a package level var.

Then I guess it's ok for now. Looks not to be an easy fix we could just slide into the LTS. But could you please log an issue so we can follow up on this?

roosterfish (approved):

cluster/stmt.go
masnax: https://github.com/canonical/microcluster/issues/220

PR #221, opened 2024-08-02
(backport v2) Parse go module version out of project name

masnax:
This helper finds microcluster among the cached build files to register internal statements. Since microcluster's module path has been appended with v2, this function now resolves to `v2` as the project name instead of `microcluster`, so this commit properly checks if the path is of the form v[number] and if so, looks one directory back instead.

This should fix external projects failing to register internal microcluster prepared statements.

Signed-off-by: Max Asnaashari <max.asnaashari@canonical.com>
(cherry picked from commit aaec49fa6823ddf8802336b3affbbe8878cc6a52)

### DIFF
--- a/cluster/stmt.go
+++ b/cluster/stmt.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"regexp"
 	"runtime"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


roosterfish (approved):

roosterfish:
@masnax regarding the tag, it will lead to a `v2.0.1` right?

PR #224, opened 2024-08-05
Add back legacy `/cluster/internal/database` endpoint

masnax:
We renamed `/cluster/internal`, `/cluster/control`, and `/cluster/1.0` to `/core/...` but this leads to an issue when upgrading from v1 to v2 where the newer members cannot connect to dqlite until all un-upgraded members are offline. This is because the dqlite leader will still be communicating over `/cluster/internal/database`. 

This makes upgrades noticeably slower, while also undoing the work we had done earlier to show upgrade progress with the `cluster list` command, because we can't connect to dqlite to inspect the schema versions. Instead, the daemon just reports `Database is still starting` because the dqlite connection cannot be established. The upgrade still does work, it just takes an additional 30-60s or so because the upgraded daemons will be constantly restarting until all cluster members have been upgraded.

To work around this, we can add back `/cluster/internal/database` as a core API endpoint. For the most part, this won't be used, as it will only be relevant while upgrading from an existing setup that uses the old API endpoints. This path will be reserved so extension servers cannot use `/cluster/internal/database` as a path. 

Functionally, when establishing a connection to dqlite, if we encounter a `404` error on `/core/internal/database`, we will try `/cluster/internal/database` as a fallback, allowing us to connect to legacy systems to compare schema versions.

Rather than adding this to v3 and backporting to v2, I thought we might as well fully drop the legacy support from v3 since it likely won't be released for some time, meaning direct upgrades from v1->v3 should be rather sparse. And since this doesn't actually affect the result of an upgrade, I don't think it's worth the extra maintenance.

### DIFF
--- a/internal/db/dqlite.go
+++ b/internal/db/dqlite.go
@@ -425,6 +425,28 @@ func dqliteNetworkDial(ctx context.Context, addr string, db *DqliteDB) (net.Conn
 	}
 
 	defer response.Body.Close()
+	if response.StatusCode == http.StatusNotFound {
+		_ = response.Body.Close()
+
+		// If the response returns a 404, that means this server is likely in the process of upgrading,
+		// so try the legacy /cluster/internal/database endpoint that older cluster members are using.
+		path := fmt.Sprintf("https://%s/%s/%s", addr, internalTypes.LegacyEndpoint, "database")
+		request.URL, err = url.Parse(path)
+		if err != nil {
+			return nil, err
+		}
+
+		err = request.Write(conn)
+		if err != nil {
+			return nil, fmt.Errorf("Failed sending HTTP request to %q: %w", request.URL, err)
+		}
+
+		response, err = http.ReadResponse(bufio.NewReader(conn), request)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to read response: %w", err)
+		}
+	}
+
 	_, err = io.Copy(io.Discard, response.Body)
 	if err != nil {
 		logger.Error("Failed to read dqlite response body", logger.Ctx{"error": err})
--- a/internal/rest/resources/resources.go
+++ b/internal/rest/resources/resources.go
@@ -49,6 +49,14 @@ var InternalEndpoints = rest.Resources{
 	},
 }
 
+// LegacyEndpoints allows access to the legacy /cluster/internal/database endpoint to facilitate upgrades to the newer endpoint path structure.
+var LegacyEndpoints = rest.Resources{
+	PathPrefix: internalTypes.LegacyEndpoint,
+	Endpoints: []rest.Endpoint{
+		databaseCmd,
+	},
+}
+
 // ValidateEndpoints checks if any endpoints defined in extensionServers conflict with other endpoints.
 // An invalid server is defined as one of the following:
 // - The PathPrefix+Path of an endpoint conflicts with another endpoint in the same server.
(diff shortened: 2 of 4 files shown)


masnax:
@roosterfish Tangentially, I noticed a potential issue with how we validate extension server path collisions.

We explicitly validate that no extension server clashes with an explicitly defined internal endpoint, but I think we should namespace all of `/core/internal`, `core/1.0`, and `/core/control` as reserved, because we may run into endpoint clashes for future API extensions to the `/core` API.

tomponline:
@masnax do you have upgrade tests in microcluster or microcloud? If not this would seem to be a good thing to add. Similar to https://github.com/canonical/lxd-ci/blob/main/tests/cluster#L149

internal/rest/resources/resources.go
```diff
@@ -49,6 +49,14 @@ var InternalEndpoints = rest.Resources{
 	},
 }
 
+// LegacyEndpoints allows access to the legacy /cluster/internal/database endpoint to facilitate upgrades to the newer endpoint path structure.
+var LegacyEndpoints = rest.Resources{
```
roosterfish, line 53: What about having a slice of aliases for the `Resources` type, same as we have for the `Endpoint`? 

This would allow having the alias concept on a higher level and we don't have to introduce the `LegacyEndpoint` type in the public facing package.

We can just do:

```go
var InternalEndpoints = rest.Resources{
	PathPrefix: internalTypes.InternalEndpoint,
    PathPrefixAliases: []string{"/cluster/internal"}
	Endpoints: []rest.Endpoint{
		databaseCmd,
        ...
	},
}
```

Ofc this also adds all of the other endpoints under this alias but since it's the internal endpoint I think it's acceptable.

internal/db/dqlite.go
```diff
@@ -425,6 +425,28 @@ func dqliteNetworkDial(ctx context.Context, addr string, db *DqliteDB) (net.Conn
 	}
 
 	defer response.Body.Close()
+	if response.StatusCode == http.StatusNotFound {
```
roosterfish, line 428: Mhh I have the feeling the renaming of the endpoints should have been covered by an API extension instead? 
Based on the existence of the extension we could decide to pick either one of the internal endpoint prefixes.

Maybe having an extension would be even better than adding the API endpoint also under the legacy path? Because all we have to change really is the prefix when the `path` var gets assigned above.

roosterfish (changes requested):

roosterfish:
> @roosterfish Tangentially, I noticed a potential issue with how we validate extension server path collisions.

True, good point. Issue added https://github.com/canonical/microcluster/issues/227

internal/db/dqlite.go
masnax: The problem is we are reaching out to other systems. Adding an additional request per dqlite connection to grab the target system's API extensions would be quite expensive.

internal/rest/resources/resources.go
masnax: I suppose we can make `PathPrefix` itself just a slice, we don't need a second field. 

Also, nothing here is exposed In a public facing package, it's all internal. I'm not sure if we should really serve the entire internal API over both `/cluster/internal` and `/core/internal`,  that might cause a maintainance headache in the future.

roosterfish: Yes you are right, it's located in the `internal/` package, nvm :)

internal/db/dqlite.go
roosterfish: Yes sure but isn't this only once during startup?

masnax: Well, it's continuously run as dqlite rebalances roles until all nodes are updated. 

Additionally, since not all v1 deployments support API extensions (or the `/core`) path, it would be equivalent to just sending out a request to `/core/1.0` and checking for a `404` on that instead.

roosterfish: Good point with the v1, then pls forget about this discussion.

roosterfish (approved):
Given we cannot ensure API extensions can be queried consistently across all previous versions, this looks good to me.

PR #231, opened 2024-08-09
Reserve core endpoints

### LINKED ISSUE #227: Reserve all of the top level API paths
See https://github.com/canonical/microcluster/pull/224#issuecomment-2270107231

To prevent future collisions with endpoint paths introduced by downstream projects that might collide with new endpoints we add ourselves to microcluster, the following top level paths should be reserved when additional endpoints get added to the core API:

* `/core/control`
* `/core/internal`
* `/core/1.0`

MggMuggins:
Prevent CoreAPI extension servers from creating endpoints under `/core`.

Fixes #227

### DIFF
--- a/internal/rest/resources/resources.go
+++ b/internal/rest/resources/resources.go
@@ -3,6 +3,7 @@ package resources
 import (
 	"fmt"
 	"path/filepath"
+	"strings"
 
 	"github.com/canonical/microcluster/v3/internal/endpoints"
 	internalTypes "github.com/canonical/microcluster/v3/internal/rest/types"
@@ -54,7 +55,10 @@ var InternalEndpoints = rest.Resources{
 // - The PathPrefix+Path of an endpoint conflicts with another endpoint in the same server.
 // - The address of the server clashes with another server.
 // - The server does not have defined resources.
-// If the Server is a core API server, its resources must not conflict with any other core API server, and it must not have a defined address or certificate.
+// - If the Server is a core API server:
+//   - The PathPrefix+Path of an endpoint must not begin with `core`.
+//   - Its resources must not conflict with any other core API server.
+//   - It must not have a defined address or certificate.
 func ValidateEndpoints(extensionServers map[string]rest.Server, coreAddress string) error {
 	serverAddresses := map[string]bool{coreAddress: true}
 	baseCoreEndpoints := []rest.Resources{UnixEndpoints, PublicEndpoints, InternalEndpoints}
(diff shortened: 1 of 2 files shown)


internal/rest/resources/resources.go
```diff
@@ -95,6 +95,15 @@ func ValidateEndpoints(extensionServers map[string]rest.Server, coreAddress stri
 			serverAddresses[server.Address.String()] = true
 		}
 
+		// Reserve the /core namespace for Microcluster
```
roosterfish: I have the feeling we use the string `microcluster` in lowercase all across the repos now.

internal/rest/resources/resources.go
```diff
@@ -95,6 +95,15 @@ func ValidateEndpoints(extensionServers map[string]rest.Server, coreAddress stri
...
+		if server.CoreAPI {
```
roosterfish: Worth adding another line to the doc string of `ValidateEndpoints` to document the new behavior?

internal/rest/resources/resources.go
```diff
@@ -124,3 +133,19 @@ func ValidateEndpoints(extensionServers map[string]rest.Server, coreAddress stri
...
 	return nil
 }
+
+// "core/"   -> "core"
+// "/core"   -> "core"
+// "/core/x" -> "core"
+// "/"       -> ""
+// Returns the first non-root path element in `path`.
+func firstPathElement(path string) string {
+	dir, file := filepath.Split(path)
```
roosterfish: I understand this recursively traverses the path from right to left?

I am wondering if it would be faster to "just" split the path on `/` and take either the first or second element of the slice depending on if it's a relative or absolute path?

roosterfish (changes requested):
Looks good thanks, only some smaller comments.

internal/rest/resources/resources.go
roosterfish: Maybe with a `strings.SplitN( , , 3)` to only ever look at the first elements?

MggMuggins: I wasn't concerned about speed given the application, and was hoping to use a library function that was tied to path semantics instead of generic strings. Went ahead and switched it though.

MggMuggins:
It occurred to me that only checking the PathPrefix wouldn't catch resources where `PathPrefix: ""`.

We also should probably require at least one endpoint in every resource. Happy to drop those commits if that's the wrong assumption.

masnax (approved):
LGTM, thanks @MggMuggins!

roosterfish (approved):

PR #235, opened 2024-08-13
Add role change data to `OnHeartbeat` hook

### LINKED ISSUE #209: Allow `OnHeartbeat` hook to identify Dqlite role changes
It might be helpful to indicate Dqlite role changes inside the `OnHearbeat` hook so that members can learn about their current role inside the cluster and can properly react on changes.

boltmark:
This PR adds functionality to inject role change information into the `OnHeartbeat` hook.

Closes https://github.com/canonical/microcluster/issues/209.

### DIFF
--- a/example/cmd/microd/main.go
+++ b/example/cmd/microd/main.go
@@ -171,9 +171,18 @@ func (c *cmdDaemon) run(cmd *cobra.Command, args []string) error {
 		},
 
 		// OnHeartbeat is run after a successful heartbeat round.
-		OnHeartbeat: func(ctx context.Context, s state.State) error {
-			logger.Info("This is a hook that is run on the dqlite leader after a successful heartbeat")
+		OnHeartbeat: func(ctx context.Context, s state.State, roleStatus map[string]types.RoleStatus) error {
+			logger.Info("This is a hook that is run on the dqlite leader after a successful heartbeat; role information for cluster members is available")
 
+			// You can check if the role of a cluster member has changed since the last heartbeat and determine
+			// its previous and current roles.
+			myStatus := roleStatus[s.Name()]
+			if myStatus.RoleChanged() {
+				logger.Infof("Role of cluster member %s changed from %s to %s", s.Name(), myStatus.Old, myStatus.New)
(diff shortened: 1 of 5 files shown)


example/cmd/microd/main.go
```diff
@@ -171,7 +171,7 @@ func (c *cmdDaemon) run(cmd *cobra.Command, args []string) error {
 		},
 
 		// OnHeartbeat is run after a successful heartbeat round.
-		OnHeartbeat: func(ctx context.Context, s state.State) error {
+		OnHeartbeat: func(ctx context.Context, s state.State, roleStatus map[string]types.RoleStatus) error {
 			logger.Info("This is a hook that is run on the dqlite leader after a successful heartbeat")
```
roosterfish: Worth adding an example here using the `RoleChanged()` func?

I am thinking about something like showing the members own status.
You can get the name from the state: `roleStatus[s.Name()].RoleChanged()`.

rest/types/cluster.go
```diff
@@ -28,6 +28,17 @@ type ClusterMemberLocal struct {
 // MemberStatus represents the online status of a cluster member.
 type MemberStatus string
 
+// RoleStatus represents the previous and current role of a cluster member.
+type RoleStatus struct {
+	OldRole string
```
roosterfish: Not a blocker bot I am wondering if we should duplicate `Role` in the names too as the struct itself already indicates we are talking about roles here?

roosterfish (changes requested):
Looking good but please make use of the new func also in the example package to show off/document the feature.

rest/types/cluster.go
boltmark: I agree that it's unnecessary in the field names. Fixed!

example/cmd/microd/main.go
boltmark: Added some example usage, let me know if it looks ok.

roosterfish (approved):
LGTM!

masnax (approved):
LGTM, thanks @boltmark!

PR #237, opened 2024-08-16
Client: Add websocket client and more control over the cert selection

roosterfish:
This PR adds a new client `Websocket` function that allows upgrading a HTTP connection using either the local or remote client.

Furthermore another `RemoteClientWithCert` app function is added which allows providing a custom remote server cert in case the remote is not using the common cluster cert. 
This is helpful during the pre-init phase (MicroCloud explicit trust establishment) as well as when trying to connect to additional servers making use of mTLS.

### DIFF
--- a/internal/rest/client/client.go
+++ b/internal/rest/client/client.go
@@ -22,6 +22,7 @@ import (
 	"github.com/canonical/lxd/shared/api"
 	"github.com/canonical/lxd/shared/logger"
 	"github.com/canonical/lxd/shared/tcp"
+	"github.com/gorilla/websocket"
 
 	"github.com/canonical/microcluster/v3/rest/types"
 )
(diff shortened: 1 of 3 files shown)


roosterfish:
@masnax do we also want another `Websocket` function in the `client/client.go` file?

internal/rest/client/client.go
```diff
@@ -312,6 +318,54 @@ func (c *Client) QueryStruct(ctx context.Context, method string, endpointType ty
...
+	}
+
+	// Setup a new websocket dialer using the already existing HTTP client.
+	// As the client might be either local or remote always copy the relevant TLS config too.
+	dialer := websocket.Dialer{
+		NetDialContext:    tr.DialContext,
+		NetDialTLSContext: tr.DialTLSContext,
+		TLSClientConfig:   tr.TLSClientConfig,
+		Proxy:             tr.Proxy,
+		HandshakeTimeout:  time.Second * 5,
```
masnax: Instead of hard-coding this timeout, I think it might be more configurable to use the timeout from the context, and then fallback to a default if the context has no timeout.

In `rawQuery`, we check that a context has a deadline, and if not we apply a default 30s timeout. I'm not sure if something like that would be useful here as well:
```go
// Assign a context timeout if we don't already have one.
	_, ok := ctx.Deadline()
	if !ok {
		timeoutCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
		ctx = timeoutCtx
		defer cancel()
	}
```

masnax (changes requested):
Looking good, just one nit/question.

In regards to adding an exposed `Websocket` helper, I think it would be useful to have one just so we don't force projects to use the internal client struct, which can lead to big refactors if we decide to add extra cases to the exposed client struct.

internal/rest/client/client.go
roosterfish: I like the idea of adding the timeout to the context if a deadline doesn't yet exist.

To be consistent with the timeout duration I have copied the one from the LXD client https://github.com/canonical/lxd/blob/main/client/lxd.go#L423.
But in LXD we don't use `dialer.DialContext` but just `dialer.Dial`. 
Interestingly in gorilla they apply the [exact same pattern](https://github.com/gorilla/websocket/blob/main/client.go#L241) in case `HandshakeTimeout` is set.

So by setting the deadline ourselves we can omit setting the `HandshakeTimeout` field.

roosterfish: Fixed.

masnax (approved):
LGTM! Thanks @roosterfish

internal/rest/client/client.go
masnax: Oh, one thing that occurred to me is that by using `DialContext` with a 30s default context timeout, that means this websocket will only live for a maximum of 30s, right?

Would it be better in this case to require that the context _must have_ a deadline, rather than setting a default? I'm assuming since this is a generic websocket client helper, there may be use-cases for longer-lived websockets.

roosterfish: The context provided for `DialContext` only covers the dialing phase (sending the HTTP upgrade request and waiting for a response). 
The returned `*websocket.Conn` isn't anymore affected by the context.

So the default ensures that the websocket can be established during this time.

PR #238, opened 2024-08-16
Reserve core endpoints backport to v2

MggMuggins:
Backport of #231

Not sure if I did it right but also backported #236 so that the tests pass


masnax (approved):
Thanks @MggMuggins

PR #241, opened 2024-08-22
Rest: Return validation errors from the database handlers

### LINKED ISSUE #240: "Forbidden" database errors are silently ignored
Let's asume a Canonical K8s (snap) cluster with invalid certificates. When starting k8sd, we'll get "Forbidden" errors. The problem is that:

* those errors are not logged
* microcluster returns status code 200 instead of 403
  * https://github.com/canonical/microcluster/blob/9da23064cf0ad86768e05db4695a9d239bb717c1/internal/rest/rest.go#L246-L255 

This significantly hinders the debugging process since the k8sd logs will not contain any relevant messages. All we get is that the dialing failed as it expected a protocol switch (101) but got 200 (why not 403?) https://github.com/canonical/microcluster/blob/42551841a56dd2d4697d577428bdb4d7d5f877d0/internal/db/dqlite.go#L441

roosterfish:
Fixes https://github.com/canonical/microcluster/issues/240

This PR ensures that any request validation for the database handlers happens before hijacking the connection so that errors can be returned to the caller.

### DIFF
--- a/internal/rest/resources/database.go
+++ b/internal/rest/resources/database.go
@@ -29,7 +29,7 @@ func databasePost(state state.State, r *http.Request) response.Response {
 
 	_, err := strconv.Atoi(versionHeader)
 	if err != nil {
-		return response.BadRequest(fmt.Errorf("Invalid dqlite vesion: %w", err))
+		return response.BadRequest(fmt.Errorf("Invalid dqlite version: %w", err))
 	}
 
 	// Handle leader address requests.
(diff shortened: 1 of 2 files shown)


masnax (approved):
LGTM!

masnax:
Should this get backported to `v2`?

roosterfish:
> Should this get backported to `v2`?

Yes I'll create a PR.

PR #242, opened 2024-08-22
Replace `PreBootstrap` with `PreInit` hook

masnax:
`PreBootstrap` runs immediately after we apply the daemon name & address, while `PreJoin` is intended to run after joining dqlite, but before completing any other join steps.

 It's useful to have the behaviour of `PreBootstrap` for joining too, so this PR replaces it with `PreInit` which takes `bootstrap bool` as an argument.

This required a bit of refactoring because StartAPI isn't necessarily the best place for this hook to run, since we request the cluster to add the new node first.

So the changes here are as follows:

* Moves the `daemonConfig` handling from StartAPI into its own `SetConfig` helper that's exposed on the internal State struct. We will manually call this when bootstrapping/joining.
* `PreInit` will fire immediately after applying this config so the daemon has a name & address.
* LXD dependency has been updated to pull in https://github.com/canonical/lxd/pull/13960
* Ability to change `server.crt` over the API has been added. This will only work before the database is initialized.


masnax:
@bschimke95 You should be able to update the certificate like so:
```go
  PreInit: func(ctx context.Context, s state.State, bootstrap bool, initConfig map[string]string) error {
      err := os.Remove(filepath.Join(s.FileSystem().StateDir, "server.crt"))
      if err != nil {
        return err
      }

      err = os.Remove(filepath.Join(s.FileSystem().StateDir, "server.key"))
      if err != nil {
        return err  Not Committed Yet
      }

      cert, err := shared.KeyPairAndCA(s.FileSystem().StateDir, string(types.ServerCertificateName), shared.CertServer, shared.CertOptions{AddHosts: true, CommonName: s.Name(), SubjectAlternativeNames: []string{initConfig["cluster_name"]}})
      if err != nil {
        return err
      }

      c, err := m.LocalClient()
      if err != nil {
        return err
      }

      err = c.UpdateCertificate(cmd.Context(), types.ServerCertificateName, types.KeyPair{
        Cert: string(cert.PublicKey()),
        Key:  string(cert.PrivateKey()),
      })
      if err != nil {
        return err
      }

      return nil
    },

```

internal/daemon/daemon.go
```diff
@@ -488,34 +491,18 @@ func (d *Daemon) initServer(resources ...rest.Resources) *http.Server {
...
-		if err != nil {
-			return err
-		}
-	} else {
-		err := d.config.Load()
-		if err != nil {
-			return err
-		}
-	}
+// setConfig Applies and commits to memory the supplied daemon configuration.
```
bschimke95: nit: applies

internal/daemon/daemon.go
```diff
@@ -353,8 +356,8 @@ func (d *Daemon) applyHooks(hooks *state.Hooks) {
 		d.hooks = *hooks
 	}
 
-	if d.hooks.PreBootstrap == nil {
-		d.hooks.PreBootstrap = noOpInitHook
+	if d.hooks.PreInit == nil {
+		d.hooks.PreInit = noOpGenericInitHook
```
bschimke95, line 360: nit: I'd just keep `noOpInitHook`

internal/rest/resources/control.go
```diff
@@ -59,6 +59,19 @@ func controlPost(state state.State, r *http.Request) response.Response {
 		return response.SmartError(err)
 	}
 
+	daemonConfig := trust.Location{Address: req.Address, Name: req.Name}
+	err = intState.SetConfig(daemonConfig)
```
bschimke95, line 63: nit: error check can be inlined (here and below)

bschimke95 (commented):
Thanks for that feature. Just added a couple of nit comments while going through the code.

internal/rest/resources/certificates.go
```diff
@@ -82,6 +82,11 @@ func clusterCertificatesPut(s state.State, r *http.Request) response.Response {
 	var certificateDir string
 	if certificateName == string(types.ClusterCertificateName) {
 		certificateDir = s.FileSystem().StateDir
+	} else if certificateName == string(types.ServerCertificateName) {
```
roosterfish, line 85: Mh, in this case do we have to stop doing the cluster notification a few lines above? I guess otherwise we will update the server certificate on all nodes.

Before this wasn't an issue because if the cert name wasn't `cluster` it was checking in the `certificates` dir.

internal/daemon/daemon.go
```diff
@@ -493,23 +493,7 @@ func (d *Daemon) initServer(resources ...rest.Resources) *http.Server {
 
 // StartAPI starts up the admin and consumer APIs, and generates a cluster cert
 // if we are bootstrapping the first node.
-func (d *Daemon) StartAPI(ctx context.Context, bootstrap bool, initConfig map[string]string, newConfig *trust.Location, joinAddresses ...string) error {
-	if newConfig != nil {
```
roosterfish, line 494: There was the scenario where if the cluster isn't yet initialized (pre-init) and you query the cluster status endpoint you will get an error as the node address couldn't be parsed as it wasn't yet set.

Would this return this bug?

roosterfish (changes requested):

internal/rest/resources/certificates.go
masnax: The cluster notification has a check that requires the database to be online. It will actually catch more states than the check in this block which explicitly wants `StatusNotReady`

internal/daemon/daemon.go
masnax: It shouldn't, the behaviour of this `if` block is still preserved. 
We call `StartAPI` in 3 places:
* restart - (uses the `else` case), it was actually already redundant since we call `config.Load` [here](https://github.com/canonical/microcluster/blob/a4c5273899ac294750d63089f81f78fd1894df67/internal/daemon/daemon.go#L418-L426)
* join - it's still called before we run `StartAPI` [here](https://github.com/canonical/microcluster/blob/6dc1ebc092f502d6e71a22dad9ff2ef5977323aa/internal/rest/resources/control.go#L63)
* bootstrap - it's still called before we run `StartAPI` [here](https://github.com/canonical/microcluster/blob/6dc1ebc092f502d6e71a22dad9ff2ef5977323aa/internal/rest/resources/control.go#L63)

We also support returning an empty addess from the status endpoint, so it shouldn't be an issue anyway.

internal/rest/resources/certificates.go
masnax: https://github.com/canonical/microcluster/blob/6375e4d511c6458c6011e0a7a83057742bd07719/internal/rest/resources/certificates.go#L47-L53

roosterfish: True, I haven't seen the `s.Database().IsOpen(r.Context())` check. Thx.

internal/daemon/daemon.go
roosterfish: Thanks for pointing me to https://github.com/canonical/microcluster/blob/a4c5273899ac294750d63089f81f78fd1894df67/internal/rest/resources/api_1.0.go#L21-L25 which was what I meant originally :)

roosterfish (approved):

internal/rest/resources/control.go
masnax: We tend to avoid inline errors for consistency, but thanks for the suggestion.

PR #244, opened 2024-08-27
Token: Validate name to be FQDN

roosterfish:
Thanks @bschimke95 for reporting this.

Check if the supplied token name is a valid FQDN as it will be used later in the process when joining new cluster members into the existing cluster.
Those cluster member names are also validated to be a valid FQDN.

### DIFF
--- a/internal/rest/resources/tokens.go
+++ b/internal/rest/resources/tokens.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"database/sql"
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"net/url"
 	"time"
@@ -16,6 +17,7 @@ import (
 	"github.com/canonical/microcluster/v3/cluster"
 	internalTypes "github.com/canonical/microcluster/v3/internal/rest/types"
 	"github.com/canonical/microcluster/v3/internal/state"
+	"github.com/canonical/microcluster/v3/internal/utils"
 	"github.com/canonical/microcluster/v3/rest"
 	"github.com/canonical/microcluster/v3/rest/access"
 	"github.com/canonical/microcluster/v3/rest/types"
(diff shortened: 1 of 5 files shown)


internal/rest/resources/tokens.go
```diff
@@ -43,6 +45,11 @@ func tokensPost(state state.State, r *http.Request) response.Response {
 		return response.BadRequest(err)
 	}
 
+	err = utils.ValidateFQDN(req.Name)
+	if err != nil {
```
bschimke95, line 49: nit: error check can be in-lined

bschimke95 (commented):
Thanks for the quick fix!
Only a minor nit from my side.

internal/rest/resources/tokens.go
roosterfish: We usually don't use the pattern `if err := something(); err != nil {}` across the code base.

If there are longer lines it can impact the readability.

bschimke95: alright, thanks for the context!

bschimke95: alright, thanks for the context!

internal/rest/resources/tokens.go
```diff
@@ -43,6 +45,11 @@ func tokensPost(state state.State, r *http.Request) response.Response {
...
+		return response.SmartError(fmt.Errorf("Invalid token name %q: %w", req.Name, err))
```
masnax: I think we should be explicit and report `Token name is not a valid FQDN: %w`

masnax (changes requested):
Just 1 nit from me :)

internal/rest/resources/tokens.go
roosterfish: Fixed. I have also updated the other messages where we return the error from `ValidateFQDN`.

roosterfish: We should probably also return `BadRequest` from within `ValidateFQDN` otherwise those are getting reported as `InternalError`.

masnax (approved):
LGTM! Thanks :)

PR #246, opened 2024-08-28
Utils: Return BadRequets from ValidateFQDN

roosterfish:
Whilst adding https://github.com/canonical/microcluster/pull/244 I found that an `InternalError` is returned from `ValidateFQDN` instead of `BadRequest`.


masnax (approved):
LGTM!

PR #248, opened 2024-08-29
Update LXD for `KeyPairAndCA`

masnax:


roosterfish (changes requested):
Any reason for not cherry-picking the commits out of https://github.com/canonical/microcluster/pull/242? 

The version of LXD seems to be more recent than what we currently have in `v3`.

masnax:
@roosterfish Good point, updated now.

roosterfish (approved):
LGTM!

PR #249, opened 2024-09-06
Client: Add new `QueryRaw` function

roosterfish:
This PR adds another `QueryRaw` function to the client which allows accessing the raw `*http.Response`.
That is required for the https://github.com/canonical/microcloud/pull/383 in order to access the remotes certificate used in the TLS connection.

Furthermore the parsing of API responses is moved into the public facing package to allow any user of `QueryRaw` to both access the raw response and parsing it as usual.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -31,6 +31,15 @@ func (c *Client) Query(ctx context.Context, method string, prefix types.Endpoint
 	return c.QueryStruct(queryCtx, method, prefix, path, in, &out)
 }
 
+// QueryRaw is a helper for initiating a request on any endpoints defined external to microcluster.
+// Unlike Query it returns the raw HTTP response.
+func (c *Client) QueryRaw(ctx context.Context, method string, prefix types.EndpointPrefix, path *api.URL, in any) (*http.Response, error) {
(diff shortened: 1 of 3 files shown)


client/client.go
```diff
@@ -31,6 +31,16 @@ func (c *Client) Query(ctx context.Context, method string, prefix types.Endpoint
 	return c.QueryStruct(queryCtx, method, prefix, path, in, &out)
 }
 
+// QueryRaw is a helper for initiating a request on any endpoints defined external to microcluster. This function should be used for all client
```
masnax: Maybe remove `This function should be used for all client methods defined externally from microcluster` here since it's more of a special case

masnax (changes requested):
Just one tiny nit :)

client/client.go
roosterfish: Thanks for bringing this up.
Had the same thought as you but left it in to clearly indicate that you should not use the internal `QueryStruct/QueryStructRaw` methods.

But probably at some time we will anyway revoke access to those so we can already start thinking this way. Removed :)

masnax (approved):

PR #257, opened 2024-09-26
Trivy vulnerability scanning

hamistao:
This adds a new workflow that uses Trivy to scan microcluster's dependencies daily.

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -1,0 +1,47 @@
+name: Vulnerability Scanning with Trivy
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 0 * * *'  # Test Trivy daily at midnight
+
+permissions:
+  contents: read
+  security-events: write # for uploading SARIF results to the security tab
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.event_name }}
+  cancel-in-progress: true
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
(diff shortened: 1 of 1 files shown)


hamistao:
@tomponline @simondeziel This is also ready

.github/workflows/security.yml
```diff
@@ -0,0 +1,42 @@
...
+          format: "sarif"
+          output: "trivy-microcluster-repo-scan-results.sarif"
+          severity: "LOW,MEDIUM,HIGH,CRITICAL"
+
+      - name: Upload Trivy scan results to GitHub Security tab
+        uses: github/codeql-action/upload-sarif@v3
+        with:
+          sarif_file: "trivy-microcluster-repo-scan-results.sarif"
+          sha: ${{ github.sha }}
+          ref: refs/heads/main
```
simondeziel: Should we avoid hardcoding `main` here?

hamistao: would you suggest using a matrix strategy instead?

simondeziel: I just checked and there is no `main` branch, just `v2` and `v3` ATM so yeah, a matrix sounds like a good idea.

hamistao: Indeed, my fork does only have a main branch for some reason so this is why my tests worked, thanks for the warning!

hamistao: Done!

simondeziel (approved):

tomponline (approved):

PR #264, opened 2024-10-10
Allow empty AddrPorts

masnax:
https://github.com/canonical/microcluster/pull/179 attempted to allow empty addrports for the `/core/1.0`  API, but did not consider that after sending the request, parsing into a struct it on the other side would still fail. 

So instead, the `types.ParseAddrPort` helper now fully supports empty addresses.


roosterfish (approved):
Thanks!

roosterfish:
Just wondering where did this cause trouble?

masnax:
> Just wondering where did this cause trouble?

Daemon enforcement. MicroCeph/OVN were correctly sending their results from `/core/1.0` but we could not parse them in MicroCloud, since there was no init address.

PR #270, opened 2024-10-22
go: Update deps

roosterfish:
Update the LXD dep to get rid of the `Request will not be counted for the API metrics` warning log message on API requests.


masnax (approved):

roosterfish:
We don't have to necessarily backport this to `v2` as the version of LXD included there does not yet introduce the change that is causing the warning log message.

PR #273, opened 2024-10-24
dqlite: Use `go-dqlite` v3 and build libdqlite from master branch

roosterfish:


roosterfish:
@kadinsayani is there a specific reason for running the unit and system tests as root? Is this required by TICS?

kadinsayani:
> @kadinsayani is there a specific reason for running the unit and system tests as root? Is this required by TICS? 

It's not required by TICS. The only requirement is the presence of GOCOVERDIR.

kadinsayani (approved):
LGTM! Thanks!

MggMuggins (approved):
Looks great! I'd love to see this get merged so that we can implement the equivalent of https://github.com/canonical/lxd/pull/14378

go.mod
```diff
@@ -20,6 +20,7 @@ require (
 require (
 	github.com/Rican7/retry v0.3.1 // indirect
 	github.com/armon/go-proxyproto v0.1.0 // indirect
+	github.com/canonical/go-dqlite v1.22.0 // indirect
```
tomponline, line 23: @roosterfish what is left that is depending on go-dqlite v1?

roosterfish: I am pretty sure that comes from the old version of LXD before merging the dqlite v2 PR.

roosterfish: Should be fixed after we track `@stable-5.21`.

tomponline: But you're not changing this branch to track `@stable-5.21` so should already be gone by updating the lxd dep?

roosterfish: The LXD dep used here was still an older one before LXD was switched to dqlite v2.

PR #279, opened 2024-11-06
Manually register error mappings

roosterfish:
To remove pulling other versions of dqlite the registration of error mappings [got moved out](https://github.com/canonical/lxd/pull/14408) of `lxd/response`.
Therefore we have to set the errors manually.


tomponline (approved):

roosterfish:
The two errors from the mapping aren't used by microcluster directly but we cannot rule out that any downstream returns those from their API handlers.

So for the sake of completeness we still set them when starting the daemon.

PR #283, opened 2024-11-07
github: Use cache when DB download fails

hamistao:
This updates the Trivy workflow to cache the vulnerability database when the download is successful and, when it is not, use the latest cached database instead. Motivated by some failures seen on some recent runs: https://github.com/canonical/microcluster/actions/runs/11694779309/job/32569052448.

Installing Trivy using our GitHub action from LXD is now needed since we are now making use of the `cache-dir` flag.

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -18,7 +18,7 @@ defaults:
 
 jobs:
   trivy-repo:
-    name: Trivy vulnerability scanner - Repository
+    name: Trivy - Repository
     runs-on: ubuntu-22.04
     strategy:
       matrix:
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -31,13 +31,36 @@ jobs:
         with:
           ref: ${{ matrix.version }}
 
-      - name: Run Trivy vulnerability scanner in repo mode
-        uses: aquasecurity/trivy-action@master
+      - name: Install Trivy
+        uses: canonical/lxd/.github/actions/install-trivy@main
+
+      - name: Download Trivy DB
+        id: db_download
+        run: trivy fs --download-db-only --cache-dir /home/runner/vuln-cache
+        continue-on-error: true
+
+      - name: Use previous downloaded database
+        if: ${{ steps.db_download.outcome == 'failure' }}
+        uses: actions/cache/restore@3624ceb22c1c5a301c8db4169662070a689d9ea8 # v4.1.1
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/security.yml
```diff
@@ -31,13 +31,36 @@ jobs:
...
+        uses: canonical/lxd/.github/actions/install-trivy@main
+
+      - name: Download Trivy DB
+        id: db_download
+        run: trivy fs --download-db-only --cache-dir /home/runner/vuln-cache
+        continue-on-error: true
+
+      - name: Use previous downloaded database
+        if: ${{ steps.db_download.outcome == 'failure' }}
+        uses: actions/cache/restore@3624ceb22c1c5a301c8db4169662070a689d9ea8 # v4.1.1
```
roosterfish, line 44: Where does this point to? Which repo?

hamistao: Do you mean `actions/cache`?
If so, it is a GitHub action for caching data as a way to optimize workflows: https://github.com/actions/cache

roosterfish: Ah my bad, I was referring to the commit `3624...` and was thinking you restore the cache of a previous action ran for this commit. Never mind :)

roosterfish: Actually it's the version of the action.

roosterfish: And you restore whatever is stored under `trivy-latest-cache`.

hamistao: > And you restore whatever is stored under trivy-latest-cache.

Precisely!

roosterfish (approved):
Thanks, LGTM!

roosterfish:
@hamistao looks this doesn't work as the cache `trivy-latest-cache` never got populated before, see https://github.com/canonical/microcluster/actions/runs/11788311631/job/32835187066.

Can we populate it manually?

tomponline:
@hamistao @roosterfish have we checked what other teams are doing for this?

hamistao:
> Can we populate it manually?

Maybe there is a simpler way, but I think that would require us to make a Trivy database accessible to the workflow and create a dummy workflow to populate the cache. We should be good to go as soon as we get a first successful run and from that point on the cache would be always present.

> have we checked what other teams are doing for this?

I only checked `k8s-snap` and it is not doing much: https://github.com/canonical/k8s-snap/actions/workflows/cron-jobs.yaml

@tomponline @roosterfish

PR #287, opened 2024-11-13
github: Fix database cache update

hamistao:
A github cache can not be updated using the same key, so instead we update using a different key each time but use the same key to restore it.

This also adds a step to download/update the vulnerability database as a way to avoid having the v3 and v2 workflows updating the cache concurrently, as that can cause the update to fail.

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -17,9 +17,29 @@ defaults:
     shell: bash
 
 jobs:
+  update-vulnerability-database:
+    name: Trivy - Update cached database
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Install Trivy
+        uses: canonical/lxd/.github/actions/install-trivy@main
+
+      - name: Download Trivy DB
+        id: db_download
+        run: trivy fs --download-db-only --cache-dir /home/runner/vuln-cache
+        continue-on-error: true
+
+      - name: Cache Trivy vulnerability database
+        if: ${{ steps.db_download.outcome == 'success' }}
+        uses: actions/cache/save@6849a6489940f00c2f30c0fb92c6274307ccb58a # v4.1.2
+        with:
+          path: /home/runner/vuln-cache
+          key: trivy-latest-cache
+
   trivy-repo:
     name: Trivy - Repository
     runs-on: ubuntu-22.04
+    needs: update-vulnerability-database
     strategy:
       matrix:
         version:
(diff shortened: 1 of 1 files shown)


.github/workflows/security.yml
```diff
@@ -17,30 +17,43 @@ defaults:
...
+        continue-on-error: true
+
+      - name: Cache Trivy vulnerability database
+        if: ${{ steps.db_download.outcome == 'success' }}
+        uses: actions/cache@3624ceb22c1c5a301c8db4169662070a689d9ea8 # v4.1.1
+        with:
+          path: /home/runner/vuln-cache
+          key: trivy-cache-${{ github.run_id }}
+          restore-keys: |
+            trivu-latest-cache
```
simondeziel: `s/trivu/trivy/`

hamistao: thanks for noticing, fixed!

roosterfish: @hamistao did you forget to push?

.github/workflows/security.yml
```diff
@@ -17,30 +17,43 @@ defaults:
...
+
   trivy-repo:
     name: Trivy - Repository
     runs-on: ubuntu-22.04
+    needs: update-vulnerability-database
```
roosterfish, line 42: Can we do the same as in https://github.com/canonical/microcloud/pull/485 using the `restore-keys:` option?

roosterfish (changes requested):

.github/workflows/security.yml
hamistao: Yes, the current implementation is actually a product of my misunderstanding of the `restore-keys` option.
The changes are pushed and now it works similarly to MicroCloud (the correct way to use it).
I will test this first and then ping you for another review.

.github/workflows/security.yml
hamistao: In the current version we are using a different key altogether so I will close this

.github/workflows/security.yml
roosterfish: Thanks!

roosterfish (approved):

hamistao:
@roosterfish Are you authorized to merge this? If so, is there anyone else you would like to review this?

roosterfish:
> Are you authorized to merge this

Forgot about it, thanks for the reminder.

PR #239, opened 2024-08-20
v2: Backport websocket client functions

roosterfish:
Backports https://github.com/canonical/microcluster/pull/237 into `v2`.

PR #290, opened 2024-11-14
Backport PreInit hook support, and deprecate PreBootstrap

masnax:
k8s-snap has requested PreInit to be included in the LTS as they are dependent on it for their LTS.

This feature clashes with the PreBootstrap hook and was meant to replace it, so instead PreBootstrap is set as deprecated, and will be replaced with a noop function if PreInit is defined by the upstream package.

### DIFF
--- a/example/cmd/microd/main.go
+++ b/example/cmd/microd/main.go
@@ -111,13 +111,13 @@ func (c *cmdDaemon) run(cmd *cobra.Command, args []string) error {
 			return nil
 		},
 
-		PreBootstrap: func(ctx context.Context, s state.State, initConfig map[string]string) error {
+		PreInit: func(ctx context.Context, s state.State, bootstrap bool, initConfig map[string]string) error {
 			logCtx := logger.Ctx{}
 			for k, v := range initConfig {
 				logCtx[k] = v
 			}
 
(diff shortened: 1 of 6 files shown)


masnax:
While setting it as deprecated is more correct, I'm a bit concerned with breaking pipelines due to linter issues which will need to be ignored or addressed if they update to `v2.0.6`.

roosterfish (dismissed):
Looks good to me but I think we should rather release this in a `v2.x.0` instead of a patch version as it deprecates existing features. 
Whilst thinking more about this probably adding the [websocket client](https://github.com/canonical/microcluster/pull/239) function (as an example) would have also been a good candidate for a `v2.x.0` instead of `v2.0.x`. 

Either way I think it's not considered to be a breaking change if downstream pipelines will now produce a warning due to the deprecation of the func.
I would even say that is wanted so that they can either switch the func or proactively add an ignore statement themselves.

bschimke95 (approved):
LGTM, thanks for the backport!

masnax:
I've changed this to now error out if both PreBootstrap and PreInit are defined, rather than silently unset PreBootstrap.

roosterfish (approved):

PR #291, opened 2024-11-16
internal/trust/remotes: check remote.Name to be a valid file name

simondeziel:
The previous attempt at fixing https://github.com/canonical/microcluster/security/code-scanning/1 was flawed as it tried to fix a bogus `remote.Name` by turning it into a valid file name using `filepath.Base(remote.Name)`. However, sanitizing the input is not recommended as it's risky so it's best to reject known bad input instead.

This is exactly what the "good example" provided in https://github.com/canonical/microcluster/security/code-scanning/1 is doing: make sure the path provided is a valid file name. As such, let's do that when first accepting a `remote.Name` to derive a `.yaml` file for it.

### DIFF
--- a/internal/trust/remotes.go
+++ b/internal/trust/remotes.go
@@ -39,6 +39,20 @@ type Location struct {
 	Address types.AddrPort `yaml:"address"`
 }
 
+// disallowedFileNameSubcontents contains the list of disallowed substrings in remote names.
+var disallowedFilenameSubcontents = []string{"..", "/", "\\"}
+
+// validateRemoteName checks if the remote name contains any disallowed substrings.
+func validateRemoteName(name string) error {
+	for _, disallowed := range disallowedFilenameSubcontents {
+		if strings.Contains(name, disallowed) {
+			return fmt.Errorf("Invalid remote name %q. Contains illegal subcontent %q", name, disallowed)
+		}
+	}
+
+	return nil
+}
+
 // Load reads any yaml files in the given directory and parses them into a set of Remotes.
 func (r *Remotes) Load(dir string) error {
 	r.updateMu.Lock()
@@ -95,6 +109,11 @@ func (r *Remotes) Add(dir string, remotes ...Remote) error {
 	for _, remote := range remotes {
 		if remote.Certificate.Certificate == nil {
 			return fmt.Errorf("Failed to parse local record %q. Found empty certificate", remote.Name)
+		}
+
+		err := validateRemoteName(remote.Name)
+		if err != nil {
+			return err
 		}
 
 		_, ok := r.data[remote.Name]
--- a/internal/trust/remotes.go
+++ b/internal/trust/remotes.go
@@ -53,6 +53,20 @@ func validateRemoteName(name string) error {
 	return nil
 }
 
+// remoteYamlPath returns the path to the remote's YAML file. The path is checked to ensure it is within the given directory.
+func remoteYamlPath(dir, name string) (string, error) {
+	path, err := filepath.Abs(filepath.Join(dir, name+".yaml"))
+	if err != nil {
+		return "", fmt.Errorf("Failed to get absolute path to %q yaml: %w", name, err)
+	}
+
+	if !strings.HasPrefix(path, dir) {
+		return "", fmt.Errorf("Invalid path to %q yaml", name)
+	}
+
+	return path, nil
+}
+
 // Load reads any yaml files in the given directory and parses them into a set of Remotes.
 func (r *Remotes) Load(dir string) error {
 	r.updateMu.Lock()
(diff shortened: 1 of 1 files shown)


masnax (dismissed):

internal/trust/remotes.go
```diff
@@ -111,7 +111,11 @@ func (r *Remotes) Add(dir string, remotes ...Remote) error {
 			return fmt.Errorf("Failed to parse remote %q to yaml: %w", remote.Name, err)
 		}
 
-		path := filepath.Join(dir, filepath.Base(remote.Name+".yaml"))
+		path, err := filepath.Abs(filepath.Join(dir, remote.Name+".yaml"))
+		if err != nil || !strings.HasPrefix(path, dir) {
+			return fmt.Errorf("Failed to get absolute path to %q yaml: %w", remote.Name, err)
```
roosterfish: In case `err == nil` and `strings.HasPrefix(path, dir)` returns `false` the error message tries to include the `nil` error which is going to print something like `%!w(<nil>)`.

In the [example](https://github.com/canonical/microcluster/security/code-scanning/1) they just ignore the error but maybe it's best to have another `if` block after the call and check for `Abs` to transparently return the error (if any) and return something like "Invalid path %q to yaml" in case the prefix doesn't match.

internal/trust/remotes.go
```diff
@@ -162,7 +162,11 @@ func (r *Remotes) Replace(dir string, newRemotes ...types.ClusterMember) error {
 			return fmt.Errorf("Failed to parse remote %q to yaml: %w", remote.Name, err)
 		}
 
-		remotePath := filepath.Join(dir, filepath.Base(remote.Name+".yaml"))
+		remotePath, err := filepath.Abs(filepath.Join(dir, remote.Name+".yaml"))
+		if err != nil || !strings.HasPrefix(remotePath, dir) {
```
roosterfish: The if block should also be split to prevent printing an nil error `%!w(<nil>)`.

roosterfish (changes requested):
Can we progress with this PR to resolve https://github.com/canonical/microcluster/security/code-scanning/1?

internal/trust/remotes.go
simondeziel: I ended up not reporting `err` from `filepath.Abs` failure as no matter what's the source of error (`.Abs()` or `.HasPrefix()`), I feel that "Failed to get absolute path to %q yaml" is descriptive enough.

simondeziel:
> Can we progress with this PR to resolve [Uncontrolled data used in path expression](/canonical/microcluster/security/code-scanning/1)?

Sorry for the delay, it's ready for another round of review now.

internal/trust/remotes.go
```diff
@@ -97,6 +97,10 @@ func (r *Remotes) Add(dir string, remotes ...Remote) error {
 			return fmt.Errorf("Failed to parse local record %q. Found empty certificate", remote.Name)
 		}
 
+		if strings.Contains(remote.Name, "/") || strings.Contains(remote.Name, "\\") || strings.Contains(remote.Name, "..") {
+			return fmt.Errorf("Invalid remote name %q. Contains illegal character", remote.Name)
+		}
+
```
markylaing: You could potentially extract this into a function. E.g.
```go
var disallowedFilenameSubcontents = []string{"..", "/", "\\"}
func validateFileName(in string) error {
    for _, disallowed := range disallowedFilenameSubcontents {
        if strings.Contains(in, disallowed) {
            return fmt.Errorf("Invalid remote name %q. Contains illegal subcontent %q", remote.Name, disallowed)
        }
    }
}
```

It could also be defined elsewhere.

For now I think it's fine though, just suggestions :)

internal/trust/remotes.go
markylaing: This is now changed to remove the err from the printf. I think this is fine as I can't imagine we care too much about the error returned from filepath.Abs

markylaing (dismissed):
LGTM

There are some potential improvements but I'm happy as is

internal/trust/remotes.go
simondeziel: Thanks, I added it :)

internal/trust/remotes.go
```diff
@@ -201,7 +205,11 @@ func (r *Remotes) Replace(dir string, newRemotes ...types.ClusterMember) error {
 		_, ok := remoteData[name]
 
 		if !ok {
-			remotePath := filepath.Join(dir, filepath.Base(entry.Name()))
+			remotePath, err := filepath.Abs(filepath.Join(dir, entry.Name()))
+			if err != nil || !strings.HasPrefix(remotePath, dir) {
+				return fmt.Errorf("Failed to get absolute path to %q", entry.Name())
```
roosterfish: `Failed to get absolute path to %q yaml`

Just for consistency, the other messages contain `yaml` at the end.
Maybe this check can even be put in a func as the same code is now used at three different places?

roosterfish (changes requested):

internal/trust/remotes.go
simondeziel: Right, makes sense for both consistency and code duplication, thanks!

roosterfish (approved):

PR #292, opened 2024-11-19
Add dep flags to TICS step

masnax:
Addresses the error shown here: https://canonical.tiobe.com/tiobeweb/TICS/AnnotatedSource.html#axes=DeltaDate(1729297017),Date(1731975417),ErrorSuppression(no),CodeType(Set(production,test)),Project(microcluster),Suppressions(no),DiffType(Set(new,unchanged)),File(Path(HIE,microcluster,master,internal,rest,resources,cluster.go))&diff=false&metrics=Annotations(CS)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -7,6 +7,10 @@ on:
     - cron: '0 0 * * *'  # Test TICS daily
 
 env:
+  CGO_CFLAGS: -I/home/runner/go/deps/dqlite/include/
+  CGO_LDFLAGS: -L/home/runner/go/deps/dqlite/.libs/
+  LD_LIBRARY_PATH: /home/runner/go/deps/dqlite/.libs/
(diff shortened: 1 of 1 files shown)


kadinsayani (dismissed):
LGTM - good catch!

.github/workflows/tests.yml
```diff
@@ -114,6 +114,11 @@ jobs:
     name: Tiobe TICS
     runs-on: ubuntu-22.04
     needs: code-tests
+    env:
```
roosterfish: Let's put them under [`env:`](https://github.com/canonical/microcluster/pull/292/files#diff-1db27d93186e46d3b441ece35801b244db8ee144ff1405ca27a163bfe878957fR9) at the root then the vars don't need to be duplicated and are present for all of the jobs.

roosterfish (changes requested):

kadinsayani (approved):

roosterfish (approved):

Issue #198, opened 2024-07-24
Allow draining connections of additional servers


roosterfish:
Additional servers added through `MicroCluster.AddServers()` might get restarted in case their configuration or certificate gets modified.

By default active connections will get cancelled so microcluster should grow support to allow draining active connections before restarting the underlying listener.

To enable this feature another field `DrainConnections bool` can be added to the `Server` struct.

Internally we can use https://pkg.go.dev/net/http#Server.Shutdown which automatically tries to drain active connections:

> Shutdown works by first closing all open listeners, then closing all idle connections, and then waiting indefinitely for connections to return to idle and then shut down

roosterfish:
Done with https://github.com/canonical/microcluster/pull/295

PR #295, opened 2024-11-29
Shutdown servers when resetting cluster membership

claudiubelu:
In the ``k8s-snap``, we're seeing a significant amount of ``EOF`` errors client-side, typically when operations such as ``k8s bootstrap`` or ``k8s join-cluster`` are run and an error occurs server-side.

When closing a listener, the server won't accept any new connections. However, we should still allow for the current connections / requests to finish. Calling ``server.Shutdown`` allows us to wait for them to finish. We'll be doing this only if ``drainConnectionsTimeout`` has been set for the server.

 Note that in the case we fail to bootstrap / join the cluster, we'll be  resetting a few things, including the cluster membership. This includes  the HTTPS and unix socket servers we have open by closing them.

 However, we cannot gracefully shutdown the servers, as there's at least  one connection that is still open: the bootstrap / join request. Forcing  the connection to close before we're able to write the request response  will result in the client getting an EOF error, and no information regarding the failure.

 Running the revert actions in a goroutine will address this issue: while  the revert happens, we'll be able to return and write the HTTP response  and then close the connection, finally allowing the Servers to gracefully  shutdown, and the clients to be happy. As a bonus, this will also  significantly reduce the amount of time spent by the bootstrap / join  requests in case of failures.

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -71,6 +71,10 @@ type Args struct {
 
 	// Each rest.Server will be initialized and managed by microcluster.
 	ExtensionServers map[string]rest.Server
+
+	// DrainConnectionsTimeout is the amount of time to allow for all core server connections to drain when shutting down.
+	// If it's 0, the connections are not drained when shutting down.
+	DrainConnectionsTimeout time.Duration
 }
 
 // Daemon holds information for the microcluster daemon.
@@ -106,6 +110,8 @@ type Daemon struct {
 
 	extensionServersMu sync.RWMutex
 	extensionServers   map[string]rest.Server
+
+	drainConnectionsTimeout time.Duration
 }
 
 // NewDaemon initializes the Daemon context and channels.
@@ -174,6 +180,7 @@ func (d *Daemon) Run(ctx context.Context, stateDir string, args Args) error {
 	}
 
 	d.version = args.Version
+	d.drainConnectionsTimeout = args.DrainConnectionsTimeout
 
 	// Setup the deamon's internal config.
 	d.config = internalConfig.NewDaemonConfig(filepath.Join(d.os.StateDir, "daemon.yaml"))
@@ -804,7 +811,7 @@ func (d *Daemon) UpdateServers() error {
 // startUnixServer starts up the core unix listener with the given resources.
 func (d *Daemon) startUnixServer(serverEndpoints []rest.Resources, socketGroup string) error {
 	ctlServer := d.initServer(serverEndpoints...)
-	ctl := endpoints.NewSocket(d.shutdownCtx, ctlServer, d.os.ControlSocket(), socketGroup)
+	ctl := endpoints.NewSocket(d.shutdownCtx, ctlServer, d.os.ControlSocket(), socketGroup, d.drainConnectionsTimeout)
 	d.endpoints = endpoints.NewEndpoints(d.shutdownCtx, map[string]endpoints.Endpoint{
 		endpoints.EndpointsUnix: ctl,
 	})
@@ -837,7 +844,7 @@ func (d *Daemon) addCoreServers(preInit bool, defaultURL api.URL, defaultCert *s
 	d.extensionServersMu.RUnlock()
 
 	server := d.initServer(serverEndpoints...)
-	network := endpoints.NewNetwork(d.shutdownCtx, endpoints.EndpointNetwork, server, defaultURL, defaultCert)
+	network := endpoints.NewNetwork(d.shutdownCtx, endpoints.EndpointNetwork, server, defaultURL, defaultCert, d.drainConnectionsTimeout)
 
 	return d.endpoints.Add(map[string]endpoints.Endpoint{
 		endpoints.EndpointsCore: network,
@@ -903,7 +910,7 @@ func (d *Daemon) addExtensionServers(preInit bool, fallbackCert *shared.CertInfo
 		}
 
 		server := d.initServer(extensionServer.Resources...)
-		network := endpoints.NewNetwork(d.shutdownCtx, endpoints.EndpointNetwork, server, *url, cert)
+		network := endpoints.NewNetwork(d.shutdownCtx, endpoints.EndpointNetwork, server, *url, cert, extensionServer.DrainConnectionsTimeout)
 		networks[serverName] = network
 	}
 
--- a/internal/endpoints/network.go
+++ b/internal/endpoints/network.go
@@ -2,11 +2,13 @@ package endpoints
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"net/http"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/canonical/lxd/lxd/endpoints/listeners"
 	"github.com/canonical/lxd/lxd/util"
@@ -27,10 +29,12 @@ type Network struct {
 
 	ctx    context.Context
 	cancel context.CancelFunc
+
+	drainConnectionsTimeout time.Duration
 }
 
 // NewNetwork assigns an address, certificate, and server to the Network.
-func NewNetwork(ctx context.Context, endpointType EndpointType, server *http.Server, address api.URL, cert *shared.CertInfo) *Network {
+func NewNetwork(ctx context.Context, endpointType EndpointType, server *http.Server, address api.URL, cert *shared.CertInfo, drainConnTimeout time.Duration) *Network {
 	ctx, cancel := context.WithCancel(ctx)
 
 	return &Network{
@@ -41,6 +45,8 @@ func NewNetwork(ctx context.Context, endpointType EndpointType, server *http.Ser
 		server: server,
 		ctx:    ctx,
 		cancel: cancel,
+
+		drainConnectionsTimeout: drainConnTimeout,
 	}
 }
 
@@ -108,27 +114,39 @@ func (n *Network) Serve() {
 		case <-n.ctx.Done():
 			logger.Infof("Received shutdown signal - aborting https socket server startup")
 		default:
+			// server.Serve always returns a non-nil error.
+			// http.ErrServerClosed is returned after server.Shutdown or server.Close.
+			// net.ErrClosed is returned if the listener is closed.
 			err := n.server.Serve(n.listener)
-			if err != nil {
-				select {
-				case <-n.ctx.Done():
-					logger.Infof("Received shutdown signal - aborting https socket server startup")
-				default:
-					logger.Error("Failed to start server", logger.Ctx{"err": err})
-				}
+			if errors.Is(err, http.ErrServerClosed) || errors.Is(err, net.ErrClosed) {
+				logger.Infof("Received shutdown signal - stopped serving https socket listener")
+			} else {
+				logger.Error("Failed to start server", logger.Ctx{"err": err})
 			}
 		}
 	}()
 }
 
-// Close the listener.
+// Close the listener and server.
+// Will attempt to close the server gracefully, if configured with a drain connections timeout.
+// Note that graceful shutdown will timeout if the connections do not finish (e.g.: a request caused the server
+// to Close the endpoints on the same goroutine).
 func (n *Network) Close() error {
 	if n.listener == nil {
 		return nil
 	}
 
 	logger.Info("Stopping REST API handler - closing https socket", logger.Ctx{"address": n.listener.Addr()})
-	n.cancel()
+	defer n.cancel()
 
-	return n.listener.Close()
+	// n.listener.Close() will mean that we'll no longer accept connections.
+	// It does not shutdown the server, or its currently accepted connections.
+	// We need to shut this down separately, as the listener is not passed to the server
+	// if n.Serve() is not called.
+	err := n.listener.Close()
+	if err != nil {
+		return err
+	}
+
+	return shutdownServer(n.server, n.drainConnectionsTimeout)
 }
--- a/internal/endpoints/socket.go
+++ b/internal/endpoints/socket.go
@@ -2,12 +2,14 @@ package endpoints
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"net/http"
 	"os"
 	"os/user"
 	"strconv"
+	"time"
 
 	"github.com/canonical/lxd/shared"
 	"github.com/canonical/lxd/shared/api"
@@ -24,10 +26,12 @@ type Socket struct {
 
 	ctx    context.Context
 	cancel context.CancelFunc
+
+	drainConnectionsTimeout time.Duration
 }
 
 // NewSocket returns a Socket struct with no listener attached yet.
-func NewSocket(ctx context.Context, server *http.Server, path api.URL, group string) *Socket {
+func NewSocket(ctx context.Context, server *http.Server, path api.URL, group string, drainConnTimeout time.Duration) *Socket {
 	ctx, cancel := context.WithCancel(ctx)
 	return &Socket{
 		Path:  path.Hostname(),
@@ -36,6 +40,8 @@ func NewSocket(ctx context.Context, server *http.Server, path api.URL, group str
 		server: server,
 		ctx:    ctx,
 		cancel: cancel,
+
+		drainConnectionsTimeout: drainConnTimeout,
 	}
 }
 
@@ -93,29 +99,41 @@ func (s *Socket) Serve() {
 		case <-s.ctx.Done():
 			logger.Infof("Received shutdown signal - aborting unix socket server startup")
 		default:
+			// server.Serve always returns a non-nil error.
+			// http.ErrServerClosed is returned after server.Shutdown or server.Close.
+			// net.ErrClosed is returned if the listener is closed.
 			err := s.server.Serve(s.listener)
-			if err != nil {
-				select {
-				case <-s.ctx.Done():
-					logger.Infof("Received shutdown signal - aborting unix socket server startup")
-				default:
-					logger.Error("Failed to start server", logger.Ctx{"err": err})
-				}
+			if errors.Is(err, http.ErrServerClosed) || errors.Is(err, net.ErrClosed) {
+				logger.Infof("Received shutdown signal - stopped serving unix socket listener")
+			} else {
+				logger.Error("Failed to start server", logger.Ctx{"err": err})
 			}
 		}
 	}()
 }
 
-// Close the Socket's listener.
+// Close the listener and server.
+// Will attempt to close the server gracefully, if configured with a drain connections timeout.
+// Note that graceful shutdown will timeout if the connections do not finish (e.g.: a request caused the server
+// to Close the endpoints on the same goroutine).
 func (s *Socket) Close() error {
 	if s.listener == nil {
 		return nil
 	}
 
 	logger.Info("Stopping REST API handler - closing socket", logger.Ctx{"socket": s.listener.Addr()})
-	s.cancel()
+	defer s.cancel()
 
-	return s.listener.Close()
+	// s.listener.Close() will mean that we'll no longer accept connections.
+	// It does not shutdown the server, or its currently accepted connections.
+	// We need to shut this down separately, as the listener is not passed to the server
+	// if s.Serve() is not called.
+	err := s.listener.Close()
+	if err != nil {
+		return err
+	}
+
+	return shutdownServer(s.server, s.drainConnectionsTimeout)
 }
 
 // Remove any stale socket file at the given path.
--- a/internal/endpoints/util.go
+++ b/internal/endpoints/util.go
@@ -1,0 +1,40 @@
+package endpoints
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net"
+	"net/http"
+	"time"
+
+	"github.com/canonical/lxd/shared/logger"
+)
+
+// shutdownServer will shutdown the given server.
+// If the given timeout is 0, it will forcefully shut it down. Otherwise, it will gracefully shut it down.
+func shutdownServer(server *http.Server, timeout time.Duration) error {
+	// If the given timeout is 0, force the shutdown.
+	if timeout == 0 {
+		err := server.Close()
+		if errors.Is(err, net.ErrClosed) {
+			return nil
+		}
+		return err
+	}
+
+	// server.Shutdown will gracefully stop the server, allowing existing requests to finish.
+	shutdownCtx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+	err := server.Shutdown(shutdownCtx)
+	if err != nil {
+		logger.Error("Failed to gracefully shutdown server", logger.Ctx{"err": err})
+		closeErr := server.Close()
+		if closeErr != nil {
+			logger.Error("Failed to close server", logger.Ctx{"err": closeErr})
+			return fmt.Errorf("Encountered error while closing server: %w, after failing to gracefully shutdown the server: %w", closeErr, err)
+		}
+		return err
+	}
+	return nil
+}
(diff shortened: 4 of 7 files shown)


roosterfish:
Hi, thanks for the PR. Have you seen https://github.com/canonical/microcluster/issues/198?
The idea is to add a new `DrainConnections` field to the [`Server`](https://github.com/canonical/microcluster/blob/v3/rest/rest.go#L48) struct which allows to configure what you are proposing in this PR.

Could you please make the functionality dependent on this field? 
You should be able to pass another flag into [`NewNetwork()`](https://github.com/canonical/microcluster/blob/v3/internal/daemon/daemon.go#L906) when creating the endpoint so that when calling `Close()` on it you have the necessary information available.

claudiubelu:
> Hi, thanks for the PR. Have you seen #198? The idea is to add a new `DrainConnections` field to the [`Server`](https://github.com/canonical/microcluster/blob/v3/rest/rest.go#L48) struct which allows to configure what you are proposing in this PR.
> 
> Could you please make the functionality dependent on this field? You should be able to pass another flag into [`NewNetwork()`](https://github.com/canonical/microcluster/blob/v3/internal/daemon/daemon.go#L906) when creating the endpoint so that when calling `Close()` on it you have the necessary information available.

Hmm, there might also be the question of how long to wait for connections to drain. Currently, we're waiting 5 seconds, but that could be configurable. We could have `DrainConnectionsTimeout`, and setting it to `0` would disable this aspect. Wdyt?

roosterfish:
> We could have `DrainConnectionsTimeout`, and setting it to `0` would disable this aspect. Wdyt?

Sounds good to me. By default (in case `DrainConnectionsTimeout` is 0) I would use the endpoints `ctx` which is present for both types `Network` and `Socket` when calling `Shutdown(ctx)` on the server. 
In your changes you currently use `context.Background()`. Only if `DrainConnectionsTimeout != 0` I would derive a new context with timeout from the endpoint's `ctx`.

claudiubelu:
> > We could have `DrainConnectionsTimeout`, and setting it to `0` would disable this aspect. Wdyt?
> 
> Sounds good to me. By default (in case `DrainConnectionsTimeout` is 0) I would use the endpoints `ctx` which is present for both types `Network` and `Socket` when calling `Shutdown(ctx)` on the server. In your changes you currently use `context.Background()`. Only if `DrainConnectionsTimeout != 0` I would derive a new context with timeout from the endpoint's `ctx`.

Added the `DrainConnectionsTimeout`, though it's been added in the `daemon.Args` as well, which is to be used for core servers.

Also, the original intention was to solve the `EOF` issue we've been seeing around, and I've included a fix for that as well.

rest/rest.go
```diff
@@ -65,4 +66,8 @@ type Server struct {
 
 	// Resources is the list of resources offered by this server.
 	Resources []Resources
+
+	// DrainConnectionsTimeout is the amount of time to allow for all connections to drain when shutting down.
+	// If it's 0, the connections are not drained when shutting down.
+	DrainConnectionsTimeout time.Duration
```
roosterfish, line 72: Thanks for omitting `DrainConnections` and make the decision based on whether or not a timeout is set.

internal/endpoints/util.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/canonical/lxd/shared/logger"
+)
+
+// shutdownServer will shutdown the given server.
+// If the given timeout is 0, it will forcefully shut it down. Otherwise, it will gracefully shut it down.
+func shutdownServer(ctx context.Context, server *http.Server, timeout time.Duration) error {
+	// If the given timeout is 0, force the shutdown.
+	if timeout == 0 {
+		err := server.Close()
+		if errors.Is(err, syscall.EINVAL) {
```
roosterfish: Can you please put a comment telling why the check is required? I assume in case you are trying to close an already closed connection.

internal/endpoints/util.go
```diff
@@ -0,0 +1,38 @@
...
+			return nil
+		}
+		return err
+	}
+
+	// server.Shutdown will gracefully stop the server, allowing existing requests to finish.
+	shutdownCtx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+	if err := server.Shutdown(shutdownCtx); err != nil {
```
roosterfish: We usually tend to split those lines for readability, e.g.

```go
err := server.Shutdown(shutdownCtx)
if err != nil {
```

roosterfish: Same for the `closeErr` below.

internal/daemon/daemon.go
```diff
@@ -135,6 +141,10 @@ func NewDaemon(project string) *Daemon {
 			if err != nil {
 				return err
 			}
+			err = d.endpoints.Shutdown()
```
roosterfish: Shouldn't we rather extend `Down()` instead of adding another `Shutdown()`? As we are already looping through the listeners in `Down()` you can also call `endpoint.ShutdownServer()` from there.

internal/daemon/daemon.go
```diff
@@ -1108,6 +1119,9 @@ func (d *Daemon) State() state.State {
 
 			return d.endpoints.Down()
 		},
+		ShutdownServers: func() error {
```
roosterfish: Same here, if we merge `Shutdown()` into `Down()` we also don't require another func on the state.

internal/rest/resources/control.go
```diff
@@ -73,7 +73,19 @@ func controlPost(state state.State, r *http.Request) response.Response {
...
+		// things, including the cluster membership. This includes the HTTPS and unix socket servers we have
+		// open by closing them.
+		// However, we cannot gracefully shutdown the servers, as there's at least one connection that is
+		// still open: the bootstrap / join request. Forcing the connection to close before we're able
+		// to write the request response will result in the client getting an EOF error, and no information
+		// regarding the failure.
+		// Running the revert actions in a goroutine will address this issue: while the revert happens,
+		// we'll be able to return and write the HTTP response and then close the connection, finally
+		// allowing the Servers to gracefully shutdown, and the clients to be happy.
+		go reverter.Fail()
```
roosterfish: Is this working just because the sending of the response is faster than the revert? I fear we require some form of synchronization so that it works consistently.

roosterfish: Can we do something like [this](https://github.com/canonical/microcluster/blob/v3/internal/rest/resources/cluster.go#L330) to flush the response beforehand?

roosterfish (changes requested):
Thank you for this.
Good call also allowing to set the timeout for the core listeners.

Please have a look at my few comments.

internal/rest/resources/control.go
claudiubelu: Indeed, we're doing this to allow the request to actually finish and the response to be written back before closing down the server. The ``reverter`` also includes shutting down the server, so we need to finish our request beforehand.

The `ManualResponse` bit from `clusterMemberPut` is interesting. Note that `resetClusterMember` is called before actually getting to write any response. If `resetClusterMember` would result in the servers to shutdown before getting to write the response, it would result in another EOF error. There is a way in which the `ManualResponse` may help... though it will be a bit odd: stop the servers in the `ManualResponse`.

internal/endpoints/util.go
claudiubelu: Done

internal/endpoints/util.go
claudiubelu: Indeed, but it should actually be ``net.ErrClosed``.

internal/daemon/daemon.go
claudiubelu: ditto

internal/daemon/daemon.go
claudiubelu: Indeed. That was my original implementation, as you can see from the fact that this separation was introduced in the second commit (will squash before merging).

However, the reason behind this is a bit complicated. First of all, `Down()` shutting down the listeners is a good idea, it allows us to no longer accept new connections, and it would be best to call it early in such cases. But we can't necessarily close down the servers just yet, as this may be called due to a request (e.g.: if bootstrap fails (aka during a request), we're "resetting" the cluster). Because this happens during a request, we cannot gracefully complete the request and shutdown the server on the same thread (or at least with how microcluster's code flows), which is why ``reverter.Fail()`` was goroutined. If we forcefully shutdown the servers while a connection is still in progress, the client will get an `EOF` and with no explanation of what actually happens (it doesn't get the error we're supposed to send back, which is the original reason this PR was sent).

We could go back to how I've done this previously, and having `Down()` also shutdown the servers as you suggest, but the same thread response + shutdown issue I've mentioned could still occur, like it does in `clusterMemberPut` (the listeners are stopped before writing the reply).

internal/endpoints/network.go
```diff
@@ -121,14 +126,24 @@ func (n *Network) Serve() {
...
+// Note that graceful shutdown will timeout if the connections do not finish (e.g.: a request caused the server
+// to Close the endpoints on the same goroutine).
 func (n *Network) Close() error {
 	if n.listener == nil {
 		return nil
 	}
 
 	logger.Info("Stopping REST API handler - closing https socket", logger.Ctx{"address": n.listener.Addr()})
-	n.cancel()
+	defer n.cancel()
```
roosterfish, line 140: Is there a reason for deferring the call to `cancel()`?

internal/endpoints/network.go
```diff
@@ -121,14 +126,24 @@ func (n *Network) Serve() {
...
+
+	// .Close() will mean that we'll no longer accept connections.
+	// It does not shutdown the server, or its currently accepted connections.
+	err := n.listener.Close()
```
roosterfish, line 146: That is an interesting observation. So before Microcluster just stopped the listener but the server (and active connections) where killed as soon as the main routine exited out?

From what I saw in the docs both `server.Shutdown()` and `server.Stop()` also close the listeners so isn't it enough to just go with what you have proposed in `shutdownServer`?

internal/rest/resources/cluster.go
```diff
@@ -370,6 +365,21 @@ func resetClusterMember(ctx context.Context, s state.State, force bool) (reExec
...
 		<-ctx.Done() // Wait until request has finished.
 
+		// NOTE(claudiub): In the case we fail to bootstrap / join the cluster, or if we remove the node
+		// from the cluster, we'll be resetting the node's cluster membership. This includes the HTTPS
+		// and unix socket servers we have open by closing them.
+		// However, we cannot gracefully shutdown the servers, as there's at least one connection that is
+		// still open: the bootstrap / join request. Forcing the connection to close before we're able
+		// to write the request response will result in the client getting an EOF error, and no information
+		// regarding the failure.
+		// Running the server shutdown actions in a goroutine will address this issue: while this action
```
roosterfish: The comment requires an update now.

roosterfish (changes requested):

internal/endpoints/network.go
claudiubelu: Yes. `n.cancel` refers to `n.ctx`'s cancel, but we're also passing `n.ctx` to `shutdownServer` below, which we'll use to gracefully shutdown the server. If we're canceling the context now, `server.Shutdown` will error out because the context was canceled:

```
Dec 11 22:53:41 ubuntu k8s.k8sd[162622]: time="2024-12-11T22:53:41Z" level=error msg="Failed to gracefully shutdown server" err="context canceled"
```

internal/endpoints/network.go
claudiubelu: Correct, which is why we saw the `EOF` errors, but those `EOF` errors mentioned were non-deterministic simply because it was a race between finishing up the request and writing the response and the main routine exiting out. If we forcefully close a connection before the reply is written, the client would get an `EOF` error.

And yeah, those operations will also close the listeners from the start, so we can just go with `shutdownServer`.

internal/rest/resources/cluster.go
claudiubelu: Done

internal/endpoints/network.go
claudiubelu: It turns out we **do** need to close the listener separately, as it is not passed to the server if `.Listen()` is not called, which would result in a leaked listener.

roosterfish: Ah so you mean if we don't call `endpoint.Serve()` the listener isn't tracked by the server so a `server.Stop()` wouldn't close the listeners that we have created separately?

internal/endpoints/network.go
roosterfish: Mh but the `defer` still gets executed before `shutdownServer` gets called below which means the context is already cancelled.

What about doing something like this:

```go
err = shutdownServer(n.ctx, n.server, n.drainConnectionsTimeout)
if err != nil {
  //
}

n.cancel()
```

This way you ensure the context is still intact when triggering the shutdown.

roosterfish (changes requested):
Can you please again squash the commits to clean up the history? Please also check my comment on the endpoint ctx.

claudiubelu:
> Can you please again squash the commits to clean up the history? Please also check my comment on the endpoint ctx.

Squashed the commits.

I won't change the `defer` part though, I think it's fine as it is. According to the golang documentation, "A defer statement defers the execution of a function until the surrounding function returns.", which means that the context gets canceled **after** the `Close` method returns. We have to cancel the context in case an errors occurs, and if no error occurs, we'd have to write something like this:

```
err = shutdownServer(n.ctx, n.server, n.drainConnectionsTimeout)
if err != nil {
  n.cancel()
  //
}

n.cancel()
```

We'd have that call duplicated, and it would have been worse if there were more error cases. Having a `defer` simplifies this. Plus, In a lot of official golang documentation defers are used, including examples with contexts: https://pkg.go.dev/context#WithTimeout

roosterfish:
Thanks for squashing.

> We'd have that call duplicated

That's a good point so let's keep the defer where it is

> A defer statement defers the execution of a function until the surrounding function returns.

This is right but doesn't account for the call to `shutdownServers` which happens afterwards. You can run this example which will always print `hello from ff2` before `hello from ff` which got deferred:

```go
ff := func() string {
    return "hello from ff"
}

ff2 := func() string {
    defer fmt.Println("hello from ff2")
    return ff()
}

fmt.Println(ff2())
```

I would propose to just run `shutdownServers` outside of the `return` to prevent this situation and just do `return nil` at the end.

roosterfish:
Sorry maybe I wasn't clear enough with the squashing. I saw your comment [here](https://github.com/canonical/microcluster/pull/295#discussion_r1881053858) and thought you will squash only the changes which are anyway reverted in one of the follow up commits.

We group the changes by package (indicated as prefix in the commit message). Meaning your changes to the daemon's config would be in a commit message like "internal/daemon: Adding DrainConnectionsTimeout field to config" and so on.

claudiubelu:
> Thanks for squashing.
> 
> > We'd have that call duplicated
> 
> That's a good point so let's keep the defer where it is
> 
> > A defer statement defers the execution of a function until the surrounding function returns.
> 
> This is right but doesn't account for the call to `shutdownServers` which happens afterwards. You can run this example which will always print `hello from ff2` before `hello from ff` which got deferred:
> 
> ```go
> ff := func() string {
>     return "hello from ff"
> }
> 
> ff2 := func() string {
>     defer fmt.Println("hello from ff2")
>     return ff()
> }
> 
> fmt.Println(ff2())
> ```
> 
> I would propose to just run `shutdownServers` outside of the `return` to prevent this situation and just do `return nil` at the end.

Yes, it does print `hello from ff2` before `hello from ff`, and that is to be expected, because the `defer`'s "surrounding function" mentioned in the golang documentation is actually the anonymous assigned to `ff2`, not the function in which `ff2` is defined.

> Sorry maybe I wasn't clear enough with the squashing. I saw your comment [here](https://github.com/canonical/microcluster/pull/295#discussion_r1881053858) and thought you will squash only the changes which are anyway reverted in one of the follow up commits.
> 
> We group the changes by package (indicated as prefix in the commit message). Meaning your changes to the daemon's config would be in a commit message like "internal/daemon: Adding DrainConnectionsTimeout field to config" and so on.

Done.

internal/endpoints/network.go
```diff
@@ -121,14 +126,26 @@ func (n *Network) Serve() {
...
 	}
 
 	logger.Info("Stopping REST API handler - closing https socket", logger.Ctx{"address": n.listener.Addr()})
-	n.cancel()
+	defer n.cancel()
+
+	// n.listener.Close() will mean that we'll no longer accept connections.
+	// It does not shutdown the server, or its currently accepted connections.
+	// We need to shut this down separately, as the listener is not passed to the server
+	// if s.Listen() is not called.
```
roosterfish: `if s.Serve() is not called`.

The listener gets passed to the server in `s.Serve()`.

internal/endpoints/socket.go
```diff
@@ -106,16 +111,28 @@ func (s *Socket) Serve() {
...
 	}
 
 	logger.Info("Stopping REST API handler - closing socket", logger.Ctx{"socket": s.listener.Addr()})
-	s.cancel()
+	defer s.cancel()
+
+	// s.listener.Close() will mean that we'll no longer accept connections.
+	// It does not shutdown the server, or its currently accepted connections.
+	// We need to shut this down separately, as the listener is not passed to the server
+	// if s.Listen() is not called.
```
roosterfish: `s.Serve()`.

internal/endpoints/util.go
```diff
@@ -0,0 +1,39 @@
...
+		return err
+	}
+
+	// server.Shutdown will gracefully stop the server, allowing existing requests to finish.
+	shutdownCtx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+	err := server.Shutdown(shutdownCtx)
+	if err != nil {
+		logger.Error("Failed to gracefully shutdown server", logger.Ctx{"err": err})
+		if closeErr := server.Close(); closeErr != nil {
```
roosterfish: Looks [this](https://github.com/canonical/microcluster/pull/295#discussion_r1879636542) was missed out.

roosterfish (changes requested):
You are right on the `defer`, my example doesn't make sense as it returns the value which changes the effective order of the prints. Forget about it :)

Almost there, just two smaller comments.

internal/endpoints/network.go
claudiubelu: indeed, my bad.

internal/endpoints/network.go
claudiubelu: Correct.

internal/endpoints/socket.go
claudiubelu: Done

internal/endpoints/util.go
claudiubelu: Done

internal/daemon/daemon.go
roosterfish: If I understand it right this is now fixed as you run StopListeners within the reExec go routine?

internal/rest/resources/control.go
roosterfish: Same as https://github.com/canonical/microcluster/pull/295#discussion_r1887018930

claudiubelu:
Seeing these errors in the logs now:

```
Dec 16 23:14:13 ubuntu k8s.k8sd[1038456]: time="2024-12-16T23:14:13Z" level=info msg="Stopping REST API handler - closing https socket" address="172.17.201.20:6400"
Dec 16 23:14:13 ubuntu k8s.k8sd[1038456]: time="2024-12-16T23:14:13Z" level=error msg="Failed to start server" err="accept tcp 172.17.201.20:6400: use of closed network connection"
Dec 16 23:14:13 ubuntu k8s.k8sd[1038456]: time="2024-12-16T23:14:13Z" level=info msg="Stopping REST API handler - closing socket" socket=/var/snap/k8s/common/var/lib/k8sd/state/control.socket
Dec 16 23:14:13 ubuntu k8s.k8sd[1038456]: time="2024-12-16T23:14:13Z" level=error msg="Failed to start server" err="accept unix /var/snap/k8s/common/var/lib/k8sd/state/control.socket: use of closed network connection"
```

Amended `Serve` from `socket.go` and `network.go` to address this issue. Could you review those changes? The logs are now (tested on a cherry-pick to v2 branch for k8s-snap):

```
Dec 17 09:12:30 ubuntu k8s.k8sd[1159247]: time="2024-12-17T09:12:30Z" level=info msg="Stopping REST API handler - closing socket" socket=/var/snap/k8s/common/var/lib/k8sd/state/control.socket
Dec 17 09:12:30 ubuntu k8s.k8sd[1159247]: time="2024-12-17T09:12:30Z" level=info msg="Received shutdown signal - stopped serving unix socket listener"
Dec 17 09:12:30 ubuntu k8s.k8sd[1159247]: time="2024-12-17T09:12:30Z" level=info msg="Stopping REST API handler - closing https socket" address="172.17.201.20:6400"
Dec 17 09:12:30 ubuntu k8s.k8sd[1159247]: time="2024-12-17T09:12:30Z" level=info msg="Received shutdown signal - stopped serving https socket listener"
```

roosterfish:
> Could you review those changes?
 
Hi, I have ran some extended tests and it looks like we face two other issue here. 

1) In case you set `DrainConnectionsTimeout` on one of the listeners (core or extended), if you CTRL+C (`kill -SIGINT <pid>`) any of the daemons, you will see the following error:

```
ERROR  [2024-12-18T16:30:09+01:00] Failed to gracefully shutdown server          err="context canceled"
```

This is because as soon as the daemon receives the signal, it cancels its main `shutdownCtx` context. 
As the endpoints (socket/network) context is derived from this one, it gets cancelled too which means the call to `server.Shutdown()` can never wait and errors out immediately.

So we have to detach `shutdownServer` from the endpoint's context to allow waiting for x seconds to drain the connections.

2) We should not return the [`err`](https://github.com/canonical/microcluster/pull/295/files#diff-34c26aa9a2a857700db2ae6c687e4156109197dbe1de7611d2ef5aeecccf8389R37) from `Shutdown` back to the caller. Otherwise you might see errors like `Error: Daemon stopped with error: context canceled` which are misleading as only the draining of the connections timed out.
But returning the error from the fallback `Close()` seems legit as this should work and returning the error in case it fails can be valuable.

claudiubelu:
> > Could you review those changes?
> 
> Hi, I have ran some extended tests and it looks like we face two other issue here.
> 
>     1. In case you set `DrainConnectionsTimeout` on one of the listeners (core or extended), if you CTRL+C (`kill -SIGINT <pid>`) any of the daemons, you will see the following error:
> 
> 
> ```
> ERROR  [2024-12-18T16:30:09+01:00] Failed to gracefully shutdown server          err="context canceled"
> ```
> 
> This is because as soon as the daemon receives the signal, it cancels its main `shutdownCtx` context. As the endpoints (socket/network) context is derived from this one, it gets cancelled too which means the call to `server.Shutdown()` can never wait and errors out immediately.
> 
> So we have to detach `shutdownServer` from the endpoint's context to allow waiting for x seconds to drain the connections.

Right. I initially had `context.Background` used in this case. Amended commit.

> 
>     2. We should not return the [`err`](https://github.com/canonical/microcluster/pull/295/files#diff-34c26aa9a2a857700db2ae6c687e4156109197dbe1de7611d2ef5aeecccf8389R37) from `Shutdown` back to the caller. Otherwise you might see errors like `Error: Daemon stopped with error: context canceled` which are misleading as only the draining of the connections timed out.

If we are going to use a separated context (`context.Background`), we shouldn't be seeing this type of error (`context canceled`) (which is not related to the operation timing out). Instead, if the operation times out (`context deadline exceeded`), or any other error occurs, it's still something we should return, handle, and log accordingly. If an error occurs, I would want to see it at least.

roosterfish (approved):
Thanks for the efforts, LGTM!

roosterfish:
@claudiubelu wanted to merge now but saw the commits aren't signed. Could you please sign them?

claudiubelu:
> @claudiubelu wanted to merge now but saw the commits aren't signed. Could you please sign them?

Done, Sorry, I forgot to sign them. :)

PR #296, opened 2024-12-01
Fixes wrapped nil lastErr issue during cluster join

claudiubelu:
When joining a cluster, we have a list of join addresses we try to connect to. However, ``lastErr`` is only set at the end of a loop, which can lead to confusing error messages like this:

```
The error was: failed to POST /k8sd/cluster/join: failed to join k8sd cluster as control plane: 1 join attempts were unsuccessful. Last error: %!w(<nil>)
```

This can happen if other errors can occur before lastErr is set, skipping the current join address. Additionally, ``lastErr`` may be set from the previous join address checks, and not from the current loop, which can be misleading.

 We're now setting ``lastErr`` appropriately. Additionally, if the  certificate fingerprint does not match, we're returning a new error.
 
 Signed-off-by: Claudiu Belu <claudiu.belu@canonical.com>

### DIFF
--- a/internal/rest/resources/control.go
+++ b/internal/rest/resources/control.go
@@ -218,13 +218,13 @@ func joinWithToken(state state.State, r *http.Request, req *internalTypes.Contro
 		cert, err := shared.GetRemoteCertificate(url.String(), "")
 		if err != nil {
 			logger.Warn("Failed to get certificate of cluster member", logger.Ctx{"address": url.String(), "error": err})
+			lastErr = err
 			continue
 		}
 
 		fingerprint := shared.CertFingerprint(cert)
 		if fingerprint != token.Fingerprint {
-			logger.Warn("Cluster certificate token does not match that of cluster member", logger.Ctx{"address": url.String(), "fingerprint": fingerprint, "expected": token.Fingerprint})
-			continue
+			return nil, fmt.Errorf("Cluster certificate token does not match that of cluster member. Expected: %q, actual: %q", fingerprint, token.Fingerprint)
 		}
 
 		d, err := internalClient.New(*url, state.ServerCert(), cert, false)


claudiubelu:
New sample with this code:

```
Joining the cluster. This may take a few seconds, please wait.
Error: Failed to join the cluster using the provided token.

The error was: failed to POST /k8sd/cluster/join: failed to join k8sd cluster as control plane: 1 join attempts were unsuccessful. Last error: Get "https://172.17.201.20:6400": Unable to connect to: 172.17.201.20:6400 ([dial tcp 172.17.201.20:6400: connect: connection refused])
```

internal/rest/resources/control.go
```diff
@@ -210,20 +211,23 @@ func joinWithToken(state state.State, r *http.Request, req *internalTypes.Contro
...
 	// Get a client to the target address.
-	var lastErr error
+	var allErrors []error
 	var joinInfo *internalTypes.TokenResponse
 	for _, addr := range token.JoinAddresses {
 		url := api.NewURL().Scheme("https").Host(addr.String())
 
 		cert, err := shared.GetRemoteCertificate(url.String(), "")
 		if err != nil {
 			logger.Warn("Failed to get certificate of cluster member", logger.Ctx{"address": url.String(), "error": err})
```
roosterfish, line 220: IIRC this is only logged as a warning because the join token contains multiple addresses and since we can try the other addresses too it's not considered to be an error that should be returned to the caller if one of the addresses cannot be reached as the join itself can still succeed with any of the remaining addresses.

internal/rest/resources/control.go
```diff
@@ -210,20 +211,23 @@ func joinWithToken(state state.State, r *http.Request, req *internalTypes.Contro
...
+			allErrors = append(allErrors, err)
 			continue
 		}
 
 		fingerprint := shared.CertFingerprint(cert)
 		if fingerprint != token.Fingerprint {
 			logger.Warn("Cluster certificate token does not match that of cluster member", logger.Ctx{"address": url.String(), "fingerprint": fingerprint, "expected": token.Fingerprint})
```
roosterfish: @masnax can we consider this a break in the protocol if one of the remote cluster members returns a different certificate? Because then I would argue we can safely do `return nil, <err>` as something is off with the remote member.

internal/rest/resources/control.go
```diff
@@ -238,11 +242,11 @@ func joinWithToken(state state.State, r *http.Request, req *internalTypes.Contro
 		}
 
 		logger.Error("Unable to complete cluster join request", logger.Ctx{"address": addr.String(), "error": err})
-		lastErr = err
+		allErrors = append(allErrors, err)
 	}
 
 	if joinInfo == nil {
-		return nil, fmt.Errorf("%d join attempts were unsuccessful. Last error: %w", len(token.JoinAddresses), lastErr)
```
roosterfish: I would vote against `errors.Join` as it simply concatenates the errors using a newline which looks odd in my opinion as we use the pattern `Error: Error more specific: Error more specific`.

Overall I think the idea is to only show the actual error from the join process itself. Let's consider `shared.GetRemoteCertificate` returns an error three times for all the join tokens addresses. It prints three warnings but `joinInfo` is still set to `nil`.
So probably its best to populate `lastErr` every time we log a warning.

internal/rest/resources/control.go
roosterfish: Same as if `internalClient.New` below returns an error.

roosterfish (changes requested):
Hi, thanks for raising the issue with `Last error: %!w(<nil>)`. 
Please have a look at my few comments.

internal/rest/resources/control.go
claudiubelu: Right, which is why we're not returning an error here, we're simply adding it to the list of ``allErrors``, which is eventually joined, wrapped, and returned if the join failed for all the available addresses (`if joinInfo == nil`).

If the join action was successful, this warning and error will be discarded anyways, but if it failed, I'd say this bit of information is still more useful to a user rather than a confusing `Last error: %!w(<nil>)`, and if we failed to join through all the join addresses, it will still be useful to know what happened for each of them, letting the user know directly through the error we return, rather than having the user scour through the service logs.

internal/rest/resources/control.go
claudiubelu: Well, ``errors.Join``, doesn't simply concatenate the error messages, ``errors.Is`` can still be used on the resulting error to match any of its components. But ultimately, it won't be relevant after marshaling the error.

internal/rest/resources/control.go
```diff
@@ -218,12 +218,14 @@ func joinWithToken(state state.State, r *http.Request, req *internalTypes.Contro
...
 		if err != nil {
 			logger.Warn("Failed to get certificate of cluster member", logger.Ctx{"address": url.String(), "error": err})
+			lastErr = err
 			continue
 		}
 
 		fingerprint := shared.CertFingerprint(cert)
 		if fingerprint != token.Fingerprint {
 			logger.Warn("Cluster certificate token does not match that of cluster member", logger.Ctx{"address": url.String(), "fingerprint": fingerprint, "expected": token.Fingerprint})
+			lastErr := fmt.Errorf("Cluster certificate token does not match that of cluster member. Expected: '%s', actual: '%s'", fingerprint, token.Fingerprint)
```
roosterfish: Looks this should be `lastErr =` to use the outer var instead.

roosterfish: This is currently causing some test failures https://github.com/canonical/microcluster/actions/runs/12120340935/job/33788817696?pr=296

internal/rest/resources/control.go
masnax: Indeed, I can't think of a situation where the cluster certificate doesn't match on one member, but might on another.

masnax: @claudiubelu If you could please replace this block with an `return nil, fmt.Error...` then we should be able to backport this into `v2` and get a new release out tomorrow.

internal/rest/resources/control.go
roosterfish: Thanks for adding the error assignment.

internal/rest/resources/control.go
```diff
@@ -218,12 +218,14 @@ func joinWithToken(state state.State, r *http.Request, req *internalTypes.Contro
...
 		if err != nil {
 			logger.Warn("Failed to get certificate of cluster member", logger.Ctx{"address": url.String(), "error": err})
+			lastErr = err
 			continue
 		}
 
 		fingerprint := shared.CertFingerprint(cert)
 		if fingerprint != token.Fingerprint {
 			logger.Warn("Cluster certificate token does not match that of cluster member", logger.Ctx{"address": url.String(), "fingerprint": fingerprint, "expected": token.Fingerprint})
+			lastErr = fmt.Errorf("Cluster certificate token does not match that of cluster member. Expected: '%s', actual: '%s'", fingerprint, token.Fingerprint)
```
roosterfish: Instead of `'%s'` you can just do `%q` which will take care of the quoting.

internal/rest/resources/control.go
roosterfish: I see what you are saying but we usually don't have multiline errors and tend to have everything in a single line.

roosterfish:
Also please prefix your commits with the packages name, see the commits from [this](https://github.com/canonical/microcluster/pull/269/commits) PR as an example.

roosterfish:
@claudiubelu almost there, can you please squash the changes so we can keep a clean history? We don't perform any squashing when merging the PR so we usually prevent adding new code in one commit (first commit) and then removing it a few commits after.

roosterfish (approved):
Thanks, LGTM!

roosterfish:
@claudiubelu I cannot merge the PR before the commit is signed.

claudiubelu:
> @claudiubelu I cannot merge the PR before the commit is signed.

Done

roosterfish (approved):

PR #301, opened 2024-12-10
internal/rest/resources: Return joinInfo even on error

claudiubelu:
When joining a cluster, if any error occurs after becoming a cluster member, we're not removing the current node from the cluster, even though we should.

That is because in ``joinWithToken`` after calling ``internalClient.AddClusterMember`` and getting a ``joinInfo`` object, we may still return ``nil`` afterwards in case of any error. Because of this, the ``controlPost`` ``reverter`` won't have any ``joinInfo`` object to use for removing itself from the cluster.

Because of this issue, if the cluster join operations fails, that node may no longer be able to join the cluster, as it appears to have already joined the cluster.

This commit addresses this issue by returning the ``joinInfo`` object after it exists.

### DIFF
--- a/internal/rest/resources/control.go
+++ b/internal/rest/resources/control.go
@@ -151,18 +151,21 @@ func controlPost(state state.State, r *http.Request) response.Response {
 		}
 	}
 
+	var joinAddrs []string
+	var localClusterMember *trust.Remote
 	if req.JoinToken != "" {
-		joinInfo, err = joinWithToken(state, r, req)
+		joinInfo, localClusterMember, err = joinWithToken(state, r, req)
 		if err != nil {
 			return response.SmartError(err)
 		}
 
-		reverter.Success()
-
-		return response.EmptySyncResponse
+		joinAddrs, err = setupLocalMember(state, localClusterMember, joinInfo)
+		if err != nil {
+			return response.SmartError(err)
+		}
 	}
 
-	err = intState.StartAPI(r.Context(), req.Bootstrap, req.InitConfig)
+	err = intState.StartAPI(r.Context(), req.Bootstrap, req.InitConfig, joinAddrs...)
 	if err != nil {
 		return response.SmartError(err)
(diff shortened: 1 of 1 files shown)


masnax (changes requested):
Thanks for catching this! 

One small maintainability improvement I'd suggest is rather than returning the struct in the error blocks, which we will have to remember to do if we ever want to extend this function, can you instead split `joinWithToken` into two parts? 

The first function can return `joinInfo` to the caller, and the second function (beginning with the call to `WriteCert`) can be called right afterward, taking joinInfo as an argument.

claudiubelu:
> Thanks for catching this!
> 
> One small maintainability improvement I'd suggest is rather than returning the struct in the error blocks, which we will have to remember to do if we ever want to extend this function, can you instead split `joinWithToken` into two parts?
> 
> The first function can return `joinInfo` to the caller, and the second function (beginning with the call to `WriteCert`) can be called right afterward, taking joinInfo as an argument.

Done

claudiubelu:
> Thanks for catching this!
> 
> One small maintainability improvement I'd suggest is rather than returning the struct in the error blocks, which we will have to remember to do if we ever want to extend this function, can you instead split `joinWithToken` into two parts?
> 
> The first function can return `joinInfo` to the caller, and the second function (beginning with the call to `WriteCert`) can be called right afterward, taking joinInfo as an argument.

Done

bschimke95:
@masnax is any additional change required for this to land? Otherwise, I'd love to see this PR merged.

masnax (approved):
@bschimke95 Nope, lgtm, thanks @claudiubelu for the fix!

masnax:
@claudiubelu Seems you haven't properly ssh/gpg signed your commits. Can you please do so?

claudiubelu:
> @claudiubelu Seems you haven't properly ssh/gpg signed your commits. Can you please do so?

done

PR #304, opened 2024-12-18
github: update of .jira_sync_config.yaml

mseralessandri:
I  changed the 'jira_key' and 'components' in .jira_sync_config.yaml configuration file for the synchronization with Jira.
The issues in MicroCluster will be created in a separated epic and the 'component' key will allow filtering by 'MicroCluster' key

### DIFF
--- a/.github/.jira_sync_config.yaml
+++ b/.github/.jira_sync_config.yaml
@@ -9,9 +9,8 @@ settings:
     
   # (Optional) Jira project components that should be attached to the created issue
   # Component names are case-sensitive
-  # components:
-  #  - LXD
-  #  - Microcloud
+  components:
+    - MicroCluster
       
   # (Optional) GitHub labels. Only issues with one of those labels will be synchronized.
   # If not specified, all issues will be synchronized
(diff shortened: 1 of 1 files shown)


.github/.jira_sync_config.yaml
```diff
@@ -28,12 +27,12 @@ settings:
   sync_comments: false
   
   # (Optional) (Default: None) Parent Epic key to link the issue to
-  epic_key: "LXD-1251"
+  epic_key: "LXD-2467"
```
roosterfish, line 30: Will the previously synced stories be moved into the new epic?

mseralessandri: Unfortunately not, I need to change the parent in Jira. The synchronization however works even if you move the issues in another epic.

roosterfish: Ok thanks!

roosterfish (approved):

roosterfish:
@mseralessandri can all of those bug sync epics be linked under a central epic (or another type) so they can be found easily in Jira? Just an idea.

mseralessandri:
> @mseralessandri can all of those bug sync epics be linked under a central epic (or another type) so they can be found easily in Jira? Just an idea.

You can link the epics only to objectives but then they would appear in the road map

PR #314, opened 2025-01-13
internal/recover: Use core/ prefixed endpoint to check online state

masnax:
Using just `/1.0` may not work as the upstream project might not have defined it (like in the example package) so it will always return 404


MggMuggins (approved):
Nice catch

roosterfish (approved):

PR #319, opened 2025-01-22
internal/db: Set database status based on initialization state

bschimke95:
Previously, the initial dqlite database status was always set to "DatabaseNotReady", regardless of whether the node was already bootstrapped or joined. As a result, any REST API call during microcluster startup would incorrectly return a 503 error with "Database not yet initialized", even if the node was already bootstrapped and simply restarting.

This commit refactors the handling of the database initialization state:

* Use the database status to determine if a cluster is
      bootstrapped/joined or uninitialized
* Sets the initial database status based on whether the database is already initialized.

With this change:

* Bootstrapped nodes now return "Database still starting" during startup.
* Non-bootstrapped nodes return "Database not yet initialized".

This provides more accurate status reporting and improves API responses during service restarts.

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -285,7 +285,10 @@ func (d *Daemon) init(listenAddress string, socketGroup string, heartbeatInterva
 		return fmt.Errorf("Failed to initialize trust store: %w", err)
 	}
 
-	d.db = db.NewDB(d.shutdownCtx, d.ServerCert, d.ClusterCert, d.Name, d.os, heartbeatInterval)
+	d.db, err = db.NewDB(d.shutdownCtx, d.ServerCert, d.ClusterCert, d.Name, d.os, heartbeatInterval)
+	if err != nil {
+		return fmt.Errorf("Failed to initialize database: %w", err)
+	}
 
 	listenAddr := api.NewURL()
 	if listenAddress != "" {
@@ -343,9 +346,15 @@ func (d *Daemon) init(listenAddress string, socketGroup string, heartbeatInterva
 
 	d.db.SetSchema(schemaExtensions, d.Extensions)
 
-	err = d.reloadIfBootstrapped()
-	if err != nil {
-		return err
+	status := d.db.Status()
+	if status == types.DatabaseStarting {
+		// Database is already bootstrapped, reload the daemon to ensure the latest configuration is applied.
+		err := d.reload()
+		if err != nil {
+			return fmt.Errorf("Failed to reload daemon: %w", err)
+		}
+	} else if status == types.DatabaseNotReady {
+		logger.Warn("Microcluster database is uninitialized")
 	}
 
 	err = d.trustStore.Refresh()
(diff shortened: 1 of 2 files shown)


internal/daemon/daemon.go
```diff
@@ -343,9 +346,16 @@ func (d *Daemon) init(listenAddress string, socketGroup string, heartbeatInterva
 
 	d.db.SetSchema(schemaExtensions, d.Extensions)
 
-	err = d.reloadIfBootstrapped()
+	bootstrapped, err := d.db.IsInitialized()
 	if err != nil {
-		return err
+		return fmt.Errorf("Failed to check if daemon is bootstrapped: %w", err)
```
roosterfish: `Failed to check if daemon is initialized`

To be consistent with the message below "Failed to check if database is initialized".

roosterfish: And should the variable rather be called `initialized :=`?

internal/db/dqlite.go
```diff
@@ -113,6 +124,27 @@ func (db *DqliteDB) SchemaVersion() (versionInternal uint64, versionExternal uin
 	return db.schema.Version()
 }
 
+// IsInitialized checks if the database is already bootstrapped or joined to a cluster.
+func (db *DqliteDB) IsInitialized() (bool, error) {
+	if _, err := os.Stat(filepath.Join(db.os.DatabaseDir, "info.yaml")); err != nil {
+		if os.IsNotExist(err) {
+			logger.Warn("microcluster database is uninitialized")
```
roosterfish: In case of `reloadIfBootstrapped` those warnings made sense as they were fired as part of the daemon's `init()`.

Now those would appear every time `IsInitialized` is consumed. 
Instead can they be returned by `IsInitialized` together with `false` so that they can be logged within the `init()` in case `if !bootstrapped {}`?

roosterfish (changes requested):
Thanks for this. A few minor nits for consistency.

To be more precise with the effect of this change, basically every API request to the Unix socket endpoints (as well as core and extension) will return the "not initialized" error until `StartAPI->StartWithCluster->Join` is called which will set the DB's status to "starting" for a very short period of time and then switches to its final running state.

internal/db/dqlite.go
bschimke95: Fair point, as discussed offline I ended up querying the DatabaseStatus instead of maintaining two functions which both basically expose the state of the database.

internal/db/dqlite.go
```diff
@@ -113,6 +124,26 @@ func (db *DqliteDB) SchemaVersion() (versionInternal uint64, versionExternal uin
...
+// This is an internal helper function; external callers should use Status() instead.
+func (db *DqliteDB) isInitialized() (bool, error) {
+	if _, err := os.Stat(filepath.Join(db.os.DatabaseDir, "info.yaml")); err != nil {
+		if os.IsNotExist(err) {
+			return false, nil
+		}
+		return false, err
+	}
+
+	if _, err := os.Stat(filepath.Join(db.os.StateDir, "daemon.yaml")); err != nil {
```
roosterfish: If I understood @masnax right this block is not required and the check for dqlite's `info.yaml` is already enough to identify if the node is bootstrapped/joined.

internal/daemon/daemon.go
```diff
@@ -343,9 +346,14 @@ func (d *Daemon) init(listenAddress string, socketGroup string, heartbeatInterva
 
 	d.db.SetSchema(schemaExtensions, d.Extensions)
 
-	err = d.reloadIfBootstrapped()
-	if err != nil {
-		return err
+	status := d.db.Status()
+	if status == types.DatabaseStarting {
+		// Database is already bootstrapped, reload the daemon to ensure the latest configuration is applied.
+		if err := d.reload(); err != nil {
```
roosterfish: Can you please put the `err != nil` in the next line following the standard we established across the repo?

roosterfish: Also please start the error message with uppercase `Failed...`.
Just whilst you are refactoring the code, can you please also let the warning start with an uppercase letter?

roosterfish (changes requested):
Just a few more minor comments, the change itself looks great.

internal/daemon/daemon.go
bschimke95: Ah, force of habit from our code-base :)

internal/db/dqlite.go
bschimke95: removed

roosterfish (approved):
Thanks!

PR #320, opened 2025-01-24
Backport: internal/db: Set database status based on initialization state

bschimke95:
Previously, the initial dqlite database status was always set to "DatabaseNotReady", regardless of whether the node was already bootstrapped or joined. As a result, any REST API call during microcluster startup would incorrectly return a 503 error with "Database not yet initialized", even if the node was already bootstrapped and simply restarting.

This commit refactors the handling of the database initialization state:

* Use the database status to determine if a cluster is bootstrapped/joined or uninitialized
* Sets the initial database status based on whether the database is already initialized.

With this change:

* Bootstrapped nodes now return "Database still starting" during startup.
* Non-bootstrapped nodes return "Database not yet initialized".

This provides more accurate status reporting and improves API responses during service restarts.

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -287,7 +287,10 @@ func (d *Daemon) init(listenAddress string, socketGroup string, heartbeatInterva
 		return fmt.Errorf("Failed to initialize trust store: %w", err)
 	}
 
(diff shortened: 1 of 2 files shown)


roosterfish (approved):
Thanks!

PR #322, opened 2025-01-31
Rest: Handle IPv6 link local addresses with zone identifier in Authenticate

### LINKED ISSUE #268: Cannot start a cluster on IPv6 link local scoped addresses
This affects microcluster `v2` and `v3`. 
Trying to start a cluster on IPv6 link local scoped addresses is not working as expected and does not report any errors either.

See this reproducer using the example package. Use any of your link local addresses and modify the scope (interface):

```bash
#!/bin/bash

rm -rf /tmp/m*

microd --state-dir /tmp/m1 &
microctl --state-dir /tmp/m1 waitready
microd --state-dir /tmp/m2 &
microctl --state-dir /tmp/m2 waitready
microd --state-dir /tmp/m3 &
microctl --state-dir /tmp/m3 waitready

microctl --state-dir /tmp/m1 init m1 [fe80::173d:4cbe:eb86:c0a9%wlp0s20f3]:9001 --bootstrap

# Fake custom cert
cp /tmp/m1/cluster.crt /tmp/m1/certificates/foo.crt
cp /tmp/m1/cluster.key /tmp/m1/certificates/foo.key
cp /tmp/m1/cluster.crt /tmp/m1/certificates/foo2.crt
cp /tmp/m1/cluster.key /tmp/m1/certificates/foo2.key


# Join additional members
token=$(microctl --state-dir /tmp/m1 tokens add m2)
microctl --state-dir /tmp/m2 init m2 [fe80::173d:4cbe:eb86:c0a9%wlp0s20f3]:9002 --token "$token"
token=$(microctl --state-dir /tmp/m1 tokens add m3)
microctl --state-dir /tmp/m3 init m3 [fe80::173d:4cbe:eb86:c0a9%wlp0s20f3]:9003 --token "$token"

microctl --state-dir /tmp/m1 cluster list

```

rapour:
This PR suggests a fix for #268 

Closes https://github.com/canonical/microcluster/issues/268

### DIFF
--- a/internal/db/dqlite.go
+++ b/internal/db/dqlite.go
@@ -379,6 +379,11 @@ func (db *DqliteDB) heartbeat(leaderInfo dqliteClient.NodeInfo, servers []dqlite
 
 // dqliteNetworkDial creates a connection to the internal database endpoint.
 func dqliteNetworkDial(ctx context.Context, addr string, db *DqliteDB) (net.Conn, error) {
+	addrPort, err := types.ParseAddrPort(addr)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to parse the address: %w", err)
+	}
+
 	peerCert, err := db.clusterCert().PublicKeyX509()
 	if err != nil {
 		return nil, err
@@ -396,13 +401,13 @@ func dqliteNetworkDial(ctx context.Context, addr string, db *DqliteDB) (net.Conn
 		ProtoMajor: 1,
 		ProtoMinor: 1,
 		Header:     make(http.Header),
-		Host:       addr,
+		Host:       addrPort.String(),
 	}
 
-	path := fmt.Sprintf("https://%s/%s/%s", addr, internalTypes.InternalEndpoint, "database")
-	request.URL, err = url.Parse(path)
-	if err != nil {
-		return nil, err
+	request.URL = &url.URL{
+		Scheme: "https",
+		Host:   addrPort.String(),
+		Path:   fmt.Sprintf("/%s/%s", internalTypes.InternalEndpoint, "database"),
 	}
 
 	request.Header.Set("Upgrade", "dqlite")
--- a/rest/types/addrport.go
+++ b/rest/types/addrport.go
@@ -85,6 +85,15 @@ func (a *AddrPort) UnmarshalYAML(unmarshal func(v any) error) error {
 	return nil
 }
 
+// WithZone returns an AddrPort that's the same as the receiver AddrPort but with the provided zone. If zone is empty,
+// the zone is removed. If IP of AddrPort is an IPv4 address, WithZone is a no-op and returns AddrPort unchanged.
+func (a AddrPort) WithZone(zone string) AddrPort {
+	return AddrPort{AddrPort: netip.AddrPortFrom(
+		a.Addr().WithZone(zone),
+		a.Port(),
+	)}
+}
+
 // String returns the string representation of the AddrPort, or an empty string if it is empty.
 func (a AddrPort) String() string {
 	if a == (AddrPort{}) {
--- a/rest/types/addrport_test.go
+++ b/rest/types/addrport_test.go
@@ -1,0 +1,96 @@
+package types
+
+import (
+	"net/netip"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestAddrPortWithZone(t *testing.T) {
+	cases := []struct {
+		name     string
+		addrPort AddrPort
+		zone     string
+		want     AddrPort
+	}{
+		{
+			name: "IPv4",
+			addrPort: AddrPort{AddrPort: netip.AddrPortFrom(
+				netip.MustParseAddr("10.0.0.10"),
+				9090,
+			)},
+			zone: "",
+			want: AddrPort{AddrPort: netip.AddrPortFrom(
+				netip.MustParseAddr("10.0.0.10"),
+				9090,
+			)},
+		},
+		{
+			name: "IPv4WithAddedZone",
+			addrPort: AddrPort{AddrPort: netip.AddrPortFrom(
+				netip.MustParseAddr("10.0.0.10"),
+				9090,
+			)},
+			zone: "eth0",
+			want: AddrPort{AddrPort: netip.AddrPortFrom(
+				netip.MustParseAddr("10.0.0.10"),
+				9090,
+			)},
+		},
+		{
+			name: "IPv6",
+			addrPort: AddrPort{AddrPort: netip.AddrPortFrom(
+				netip.MustParseAddr("fe80::b73c:ca5c:8035:23dc"),
+				9090,
+			)},
+			zone: "",
+			want: AddrPort{AddrPort: netip.AddrPortFrom(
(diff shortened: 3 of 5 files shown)


internal/db/dqlite.go
```diff
@@ -397,7 +398,7 @@ func dqliteNetworkDial(ctx context.Context, addr string, db *DqliteDB) (net.Conn
 		Host:       addr,
 	}
 
-	path := fmt.Sprintf("https://%s/%s/%s", addr, internalTypes.InternalEndpoint, "database")
+	path := fmt.Sprintf("https://%s/%s/%s", strings.Replace(addr, "%", "%25", 1), internalTypes.InternalEndpoint, "database")
```
roosterfish: The `url.Parse` (and the internal [`parse`](https://github.com/golang/go/blob/master/src/net/url/url.go)) do have some logic checks for this but as you have mentioned they expect the zone identifier to be already encoded (check for `%25`).
So I wonder if we can circumvent this by crafting the `request.URL` in a different way. Meaning that we provide the host and URL separately. Maybe the `url` package already has some tooling that could be useful for this?

rest/types/addrport.go
```diff
@@ -94,6 +94,19 @@ func (a AddrPort) String() string {
 	return a.AddrPort.String()
 }
 
+// WithoutZone returns the AddrPort with the zone identifier removed from the address. It will return the same
+// AddrPort as the receiver if the address is ipv4 or ipv6 without zone identifier.
+func (a AddrPort) WithoutZone() AddrPort {
```
roosterfish: It bugs me a little bit that there is nothing in the standard library for such checks so we have to remove the zone ID suffix ourselves for this check.

I think it's ok that we store the IPv6 with zone ID internally as this is what was ultimately configured so it's getting passed down in `hostAddress`. 
Have you checked if the `x/net/ipv6` package has something useful we can use to deal with the zone ID in the address?

roosterfish (commented):
Hi, thanks for your investigation on this. Please see my comments below.

roosterfish:
Can you please rebase from `v3`. We have added some more linter rules.

internal/db/dqlite.go
rapour: I couldn't find anything useful from the `url` package as it probably doesn't recognize the textual representation of ipv6 in RFC4007:
https://github.com/golang/go/blob/b07b20fbb591ac77158e1089299ce5acad71ffde/src/net/url/url_test.go#L704-L709
If you had any other directions in mind I would be happy to investigate. However, we can think of an alternative too. My impression is that `addr` string here is of type `AddrPort`, since `NewCluster` parses the input address into `AddrPort`. If it is indeed the case, we could first  parse the `addr` string into `AddrPort` and then handle these subtleties there in a canonical manner.

rest/types/addrport.go
rapour: The approach here is to strip the zone ID so it can be compared to the host and it uses the following feature from the standard library: 

https://github.com/golang/go/blob/b07b20fbb591ac77158e1089299ce5acad71ffde/src/net/netip/netip.go#L701-L704

Although the check is manual, the way to strip the zone ID is something done via the standard library. Interestingly, I found another way to strip the zone ID by using `WithZone("")`:
https://github.com/golang/go/blob/b07b20fbb591ac77158e1089299ce5acad71ffde/src/net/netip/netip.go#L489-L492
We can move to this one but at the end, both do the same and strip the zone ID so it becomes comparable to `Host`. The `netip` package has some comparison methods but I suspect any other solution would be as mediocre as this one since http request's `r.Host` is a string and will need parsing before any other checks. The `RemoteAddr` of the http request also has the full `host%zone` info which is comparable to the original `hostAddress` but there are no guarantees about the format of it and I'm not sure whether it's a good idea to replace it with the `r.Host`. 

As a side note, please also [see this](https://github.com/golang/go/issues/71362).

roosterfish: I like the idea of using the standard libraries `WithZone("")` using an empty string. Let's do this.

internal/db/dqlite.go
roosterfish: Is it your idea to modify the `AddrPort` type's `String()` to include the encoding? That sounds reasonable and also makes it usable for future use cases.

rest/types/addrport.go
rapour: Done, now we have the same interface for `types.AddrPort` as there is for `netip.Addr` when it comes to zone ID modifications.

internal/db/dqlite.go
rapour: I'm afraid changing the `String()` encoding may break other parts of the code that expect non-encoded `host%zone`. How do you feel about something like this: 
```
func dqliteNetworkDial(ctx context.Context, addr string, db *DqliteDB) (net.Conn, error) {
	addrPort, err := types.ParseAddrPort(addr)
	if err != nil {
		return nil, fmt.Errorf("Failed to parse the address: %w", err)
	}
	....
	request.URL = &url.URL{
		Scheme: "https",
		Host:   addrPort.String(),
		Path:   fmt.Sprintf("/%s/%s", internalTypes.InternalEndpoint, "database"),
	}
	...
}
```
This way we can actually circumvent the `url.Parse` and still keep the address validation.

roosterfish: That's exactly what I meant in my original comment:

> So I wonder if we can circumvent this by crafting the request.URL in a different way. Meaning that we provide the host and URL separately.

So if I understood your comment right, we don't have to apply any special encoding when doing `Host: addrPort.String()`? That seems much preferable.

rest/types/addrport.go
roosterfish: Can we use the `WithZone("")` from the std lib? I thought we use this one instead of having our own `WithoutZone()`.

rest/types/addrport_test.go
```diff
@@ -8,7 +8,6 @@ import (
 )
 
 func TestAddrPortWithoutZone(t *testing.T) {
-
```
roosterfish: We don't squash the commits when merging. Can you please squash this commit with the one where you have introduced the test?
This way you don't have a commit in the PR which adds lines which are being removed again by one of the follow up commits.

roosterfish:
Ah sorry. I haven't seen that you have already all the changes in the last commits.
Please see my other comment regarding squashing.

rest/types/addrport.go
roosterfish: Saw the latest commits too late. Fixed.

rest/types/addrport_test.go
roosterfish: This also applies to the other latest changes regarding `WithZone("")`.

internal/db/dqlite.go
rapour: Yes Exactly

rest/types/addrport_test.go
rapour: Done

roosterfish (dismissed):
Thanks for this!

roosterfish:
@rapour I cannot merge before your commits aren't signed as this is a rule of the repo.

roosterfish (approved):
LGTM!

PR #323, opened 2025-02-03
CI: Add linter rules from LXD

roosterfish:
This adds two of the linter rules from the LXD repo which should ease PR reviews from outside and keeps the two code bases consistent.

### DIFF
--- a/test/lint/no-oneline-assign-and-test.sh
+++ b/test/lint/no-oneline-assign-and-test.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+set -eu
+
+echo "Checking for oneline assign & test..."
+
+# Recursively grep go files for if statements that contain assignments.
+! git grep --untracked -P -n '^\s+if.*:=.*;.*{\s*$' -- '*.go'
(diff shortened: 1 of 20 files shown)


test/lint/no-oneline-assign-and-test.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -eu
+
+echo "Checking for oneline assign & test..."
+
+# Recursively grep go files for if statements that contain assignments.
+! git grep --untracked -P -n '^\s+if.*:=.*;.*{\s*$' -- '*.go' ':!:test/mini-oidc/storage/*.go'
```
MggMuggins: Probably don't need to include the mini-oidc exception here, but if the goal is to keep microcluster and LXD completely in sync I don't see anything wrong with this.

MggMuggins (dismissed):
Looks great, thanks!

test/lint/no-oneline-assign-and-test.sh
roosterfish: Ah thanks for catching this. I saw it whilst copy pasting it but forgot to delete it.

simondeziel (dismissed):
LGTM, I couldn't verify if those new test scripts are executable but they need to be :)

test/lint/no-oneline-assign-and-test.sh
roosterfish: Fixed. Also removed it from the other lint file.

simondeziel (approved):

PR #345, opened 2025-03-11
fix(cluster/stmt): proper project when vendored

troyanov:
Improve caller project detection when `go mod vendor` is used and when project directory structure contains `src`.

Previously this would be incorrectly assumed to be under GOPATH

Resolves https://github.com/canonical/microcluster/issues/350

### DIFF
--- a/cluster/stmt.go
+++ b/cluster/stmt.go
@@ -119,7 +119,16 @@ func GetCallerProject() string {
 		return base
 	}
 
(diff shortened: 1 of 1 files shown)


markylaing:
@troyanov thanks for your contribution but we will be merging #351. After which you will not need to call `GetCallerProject` at all :)

The function will remain in v2 and marked as deprecated. We'll remove it in v3.

PR #351, opened 2025-03-13
DB: Remove project

### LINKED ISSUE #350: Cannot initialize Microcluster if it is used as a vendored dependency
If I use `go mod vendor` then `file` equals to `/home/troyanov/Documents/work/canonical/go-test-bug-repro/src/vendor/github.com/canonical/microcluster/v2/cluster/cluster_members.mapper.go` 

https://github.com/canonical/microcluster/blob/90764cef5926a46e034e5871d66f64a9740f9c98/cluster/stmt.go#L123

Once `strings.Cut` logic is applied it becomes `vendor/github.com/canonical/microcluster/v2/cluster/cluster_members.mapper.go` and `GetCallerProject` returns `canonical` instead of `microcluster` (as it would return if `-mod=mod` is used)

This makes initialization failing with an error:
>Failed to check for duplicates: Failed to get "coreClusterMemberID" prepared statement: No prepared statement registered with code 3

roosterfish:
This is to explore the idea of not having the extra `project` field but instead tracking both internal and external SQL statements together.

Ultimately this allows getting rid of the "get caller" concept.
It was introduced as part of https://github.com/canonical/microcluster/pull/65 but doesn't seem to be required anymore.

Fixes https://github.com/canonical/microcluster/issues/350.

### DIFF
--- a/cluster/stmt.go
+++ b/cluster/stmt.go
@@ -29,8 +29,7 @@ func RegisterStmt(sql string) int {
 }
 
 // PrepareStmts prepares all registered statements and stores them in preparedStmts.
-// The project argument is kept for backwards compatibility but is deprecated.
-func PrepareStmts(db *sql.DB, project string, skipErrors bool) error {
+func PrepareStmts(db *sql.DB, skipErrors bool) error {
 	logger.Infof("Preparing statements")
 
 	for code, stmt := range stmts {
--- a/cluster/stmt.go
+++ b/cluster/stmt.go
@@ -72,6 +72,8 @@ func StmtString(code int) (string, error) {
 }
 
 // GetCallerProject will get the go project name of whichever function called `GetCallerProject`.
+//
+// Deprecated: The caller project is no longer required and causes issues when vendoring.
 func GetCallerProject() string {
 	sep := string(os.PathSeparator)
 
--- a/cluster/stmt.go
+++ b/cluster/stmt.go
@@ -12,8 +12,8 @@ import (
 	"github.com/canonical/lxd/shared/logger"
 )
 
-var stmtsByProject = map[string]map[int]string{} // Statement code to statement SQL text
-var preparedStmts = map[int]*sql.Stmt{}          // Statement code to SQL statement.
+var stmts = map[int]string{}            // Statement code to statement SQL text
+var preparedStmts = map[int]*sql.Stmt{} // Statement code to SQL statement.
 
 // RegisterStmt register a SQL statement.
 //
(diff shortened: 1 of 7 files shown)


cluster/stmt.go
```diff
@@ -22,47 +17,31 @@ var preparedStmts = map[int]*sql.Stmt{}          // Statement code to SQL statem
...
 
-	// Have a unique code for each statement, regardless of project,
-	// so we can access them without knowing the project.
+	// Have a unique code for each statement.
 	var code int
-	for _, stmts := range stmtsByProject {
-		code += len(stmts)
+	for _, stmt := range stmts {
+		code += len(stmt)
 	}
```
markylaing: This looks a bit off. `stmt` is a string, so range over the map of statement and add the length of each string to get the new code?

Shouldn't it just be `code := len(stmts) +1`?

roosterfish: Right, the loop can also be removed. Fixed.

markylaing:
This looks good :+1: 

The question now is whether this can be backported. I think the simple answer is no, unless we stub `GetCallerProject` and re-add the `project` argument to `RegisterStmt`, this will break micro* projects using v2. What do you think?

roosterfish:
> The question now is whether this can be backported.

`RegisterStmt` never had a project arg. We can add it back to `PrepareStmts` and keep `GetCallerProject` as it is but marking it as deprecated.
The remainder of the changes is in `internal/` anyway.

markylaing:
> > The question now is whether this can be backported.
> 
> `RegisterStmt` never had a project arg. We can add it back to `PrepareStmts` and keep `GetCallerProject` as it is but marking it as deprecated. The remainder of the changes is in `internal/` anyway.

Ah sorry I got my wires crossed there. Cool I'm happy with it. 

Can we add more of an explanation about the change in the commits and/or PR including the history of why `GetCallerProject` was ever used and also justification for why it is not needed? (E.g. we're in the clear as long as projects that use microcluster as well as import from other microcluster projects never import the packages where `RegisterStmt` is called)

roosterfish:
Successfully tested this now with custom builds of MicroCloud, MicroCeph and MicroOVN using this modified version. It's good to go.

The three most recent commits in this PR will only stay on `v3` and won't get backported into `v2`.

cluster/stmt.go
```diff
@@ -71,6 +71,7 @@ func StmtString(code int) (string, error) {
 	return "", fmt.Errorf("No prepared statement registered with code %d", code)
 }
 
+// Deprecated.
```
markylaing: I think the format of this needs to be something like this to mark things as deprecated in Go docs:

```golang
// GetCallerProject will get the go project name of whichever function called `GetCallerProject`.
//
// Deprecated: The caller project is no longer required and causes issues when vendoring.
// This is no longer required when calling `PrepareStmts`.
```

https://go.dev/blog/godoc

markylaing (changes requested):
One tiny nit and then good to go :)

cluster/stmt.go
roosterfish: Thanks, fixed. I haven't added the "This is no longer required when calling `PrepareStmts`" line as you still have to pass the project (empty string) to not introduce a breaking change.

markylaing (approved):

PR #352, opened 2025-03-18
doc: update example package developer docs

minaelee:
Full update of the instructions for using the example package.

### DIFF
--- a/example/README.md
+++ b/example/README.md
@@ -1,129 +1,295 @@
-# Example MicroCluster Daemon and CLI
+# Example implementation of Microcluster
 
-This is an example package containing a MicroCluster daemon command (`microd`) and a control command (`microctl`) to be
-used as a template for creating projects with MicroCluster.
+This example package can be used as a starting point for creating projects with Microcluster. This package contains:
+- A Microcluster daemon command (`microd`) and a control command (`microctl`)
+- Examples for the built-in Microcluster API
+- Examples of how Microcluster can be extended with additional listeners with user-defined endpoints and schema versions
 
-In addition to having examples for the built-in MicroCluster API, this example shows how MicroCluster can be extended
-with additional listeners with user defined endpoints and schema versions.
+## Contents
 
-The daemon can be started with `microd` and is controlled by `microctl`.
+- [How to build and run the package](#how-to-build-and-run-the-package)
+  - [Prerequisites](#prerequisites)
+  - [Build](#build)
+  - [Run](#run)
+  - [Control](#control)
+- [Tutorial](#tutorial)
+  - [Step 1: Build the package](#step-1-build-the-package)
+  - [Step 2: Start three Microcluster daemons](#step-2-start-three-microcluster-daemons)
+  - [Step 3: Start the Dqlite database](#step-3-start-the-dqlite-database)
+  - [Step 4: Join the other members to the cluster](#step-4-join-the-other-members-to-the-cluster)
+  - [Step 5: Interact with the cluster](#step-5-interact-with-the-cluster)
+    - [Step 5.1: Remove a cluster member](#step-51-remove-a-cluster-member)
+    - [Step 5.2: Perform SQL queries](#step-52-perform-sql-queries)
+    - [Step 5.3: Perform an extended API interaction](#step-53-perform-an-extended-api-interaction)
+  - [Step 6: Shut down the Microcluster](#step-6-shut-down-the-microcluster)
 
-## Building
-`make`
+## How to build and run the package
 
-## Running
-This starts up three daemons.
+### Prerequisites
+
+#### Required packages
+
+- `Go` must be installed on your system.
+  - If [snaps are available on your system](https://snapcraft.io/docs/installing-snapd), you can install Go by running:
+
+    ```
+    sudo snap install go --classic
+    ```
+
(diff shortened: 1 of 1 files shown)


example/README.md
```diff
@@ -1,129 +1,226 @@
...
-|      |                |       | JpmhZTvkvI8X4GLYU2KxuHR3+d3G1rePjErWrWle                         |        |
-|      |                |       | -----END CERTIFICATE-----                                        |        |
-|      |                |       |                                                                  |        |
-+------+----------------+-------+------------------------------------------------------------------+--------+
-```
-* Remove a cluster member
+microctl --state-dir /tmp/mc1 init mc1 127.0.0.1:9001 --bootstrap
+```
+
+You might see a warning that "The 'missing_extension' is not registered". Diregard this warning.
```
troyanov:
```suggestion
You might see a warning that "The 'missing_extension' is not registered". Disregard this warning.
```

minaelee: Fixed, thanks!

example/README.md
```diff
@@ -1,129 +1,226 @@
...
-## Building
-`make`
+## How to build and run the package
+
+### Prerequisites
+
+The following must be installed on your system:
+
+- `Go`
+- [Dqlite](https://github.com/canonical/dqlite?tab=readme-ov-file#install)
```
roosterfish: You can run `make deps` in order to install dqlite.

example/README.md
```diff
@@ -1,129 +1,226 @@
...
+```
+
+This demonstrates using the extended API to communicate between `mc1` and `mc2`.
+
+### Step 6: Shut down the Microcluster
+
+To shut down the Microcluster, stop the daemons using the PID variables defined in Step 2:
+
+```bash
+kill -9 $proc1 $proc2 $proc3
```
roosterfish: A simple `kill` (sending TERM) without `-9` should be sufficient.

roosterfish (changes requested):
Looking good, only two small comments.

example/README.md
minaelee: Here's what I found when testing in a LXC container:

Running `make deps` results in a series of errors until I install the following packages (along with `gcc` and Go):
`autoconf`, `libtool`, `pkg-config`, `libsqlite3-dev`, `libuv1-dev`. Afterward, `make deps` succeeds. However, after that, if I run `make` from the `example/` directory, I still get: 
```
# github.com/canonical/go-dqlite/v3/internal/bindings
/root/go/pkg/mod/github.com/canonical/go-dqlite/v3@v3.0.0/internal/bindings/server.go:12:10: fatal error: dqlite.h: No such file or directory
   12 | #include <dqlite.h>
      |          ^~~~~~~~~~
compilation terminated.
```
Adding `libdqlite-dev` via:
```
sudo add-apt-repository ppa:dqlite/dev
sudo apt update
sudo apt install libdqlite-dev
```
fixes that error - however, these are the installation instructions from Dqlite to begin with.

Thus, if someone is just trying to run `make` for the example to see what it's about, the shortest way to get there seems to be to install Dqlite according to that repository's instructions, rather than via `make deps` from the root of the repository. Thoughts?

roosterfish: Sorry I was missing the deps you have to install to be able to run `make deps`. In the pipeline they are getting installed here https://github.com/canonical/microcluster/blob/v3/.github/actions/install-builddeps/action.yml#L11.

The other important bit are those four environment variables which make sure the dqlite deps can be loaded during build time https://github.com/canonical/microcluster/blob/v3/.github/workflows/tests.yml#L10-L13.
It's important the the path points to the users home directory. `/home/runner` is only valid on the GitHub runners.

If you install the deps and set the env vars, you can run `make deps` followed by a `make -C example` to build the example package.

We should also mention you need to have Go installed, for example with `snap install go --classic`.

minaelee: For someone who is trying out the Microcluster example tutorial, the requirements I listed originally (Go, gcc, and dqlite installed via the dqlite instructions) is the simplest way to get the example package working. I've tested and the example package builds and runs with those three items only.

We could add a second set of instructions in the main README.md for getting `make` to work there, as currently there are no instructions for that, then link that as a secondary option in the example tutorial instructions (e.g., you can either install these minimal requirements just to build the example package if you're just interested in trying the tutorial, OR you can follow the instructions in the main README.md which installs the same requirements plus additional requirements.) Thoughts? We can discuss Monday as well.

roosterfish: We should not recommend installing `libdqlite-dev` as `make deps` will prepare the specific version of dqlite we use in `v3` (latest) and `v2` (the LTS). 
Just by running `make deps` you get the right one we also use during the build jobs. If this is detached from each other we always have to check if the dqlite from the deb is matching the one installed by `make deps`.

minaelee: Thanks, requirements updated & tested.

minaelee:
Tests / Code failing at `make check-static`. Unclear why; passes locally for both `make check-static` and `make -C example check-static`.

roosterfish:
> Tests / Code failing at `make check-static`. Unclear why; passes locally for both `make check-static` and `make -C example check-static`.

Should be fixed with https://github.com/canonical/microcluster/pull/364

roosterfish (approved):

roosterfish:
Can be rebased now.

minaelee:
> Can be rebased now.

Thanks, done!

PR #353, opened 2025-03-18
Endpoints: Fix implicit shutdown

roosterfish:
This addresses a regression introduced with https://github.com/canonical/microcluster/pull/295.

Before you were able to close endpoints (listeners) without closing the underlying server. As part of allowing graceful shutdown of servers (and listeners), we added code to implicitly shutdown the listeners underlying server for proper cleanup.

However it looks the code always assumed the underlying server never gets closed which allows instructing a remote Microcluster member to join the cluster without closing the connection that was used to instruct this member to join the cluster.

The PR introduces another flag that explicitly allows shutting down an endpoint's (listener's) underlying server. We can do this in most of the cases but should not do it in the following two:

* Close the server when a remote client instructed to join an existing cluster
* Close the server when restarting one of the extension servers. This allows keeping active connections alive.

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -137,7 +137,8 @@ func NewDaemon(project string) *Daemon {
 		}
 
 		if d.endpoints != nil {
-			err := d.endpoints.Down()
+			// Stop the listeners and shutdown the underlying servers.
+			err := d.endpoints.Down(true)
 			if err != nil {
 				return err
 			}
@@ -551,7 +552,9 @@ func (d *Daemon) StartAPI(ctx context.Context, bootstrap bool, initConfig map[st
 
 	d.extensionServersMu.RUnlock()
 
-	err = d.endpoints.Down(endpoints.EndpointNetwork)
+	// Close the listener but don't shutdown the underlying server.
+	// This allows staying connected with the API during the join procedure.
+	err = d.endpoints.Down(false, endpoints.EndpointNetwork)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


markylaing:
Just summarising our discussion:

#295 introduced connection draining so that when the server was shutdown, some connections would remain active until a timeout is reached, rather than abruptly closing.

As mentioned in the PR description, this also shutdown the server itself.

This is a problem for Microcluster because when a member is bootstrapped or joins another member, the listeners are closed in order to change the listen address from the pre-init listen address (for microcloud this is [::]:9443), to the address given on bootstrap. For joining members, listeners are closed in order to change the TLS certificate (though this isn't strictly necessary). 

This means the server is closed during the call to bootstrap or join, which we don't want. We just need to restart the listeners.

Just a thought @roosterfish - could this also be solved by introducing a default connection drain timeout?

In any case, we will need to differentiate between closing listeners and closing the server. So we'll go ahead and merge this PR.

markylaing (approved):

roosterfish:
Did some digging on this one. Closing the listeners during the restart when performing a bootstrap/join was added in https://github.com/canonical/microcluster/pull/43 for the exact reason we talked about yesterday: Being able to change the pre-init address.

Updating the certificate isn't affected by this as we use the same approach as in LXD to update the fancy TLS listener using the `d.ReloadCert()` func a few lines before closing the listener.

> could this also be solved by introducing a default connection drain timeout?

Are you thinking the restart of the listener could be done within the drain timeout to not interrupt the connection? The draining is a blocking operation though. See https://pkg.go.dev/net/http#Server.Shutdown.

In general the connection draining was introduced in a way to stay backwards compatible. Meaning there is no draining by default but you can enable it on any of the listeners if desired. This ensures the same behavior in the v2 LTS compared to before draining was available.

PR #356, opened 2025-03-20
doc: Update main README and doc/ contents

minaelee:
Completes all planned updates in https://warthogs.atlassian.net/browse/LXD-2645 except backporting to v2 branch, which is the final step after this and https://github.com/canonical/microcluster/pull/352 are merged.


doc/upgrades.md
```diff
@@ -10,15 +10,37 @@ A schema version is maintained per cluster member, representing the sum-total of
...
 
-https://github.com/canonical/microcluster/blob/v3/example/api/extensions.go
+```go
+// These are the extensions that are present when the daemon starts.
+var extensions = []string{
+	"custom_extension_a_0",
+	"custom_extension_a_1",
+}
+
+// Extensions returns the list of MicroOVN extensions.
```
roosterfish: The `MicroOVN` string seems odd here. Looks this is also wrong in the code.

roosterfish (changes requested):
Thanks, only a small comment.

doc/upgrades.md
minaelee: I'm assuming you mean it should say `Microcluster` in the comment, not `MicroOVN`? Fixed in both code and doc. Thanks!

roosterfish: Yes thank you.

roosterfish (approved):

PR #365, opened 2025-03-26
Bump delete timeout

claudiubelu:
30 seconds may be too short for the operation, as we're calling both ``DeleteClusterMember`` and ``RunPreRemoveHook``, which may take longer, especially if graceful shutdown is used.

If ``DeleteClusterMember`` and ``RunPreRemoveHook`` receive a context with a timeout, they should use that instead.

### DIFF
--- a/internal/rest/client/cluster.go
+++ b/internal/rest/client/cluster.go
@@ -2,6 +2,8 @@ package client
 
 import (
 	"context"
+	"math"
+	"strconv"
 	"time"
 
 	"github.com/canonical/lxd/shared/api"
@@ -57,6 +59,10 @@ func (c *Client) DeleteClusterMember(ctx context.Context, name string, force boo
 	if force {
 		endpoint = endpoint.WithQuery("force", "1")
 	}
+
+	deadline, _ := queryCtx.Deadline()
+	seconds := int(math.Floor(time.Until(deadline).Seconds()))
+	endpoint.WithQuery("timeout", strconv.Itoa(seconds))
 
 	return c.QueryStruct(queryCtx, "DELETE", internalTypes.PublicEndpoint, endpoint, nil, nil)
 }
--- a/internal/rest/client/cluster.go
+++ b/internal/rest/client/cluster.go
@@ -12,7 +12,7 @@ import (
 
 // AddClusterMember records a new cluster member in the trust store of each current cluster member.
 func AddClusterMember(ctx context.Context, c *Client, args types.ClusterMember) (*internalTypes.TokenResponse, error) {
-	queryCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	queryCtx, cancel := contextWithTimeout(ctx, 30*time.Second)
 	defer cancel()
 
 	tokenResponse := internalTypes.TokenResponse{}
@@ -26,7 +26,7 @@ func AddClusterMember(ctx context.Context, c *Client, args types.ClusterMember) 
 
 // ResetClusterMember clears the state directory of the cluster member, and re-execs its daemon.
 func ResetClusterMember(ctx context.Context, c *Client, name string, force bool) error {
-	queryCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	queryCtx, cancel := contextWithTimeout(ctx, 30*time.Second)
 	defer cancel()
 
 	endpoint := api.NewURL().Path("cluster", name)
@@ -39,7 +39,7 @@ func ResetClusterMember(ctx context.Context, c *Client, name string, force bool)
 
 // GetClusterMembers returns the database record of cluster members.
 func (c *Client) GetClusterMembers(ctx context.Context) ([]types.ClusterMember, error) {
-	queryCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	queryCtx, cancel := contextWithTimeout(ctx, 30*time.Second)
 	defer cancel()
 
 	clusterMembers := []types.ClusterMember{}
@@ -50,7 +50,7 @@ func (c *Client) GetClusterMembers(ctx context.Context) ([]types.ClusterMember, 
 
 // DeleteClusterMember deletes the cluster member with the given name.
 func (c *Client) DeleteClusterMember(ctx context.Context, name string, force bool) error {
-	queryCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	queryCtx, cancel := contextWithTimeout(ctx, 30*time.Second)
 	defer cancel()
 
 	endpoint := api.NewURL().Path("cluster", name)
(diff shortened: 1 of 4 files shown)


internal/rest/resources/cluster.go
```diff
@@ -416,7 +416,8 @@ func clusterMemberDelete(s state.State, r *http.Request) response.Response {
 		return response.SmartError(fmt.Errorf("No remote exists with the given name %q", name))
 	}
 
-	ctx, cancel := context.WithTimeout(r.Context(), time.Second*30)
+	// 30 seconds may be too short, as we're calling both DeleteClusterMember and RunPreRemoveHook.
+	ctx, cancel := context.WithTimeout(r.Context(), time.Second*60)
```
markylaing: Can we add a `timeout` query parameter instead, which is the timeout in seconds? If not provided, default to 30 seconds as this is the existing behaviour.

@roosterfish is this endpoint called by Microcluster itself or by client applications? If it's called by microcluster, perhaps we can allow a timeout to be passed in or configured somewhere.

roosterfish: It's part of the `/core/1.0/cluster/*` group of endpoints which is considered to be used from the outside, not only internally. The internal endpoints are under `/core/internal` therefore no version number.

> Can we add a timeout query parameter instead, which is the timeout in seconds? If not provided, default to 30 seconds as this is the existing behaviour.

I would very much vote for this too.

internal/rest/client/cluster.go
```diff
@@ -50,8 +50,14 @@ func (c *Client) GetClusterMembers(ctx context.Context) ([]types.ClusterMember,
 
 // DeleteClusterMember deletes the cluster member with the given name.
 func (c *Client) DeleteClusterMember(ctx context.Context, name string, force bool) error {
-	queryCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
-	defer cancel()
+	queryCtx := ctx
```
roosterfish: I am fine checking if the context already has a deadline. Please can we do the same on all of the cluster lient funcs in this file? 

Also I would keep the default of 30s.
You should pass the context derived by passing a timeout query param to this func (instead of `r.Context()`) when calling it from `deleteClusterMember()`.

roosterfish (changes requested):

internal/rest/client/cluster.go
claudiubelu: Added the same to the other functions. Kept the default 30s. Updated param in `deleteClusterMember()`.

internal/rest/resources/cluster.go
claudiubelu: Hmm, sure. Then probably the rest client should also send that information. I'm thinking of taking that timeout from the context deadline.

markylaing (approved):
I'm happy with this but will differ to @roosterfish as to whether any other handlers should also have a timeout parameter? 

Perhaps that can wait for a follow up PR though :)

bschimke95:
Honestly, I'm not a big fan of these somewhat arbitrary timeouts in a library like microcluster. We've had plenty of issues with canceled requests in the past because of this, often requiring [[awkward workarounds](https://github.com/canonical/k8s-snap/blob/7f64888fca1e6fca1d38492d91bd7f48d00a6207/src/k8s/pkg/k8sd/api/cluster_join.go#L56)](https://github.com/canonical/k8s-snap/blob/7f64888fca1e6fca1d38492d91bd7f48d00a6207/src/k8s/pkg/k8sd/api/cluster_join.go#L56) to mitigate the problem.  

A more robust approach, in my opinion, would be to rely on `r.Context` cancellation, at least for user-facing requests. This way, the client has full control over request duration, and we wouldn't need to keep adjusting timeouts manually.  

I realize this would require a fair amount of work and isn’t something we need to address *immediately*, but it might be worth considering for future improvements.

markylaing:
> A more robust approach, in my opinion, would be to rely on `r.Context` cancellation

By `r.Context` I assume you mean cancellation of the request context? If so, I agree that requests should be cancellable, but I don't agree that HTTP requests should be lasting longer than 30s. 

This endpoint makes the assumption that the pre-remove hook will not take very long. It shouldn't make that assumption because we don't know what users of the library will do.

My recommendation for long running operations like this would be to implement your own endpoint the performs all the pre-remove logic and handle the complexities of that long-running operation as you see fit. Then call it before running the microcluster builtins. Then delete the cluster member via microcluster if it succeeds to remove it from the core tables. (E.g. don't use the hook)

internal/rest/client/cluster.go
```diff
@@ -58,6 +60,10 @@ func (c *Client) DeleteClusterMember(ctx context.Context, name string, force boo
 		endpoint = endpoint.WithQuery("force", "1")
 	}
 
+	deadline, _ := queryCtx.Deadline()
+	seconds := int(math.Floor(time.Until(deadline).Seconds()))
```
roosterfish, line 64: You can do `time.Until(future).Truncate(time.Second).Seconds()` which would not require the `math` package.

roosterfish: Now after seeing the proposed change I don't like that we both set a context for the request but also tell the server to wait for the exact same period of time. It's almost like our local context races with the timeout send to the server. And having both feels redundant.

roosterfish (changes requested):
I now went through the proposed changes a second time and more carefully checked the underlying API handler `clusterMemberDelete`. I think we have to be more precise with the timeout as overall the 30s are more than enough for the regular Microcluster related bits (removing the member from the cluster). And for this purpose the fixed context timeout of 30s is feasible to me.

By adding a general `timeout` query param to accommodate long running hooks, we hide the fact that this is a hook problem and just allow increasing the timeout for the overall operation. In addition  timeout gets set to the same as the callers request timeout which feels odd as we set the same information twice.
Right now we already make use of the `r.Context()` when calling the `DeleteClusterMember` client func which then derives from the request's context and adds the 30s deadline.

To solve the problem at its root, let's add a `hook_timeout` query param which is solely used to derive a context with the specified deadline. This context gets then passed into `RunPreRemoveHook` and `PostRemove` hook but it wouldn't affect all the other API calls Microcluster is performing under the hood.
For this `DeleteClusterMember` would accept another arg e.g. `hookTimeout`.
But this would be a breaking API change as we change the functions signature and `DeleteClusterMember` is potentially in use currently. For example the `example/` package is using it too https://github.com/canonical/microcluster/blob/v3/example/cmd/microctl/cluster_members.go#L192.
So we could only have this in `v3`.

What do you think @markylaing? I like your idea of performing long running cleanup tasks in its own API handler without conflating Microcluster hooks.

markylaing:
After an internal discussion, @roosterfish and I have decided that we don't want to bump the deletion handler timeout to longer than 30 seconds.

This is because: 
- The hooks run synchronously, blocking execution of HTTP handlers.
- Plain HTTP handlers (e.g. no websocket) shouldn't be running for longer than that. Our justification for this is that there is no way to provide feedback to the caller as to why the operation is taking a long time, or what is happening server-side.
- After bumping a timeout to longer than 30 seconds, we have to ask what the upper limit is.

In general, across Microcluster we have set a default of 30s max for HTTP requests for this reason. It should be more than long enough to perform any Microcluster internal operations (including any Dqlite operations). Hooks are ultimately intended for e.g. receiving status updates when certain operations happen. We want them to be a convenience, but we don't expect them to fit all use cases. 

Our recommendation is to implement another `pre-remove` API handler, and call it before the cluster member is deleted. Your handler could use websockets to return operation metadata back to the user (see https://github.com/canonical/microcluster/blob/v3/client/client.go#L45).

PR #412, opened 2025-06-03
fix: Tune CheckReady timeout

HomayoonAlimohammadi:
### Overview

This PR tunes the context passed to `CheckReady`.  Consider this scenario: In a 3 node cluster, if 1 node is temporarily unreachable, various things might fail (AFAIK things that rely on a `GetClusterMembers` call), e.g. a heartbeat.
E.g. [here](https://github.com/canonical/microcluster/blob/v2/internal/rest/resources/heartbeat.go#L187) we're [calling `GetClusterMembers`](https://github.com/canonical/microcluster/blob/v2/internal/state/state.go#L174) with a [30 seconds timeout](https://github.com/canonical/microcluster/blob/v2/internal/rest/client/cluster.go#L42-L46). In that call, we're looping over cluster members and sending each one a `CheckReady` with the [request context](https://github.com/canonical/microcluster/blob/v2/internal/rest/resources/cluster.go#L304). This call will have a [30 seconds timeout](https://github.com/canonical/microcluster/blob/v2/internal/rest/client/ready.go#L14-L18) on its own. In an ideal world we should be smart with the timeout of this context (something like `deadline / cluster_members`), but we don't know the deadline with which our clients are calling us (`r.Context().Deadline()` is server side. Timeout is not shipped from the client to the server). But we can at least put a custom timeout on the context used for this call, which this PR aims to do.  

We can also reduce the timeout which is set [here](https://github.com/canonical/microcluster/blob/v2/internal/rest/client/ready.go#L14-L18) instead. I'm open to any suggestions or alternatives.

### DIFF
--- a/internal/rest/client/cluster.go
+++ b/internal/rest/client/cluster.go
@@ -44,7 +44,8 @@ func (c *Client) GetClusterMembers(ctx context.Context) ([]types.ClusterMember, 
 	defer cancel()
 
 	endpoint := api.NewURL().Path("cluster")
-	if deadline, ok := queryCtx.Deadline(); ok {
+	deadline, ok := queryCtx.Deadline()
+	if ok {
 		timeout := int(time.Until(deadline).Seconds())
 		endpoint = endpoint.WithQuery("timeout", strconv.Itoa(timeout))
 	}
--- a/internal/rest/client/cluster.go
+++ b/internal/rest/client/cluster.go
@@ -2,6 +2,7 @@ package client
 
 import (
 	"context"
+	"strconv"
 	"time"
 
 	"github.com/canonical/lxd/shared/api"
@@ -42,8 +43,14 @@ func (c *Client) GetClusterMembers(ctx context.Context) ([]types.ClusterMember, 
 	queryCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
 	defer cancel()
 
+	endpoint := api.NewURL().Path("cluster")
+	if deadline, ok := queryCtx.Deadline(); ok {
+		timeout := int(time.Until(deadline).Seconds())
+		endpoint = endpoint.WithQuery("timeout", strconv.Itoa(timeout))
+	}
+
 	clusterMembers := []types.ClusterMember{}
-	err := c.QueryStruct(queryCtx, "GET", internalTypes.PublicEndpoint, api.NewURL().Path("cluster"), nil, &clusterMembers)
+	err := c.QueryStruct(queryCtx, "GET", internalTypes.PublicEndpoint, endpoint, nil, &clusterMembers)
 
 	return clusterMembers, err
 }
--- a/internal/rest/resources/cluster.go
+++ b/internal/rest/resources/cluster.go
@@ -307,33 +307,54 @@ func clusterGet(s state.State, r *http.Request) response.Response {
 			return response.SmartError(err)
 		}
 
+		wg := sync.WaitGroup{}
+		wg.Add(len(apiClusterMembers))
+
 		for i, clusterMember := range apiClusterMembers {
-			addr := api.NewURL().Scheme("https").Host(clusterMember.Address.String())
-			d, err := internalClient.New(*addr, s.ServerCert(), clusterCert, false)
-			if err != nil {
-				return response.SmartError(fmt.Errorf("Failed to create HTTPS client for cluster member with address %q: %w", addr.String(), err))
-			}
+			go func(i int, clusterMember types.ClusterMember) {
+				defer wg.Done()
 
-			var (
-				checkCtx context.Context
-				cancel   context.CancelFunc
-			)
-			if deadline, ok := ctx.Deadline(); ok {
-				until := time.Until(deadline)
-				timeout := (until / time.Duration(len(apiClusterMembers))).Truncate(time.Second)
-				timeout = max(time.Second, timeout)
-				checkCtx, cancel = context.WithTimeout(ctx, timeout)
-			} else {
-				checkCtx, cancel = context.WithCancel(ctx)
-			}
+				addr := api.NewURL().Scheme("https").Host(clusterMember.Address.String())
+				d, err := internalClient.New(*addr, s.ServerCert(), clusterCert, false)
+				if err != nil {
+					logger.Errorf("Failed to create HTTPS client for cluster member with address %q: %v", addr.String(), err)
+					return
+				}
 
-			err = d.CheckReady(checkCtx)
-			if err == nil {
-				apiClusterMembers[i].Status = types.MemberOnline
-			} else {
-				logger.Warnf("Failed to get status of cluster member with address %q: %v", addr.String(), err)
-			}
-			cancel()
+				var (
+					checkCtx context.Context
+					cancel   context.CancelFunc
+				)
+				if deadline, ok := ctx.Deadline(); ok {
+					until := time.Until(deadline)
+					timeout := (until / time.Duration(len(apiClusterMembers))).Truncate(time.Second)
+					timeout = max(time.Second, timeout)
+					checkCtx, cancel = context.WithTimeout(ctx, timeout)
+				} else {
+					checkCtx, cancel = context.WithCancel(ctx)
+				}
+
+				err = d.CheckReady(checkCtx)
+				if err == nil {
+					apiClusterMembers[i].Status = types.MemberOnline
+				} else {
+					logger.Warnf("Failed to get status of cluster member with address %q: %v", addr.String(), err)
+				}
+
+				cancel()
+			}(i, clusterMember)
+		}
+
+		waitCh := make(chan struct{})
+		go func() {
+			wg.Wait()
+			close(waitCh)
+		}()
+
+		select {
+		case <-ctx.Done():
+			return response.SmartError(fmt.Errorf("failed while waiting for cluster members to respond: %w", ctx.Err()))
+		case <-waitCh:
 		}
 	}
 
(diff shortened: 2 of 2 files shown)


roosterfish (changes requested):
Thanks for the PR.

I wonder do you encounter issues waiting for 30s until the request to `GET /core/1.0/cluster` returns or do you just want to speed up the response? 
We have to be careful to not break existing setups so lowering the number might have side effects. 

Also please open up this PR against `v3` and not `v2` as we land new features/fixes to `v3` first and then decide whether or not we can backport them into `v2`.

HomayoonAlimohammadi:
Thanks for the review @roosterfish,

Yes we've been waiting for 30 seconds.
We have 2 issues basically. One of them is that heartbeats fail in the scenario that I described (1 node is cut off). I believe a solution to that would be to increase the heartbeat interval, to give enough time to the other calls (since the heartbeat timeout is based on the heartbeat interval). However, I don't think that increasing the heartbeat interval is the correct way forward here.
The other problem is a call we make to the `GetClusterMembers` endpoint (which fails after the timeout). You can argue that when a node is not reachable `GetClusterMembers` _should_ fail, but according to the [code](https://github.com/HomayoonAlimohammadi/microcluster/blob/v2/internal/rest/resources/cluster.go#L304-L309), if the nodes are reachable their status will change to `types.MemberOnline`, otherwise, we don't fail anything, but just log a warning. 

Also about raising the PR against `v3`, we're using `v2` in Canonical Kubernetes and I think we need to stick with `v2` for now. This makes a backport almost necessary for us. I'll change the base for this one to `v3` and then we can discuss how we can incorporate the fix in `v2`.

markylaing:
> ### Overview
> This PR tunes the context passed to `CheckReady`. Consider this scenario: In a 3 node cluster, if 1 node is temporarily unreachable, various things might fail (AFAIK things that rely on a `GetClusterMembers` call), e.g. a heartbeat. E.g. [here](https://github.com/canonical/microcluster/blob/v2/internal/rest/resources/heartbeat.go#L187) we're [calling `GetClusterMembers`](https://github.com/canonical/microcluster/blob/v2/internal/state/state.go#L174) with a [30 seconds timeout](https://github.com/canonical/microcluster/blob/v2/internal/rest/client/cluster.go#L42-L46). In that call, we're looping over cluster members and sending each one a `CheckReady` with the [request context](https://github.com/canonical/microcluster/blob/v2/internal/rest/resources/cluster.go#L304). This call will have a [30 seconds timeout](https://github.com/canonical/microcluster/blob/v2/internal/rest/client/ready.go#L14-L18) on its own. In an ideal world we should be smart with the timeout of this context (something like `deadline / cluster_members`), but we don't know the deadline with which our clients are calling us (`r.Context().Deadline()` is server side. Timeout is not shipped from the client to the server). But we can at least put a custom timeout on the context used for this call, which this PR aims to do.
> 
> We can also reduce the timeout which is set [here](https://github.com/canonical/microcluster/blob/v2/internal/rest/client/ready.go#L14-L18) instead. I'm open to any suggestions or alternatives.

Hi @HomayoonAlimohammadi, just to make sure I'm understanding things correctly:

- It's possible for a call to `GET /core/1.0/cluster` to take up to 30 seconds if one or more cluster members are offline?
- You can't set a deadline for the server side requests when client side.

I don't think it would work to reduce the request context in your second suggestion. If I understand correctly, that would cancel the request client side, so it wouldn't return anything.

It sounds like rather than setting an arbitrary timeout (5 seconds, similar to 30 seconds, is arbitrary), you would like control over how long to wait for other servers in the cluster to respond. If that's the case I suggest adding a `timeout` parameter. This would be a bit like the initial suggestion in https://github.com/canonical/microcluster/pull/365.

HomayoonAlimohammadi:
Hi @markylaing,
Yes. Basically, if even one cluster member is unreachable, `GetClusterMembers` call will fail. Because the timeout for that call is [30 seconds](https://github.com/HomayoonAlimohammadi/microcluster/blob/v3/internal/rest/client/cluster.go#L41-L49), and so is the timeout for [`CheckReady` call](https://github.com/HomayoonAlimohammadi/microcluster/blob/v3/internal/rest/client/ready.go#L14-L19), which is made to each member [one by one](https://github.com/HomayoonAlimohammadi/microcluster/blob/v3/internal/rest/resources/cluster.go#L297-L310).

5 seconds _is_ arbitrary (just like 30 seconds), but IMO it makes more sense, since the parent timeout is 30 seconds (`GetClusterConfig`).

~~I'm not sure how to proceed with adding a parameter. We're hoping to tune the `CheckReady` timeout, so the `clusterGet` handler should somehow know about this timeout. But how are we supposed to do that?~~
Looking at the provided PR, I'll proceed with adding a `timeout` param.

roosterfish:
> ~I'm not sure how to proceed with adding a parameter. We're hoping to tune the `CheckReady` timeout, so the `clusterGet` handler should somehow know about this timeout. But how are we supposed to do that?~ Looking at the provided PR, I'll proceed with adding a `timeout` param.

I wonder if we should rather check the cluster members state in dqlite/raft if they are online and don't even try to reach out to them inside the `/core/1.0/cluster` endpoint if they are offline?
Because then we don't require adding another flag but can instead resolve it internally.

Also it seems there already is logic to report whether a cluster member is offline or not. 
In case the `CheckReady` call to the offline member returns right away (e.g. due to "connection refused"), the API response lists this member with status `UNREACHABLE`. 

So I think just skipping this member in case it's reported as offline from the dqlite/raft perspective is a good option here.

HomayoonAlimohammadi:
@roosterfish It looks like the status of the `types.ClusterMember` is [hard-coded](https://github.com/HomayoonAlimohammadi/microcluster/blob/v3/cluster/cluster_members.go#L81) as `UNREACHABLE`. I'm also not sure how we can check the dqlite/raft member status since it's not available in the [retrieved information](https://github.com/HomayoonAlimohammadi/microcluster/blob/v3/cluster/cluster_members.go#L40-L50). Should we start storing that information?

HomayoonAlimohammadi:
We can also make `CheckReady` calls concurrently to make things faster. Lemme try adding that as well.

roosterfish:
> @roosterfish It looks like the status of the `types.ClusterMember` is [hard-coded](https://github.com/HomayoonAlimohammadi/microcluster/blob/v3/cluster/cluster_members.go#L81) as `UNREACHABLE`. I'm also not sure how we can check the dqlite/raft member status since it's not available in the [retrieved information](https://github.com/HomayoonAlimohammadi/microcluster/blob/v3/cluster/cluster_members.go#L40-L50). Should we start storing that information?

AFAIK `UNREACHABLE` is the default but it gets changed to the actual status in case the lookup was working.

HomayoonAlimohammadi:
@roosterfish That's right, but _if_ the lookup was working. Which means we _need_ the lookup. From what I understood from your previous message, we're looking for a way to prevent doing lookups for nodes that are already reported as "offline", right?
> I wonder if we should rather check the cluster members state in dqlite/raft if they are online and don't even try to reach out to them inside the `/core/1.0/cluster` endpoint if they are offline?

HomayoonAlimohammadi:
We change the member status [here](https://github.com/HomayoonAlimohammadi/microcluster/blob/v3/internal/rest/resources/cluster.go#L306) _if_ it was found to be reachable.

HomayoonAlimohammadi:
May I have some guidance on how to prevent the lookup? To me, the lookup seems an essential part that can change the status from `UNREACHABLE` (default) to `ONLINE`. Without it, all nodes will be considered unreachable.

roosterfish:
> To me, the lookup seems an essential part that can change the status from UNREACHABLE (default) to ONLINE. Without it, all nodes will be considered unreachable.

I was now digging into this a bit more. Unfortunately I wasn't able to find an option to get the member status from the underlying dqlite/raft layer.
However I was checking a bit what we do for LXD and this seems promising.

In case any of the nodes fails to complete the heartbeat, this information is getting stored inside the `heartbeat` in the `core_cluster_members` table.
We can leverage this information to decide whether or not a node is actually reachable and decide whether or not we want to call the `CheckReady` func on it.

This is how it's done in LXD:
* Check the last heartbeat time using `IsOffline` https://github.com/canonical/lxd/blob/main/lxd/db/node.go#L154. This is using a configurable threshold after which the member is considered offline
* The actual offline check performs those steps https://github.com/canonical/lxd/blob/main/lxd/db/node.go#L1153-L1157

The [default threshold](https://documentation.ubuntu.com/lxd/latest/server/#server-cluster:cluster.offline_threshold) in LXD is 20s.
I think we can go with the same number in case of Microcluster.

Now to get this into Microcluster, I would propose the following:
* Extend the daemon [`Args`](https://github.com/canonical/microcluster/blob/v3/internal/daemon/daemon.go#L48) and add another `HeartbeatOfflineThreshold time.Duration` (or similar) which defaults to 20s.
* When [looping](https://github.com/canonical/microcluster/blob/v3/internal/rest/resources/cluster.go#L298-L311) through the list of cluster members, check if the node `IsOffline` using the last `clusterMember.LastHeartbeat` field and the threshold.

roosterfish:
As a follow up, this would also help to fix a potential bug in `microcloud status` in case one of the members got lost. The status command internally uses the same endpoint and ends up in the deadline:

```bash
root@micro01:~# microcloud status
⨯ Error: Get "https://10.87.144.48:9443/core/1.0/cluster": context deadline exceeded
```

roosterfish:
@HomayoonAlimohammadi any updates on this? Or should we close it for now?

roosterfish:
Closing it now due to inactivity.

PR #413, opened 2025-06-03
Update deps and untangle relation to LXD

### LINKED ISSUE #407: Microcluster doesn't work with latest `stable-5.21` LXD dep
Microcluster (v2/v3) uses the `schema.ErrGracefulAbort` from LXD. As the error was deleted from the LXD code base in https://github.com/canonical/lxd/commit/1470442601bc18e1dadbcc37ccd02a8961a7d677, Microcluster cannot anymore be built using the latest version of LXD `stable-5.21` branch.
This also affects all importers of Microclusters (e.g. MicroCloud) and blocks dependency updates to the latest versions.

As a fix the `schema` dependencies of Microcluster should be resolved by including the types in the Microcluster codebase itself.

roosterfish:
This PR removes some dependencies to the LXD code base by moving a custom error type which doesn't anymore exist in LXD and switch to using `slices.Contains` instead of LXD's `shared.ValueInSlice` which also got removed.

Furthermore all deps are updated to the latest versions.

Fixes https://github.com/canonical/microcluster/issues/407.


roosterfish:
Ultimately the `ErrGracefulAbort` might end up at the callers level (by invoking `app.Start`) but none of the Micro* (and k8s) are currently checking for this error type in their code bases so we should be good to proceed with this.

internal/db/errors/errors.go
```diff
@@ -0,0 +1,6 @@
+package errors
+
+import "fmt"
+
+// ErrGracefulAbort is a special error that indicates a non successful schema check.
+var ErrGracefulAbort = fmt.Errorf("schema check gracefully aborted")
```
markylaing: Do you think we need a new package for this?

internal/db/update/schema.go
```diff
@@ -151,11 +152,11 @@ func (s *SchemaUpdate) Ensure(db *sql.DB) (int, error) {
 
 		if s.check != nil {
 			err := s.check(ctx, current, tx)
-			if err != nil && err != schema.ErrGracefulAbort {
+			if err != nil && err != dbErrors.ErrGracefulAbort {
```
markylaing: Can we change this to `errors.Is(...)`?

markylaing (commented):
Just a couple of nits/questions :)

internal/db/errors/errors.go
roosterfish: It is to prevent an import cycle. Both the `db` and `db/update` packages are using it.

internal/db/update/schema.go
roosterfish: Good catch, now the following lines are also cleanup up and the redundant condition is removed.

roosterfish: Had to revert the removal of the condition as it actually served a purpose.

internal/db/errors/errors.go
markylaing: That's annoying. Since this is all internal maybe we can reorganise to flatten `db/update` into `db` instead. Can be a separate PR if necessary.

roosterfish: Mh, I have now checked the code a bit closer and ultimately the error is triggered in case a schema update/upgrade cannot be performed. So placing it inside `db/update` actually isn't a bad idea to keep the update related bits together.

internal/db/update/schema.go
roosterfish: Fixed.

internal/db/errors/errors.go
roosterfish: Fixed.

markylaing (approved):
Thanks!

PR #470, opened 2025-08-28
Dqlite: Fix TLS handshake errors

### LINKED ISSUE #398: TLS handshake error regression
Hi, it looks that https://github.com/canonical/go-dqlite/pull/381 caused a regression starting from `go-dqlite` `v3.0.1`. `v3.0.0` seems to be not affected.
As it got backported into `v2` in https://github.com/canonical/go-dqlite/commit/1d1fe51f66858f85fbb4c78096023aae06aae59d, it also affects the `v2.0.1` LTS (and all the Micro* LTS releases as they use Microcluster v2 which uses `go-dqlite@v2.0.1`).

I started to notice that the following errors appear in MicroCloud (potentially also in other Micro* products using Microcluster) from time to time, sometimes more frequently:

`2025/08/26 10:46:47 http: TLS handshake error from 127.0.0.1:56144: EOF`

Sometimes they already appear when forming the cluster, sometimes they just appear throughout the lifetime of the cluster.
As far as I could see through the testing, its only printed on the current leader. 

To confirm it's caused by this PR, I have checked out `v3.0.2` and reverted commit https://github.com/canonical/go-dqlite/commit/bddbf20759568cbe8930038c7d7c3368c55af180. Running Microcluster using this modified version of `go-dqlite@v3.0.2` doesn't anymore produce the error.

This comment in the original file caught my attention as it potentially describes exactly this type of error message:

```
// The new context will be cancelled when we successfully connect
// to the leader. The original context will be used only for net.Dial.
// Motivation: threading the cancellation through to net.Dial results
// in lots of warnings being logged on remote nodes when our probing
// goroutines disconnect during a TLS handshake.
```

Please let me know if I can provide more details.

roosterfish:
See https://github.com/canonical/go-dqlite/issues/398 for reference.
Closes https://github.com/canonical/go-dqlite/issues/398.

Microcluster's custom dial function passed into `go-dqlite` might cause TLS handshake errors on the target members in case the context passed into the dial function is cancelled early on. That might be the case when a leader was already found, so all the other parallel connection attempts to the other members are cancelled.
Whilst this is not a functional issue we should have extra handling to prevent seeing those error messages in the Microcluster daemon logs.

Furthermore the passed context's deadline is now used properly. Setting it whilst using the same context the deadline is coming from doesn't bring much value.
In addition a potential leak of the request's response body is fixed.

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"log"
 	"net/http"
 	"os"
 	"path/filepath"
@@ -498,6 +499,7 @@ func (d *Daemon) initServer(resources ...rest.Resources) *http.Server {
 	return &http.Server{
 		Handler:     mux,
 		ConnContext: request.SaveConnectionInContext,
+		ErrorLog:    log.New(newLogFilter(state.Remotes().Addresses), "", 0),
 	}
 }
 
--- a/internal/daemon/logfilter.go
+++ b/internal/daemon/logfilter.go
@@ -1,0 +1,65 @@
+package daemon
+
+import (
+	"regexp"
+
+	"github.com/canonical/lxd/shared/logger"
+
+	"github.com/canonical/microcluster/v3/rest/types"
+)
+
+// logFilter represents a filter for log messages caused by known addresses.
+type logFilter struct {
+	knownAddresses func() map[string]types.AddrPort
+}
+
+// newLogFilter returns a new instances of the log filter.
+// It expects a slice of known addresses for which log messages will get filtered.
+func newLogFilter(knownAddresses func() map[string]types.AddrPort) *logFilter {
+	return &logFilter{
+		knownAddresses: knownAddresses,
+	}
+}
+
+// unwantedLogRegex represents log texts whose presence will cause log messages to be filtered out.
+var unwantedLogRegex = regexp.MustCompile(`^http: TLS handshake error from (.*):[0-9]+: EOF$`)
+
+// Write filters the given log message p and checks whether or not it contains any
+// of the unwanted log messages.
+// If there is a match, a debug message is logged with the unwanted log message
+// In all other cases the message is logged using the actual logger.
+func (l logFilter) Write(p []byte) (int, error) {
+	strippedLog := l.stripLog(p)
+	if strippedLog == "" {
+		return 0, nil
+	}
+
+	logger.Info(strippedLog)
+	return len(p), nil
+}
+
+// stripLog strips the log message to determine whether or not its unwanted.
+func (l logFilter) stripLog(p []byte) string {
+	// Get the source IP address.
+	match := unwantedLogRegex.FindSubmatch(p)
+	var sourceIP string
+	if match != nil {
+		if match[1] != nil {
+			sourceIP = string(match[1])
+		}
+	}
+
+	logStr := string(p)
+
+	// Discard the log if the source is in our list of known addresses.
+	if sourceIP != "" {
+		for _, knownAddress := range l.knownAddresses() {
+			if knownAddress.Addr().String() == sourceIP {
+				logger.Debug("Filtered out unwanted log text", logger.Ctx{"text": logStr})
+				return ""
+			}
+		}
+	}
+
+	return logStr
+}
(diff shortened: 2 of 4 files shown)


tomponline:
If deadline is zero, does WithDeadline never cancel or immediately cancel?

roosterfish:
> If deadline is zero, does WithDeadline never cancel or immediately cancel?

I got rid of this part in the changes. But I guess it will never cancel as deadline is zero. So it's the same as setting no deadline.

roosterfish:
@markylaing let's talk this through during our 1:1 next week.

internal/daemon/logfilter.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"log"
+	"strings"
+
+	"github.com/canonical/lxd/shared/logger"
+)
+
+// logBlacklist represents log texts whose presence will cause log messages to be filtered out.
+var logBlacklist = []string{
+	"http: TLS handshake error from",
```
markylaing: I think we should do a quick review of the errors that are logged by the go stdlib http server to ensure we aren't going to prevent logging other unexpected errors with this prefix. We might need to be a little more restrictive in the blacklist as LXD does with the regexp filtering

roosterfish: As discussed earlier I'll validate how feasible it is to use the list of known cluster members so we only filter out error messages from our own members, but not from external clients talking to the API.

roosterfish: Pushed the necessary changes.

I found that the prefix/newline stripping we are using in LXD's equivalent to this isn't required so I didn't include it. Also I have simplified the regex for the exact error message we are facing in this case.
Let me know what you think.

internal/daemon/daemon.go
```diff
@@ -495,9 +496,19 @@ func (d *Daemon) initServer(resources ...rest.Resources) *http.Server {
...
+	// Collect a list of known addresses from all the remotes.
+	// This is useful to be able to filter out log messages from specific members based on their source address.
+	for _, remoteAddress := range state.Remotes().Addresses() {
+		knownAddresses = append(knownAddresses, remoteAddress)
+	}
+
 	return &http.Server{
 		Handler:     mux,
 		ConnContext: request.SaveConnectionInContext,
+		ErrorLog:    log.New(newLogFilter(knownAddresses), "", 0),
```
markylaing: Just for my understanding: When a new member joins or when a member is removed, is the http server restarted? E.g. will this list of known addresses always be correct?

roosterfish: Good catch, no it won't get restarted. Looks like the log filter needs to query for the addresses on demand.

roosterfish: Fixed. Also got rid of an obsolete extra loop to craft a slice of addresses which isn't anymore required. Before I was using strings to represent the addresses but have decided to use the `AddrPort` type directly for convenience.

internal/daemon/daemon.go
```diff
@@ -498,6 +499,7 @@ func (d *Daemon) initServer(resources ...rest.Resources) *http.Server {
 	return &http.Server{
 		Handler:     mux,
 		ConnContext: request.SaveConnectionInContext,
+		ErrorLog:    log.New(newLogFilter(state.Remotes().Addresses), "", 0),
```
markylaing, line 502: Last question - is calling `state.Remotes().Addresses()` expensive? I worry that if something causes the http server to spam logs we'll have trouble accessing the truststore.

IIRC this acquires a read lock on the truststore, which should be fine especially since a write lock is only acquired for infrequent trust store changes.

roosterfish: You are right with the locks. I was also checking this code path before to be sure it's not too expensive. Inside `Addresses` there is a loop which creates the map and then releases the lock again.

roosterfish:
@markylaing when you are happy with the PR please leave it open and don't merge it yet. 
Whilst this fixes the error message we are seeing, I found we could maybe use a different method on `go-dqlite` to get to the leader which maybe won't cause those error messages on the target side. Testing this right now.

markylaing (approved):
LGTM thanks :partying_face:

roosterfish:
What I have mentioned in [this](https://github.com/canonical/microcluster/pull/470#issuecomment-3244611789) comment isn't a full alternative unfortunately. However we could think about using `go-dqlite`'s `FindLeader` instead of `Leader` to reuse an existing connection to the leader. This also reduces the likelihood of the error message we are trying to fix here, but doesn't remove it completely.

I'll add another PR for this and will merge this one now.

PR #472, opened 2025-09-03
Dqlite: TLS error followup for IPv6

roosterfish:
Follow up on https://github.com/canonical/microcluster/pull/470 to also account for IPv6 addresses. 

I initially thought this is working for IPv4 and IPv6 but more tests have shown that in case of IPv6 we need to handle a newline and the additional brackets.

### DIFF
--- a/internal/daemon/logfilter.go
+++ b/internal/daemon/logfilter.go
@@ -3,6 +3,7 @@ package daemon
 import (
 	"bytes"
 	"regexp"
+	"strings"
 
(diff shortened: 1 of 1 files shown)


roosterfish:
Pipeline failing due to change in `dqlite` which isn't yet reflected by a new release of `go-dqlite`.

roosterfish:
Can be rebased once https://github.com/canonical/microcluster/pull/473 is merged to unblock the dqlite issues.

markylaing (approved):
LGTM thanks

PR #473, opened 2025-09-04
Go: Update dependencies

roosterfish:
This updates all the Go deps including the latest LXD. 

Some changes are required to accommodate recent changes in LXD. Also some types are now moved into Microcluster as they got removed in LXD.


markylaing (approved):
Good stuff, thanks

internal/rest/client/client.go
```diff
@@ -171,20 +178,6 @@ func (c *Client) SetClusterNotification() {
 func forwardingProxy(r *http.Request) (*url.URL, error) {
 	r.Header.Set("User-Agent", clusterRequest.UserAgentNotifier)
 
-	ctx := r.Context()
-
-	val, ok := ctx.Value(request.CtxUsername).(string)
-	if ok {
-		r.Header.Add(request.HeaderForwardedUsername, val)
```
roosterfish, line 178: @tomponline this is the bit I mentioned yesterday in our 1:1. All good. I remembered wrong, I actually dropped them in Microcluster as it's not used. Seems like it was originally copied from LXD.

PR #487, opened 2025-09-19
fix: make node join more fault tolerance

louiseschmidtgen:
# Make node join more fault tolerant
This PR improves the fault tolerance of node joins in microcluster, allowing nodes to successfully join even when the original control plane leader is down or unreachable.

## Key Changes

Trust Propagation: Modified daemon startup to propagate trust only to reachable members, ensuring joins succeed even if some existing cluster members are unavailable.

Join Token Validation: Enhanced node endpoints to validate that join token addresses are reachable before proceeding with cluster operations. If all token join addresses are unreachable we fail the join operation early. 

Database Resilience: Improved dqlite timeout handling and cleanup for unreachable nodes during partial join scenarios.

Test Coverage: Added comprehensive test cases covering both join-before and join-after failure scenarios with retry mechanisms.

## Behavior

- Node joins now succeed when original leader is down (similar to k8s-snap behavior)
- Failed connections to unreachable members no longer block the entire join process
- Trust propagation continues to all reachable nodes instead of failing on first unreachable member
- Improved error handling for connection refused and network timeout scenarios

## Context: bug discovery in the k8s-snap

While testing node joins in Canonical Kubernetes I encountered this error message on node join:

```
sudo k8s join-cluster <token>

Joining the cluster. This may take a few seconds, please wait.
Error: Failed to join the cluster using the provided token.

The error was: failed after potential retry: wait check failed: failed to POST /k8sd/cluster/join: failed to join k8sd cluster as control plane: Failed to confirm new member "blue" on any existing system (3): Post "https://192.168.105.48:6400/core/internal/truststore": Unable to connect to "192.168.105.48:6400": dial tcp 192.168.105.48:6400: connect: no route to host
```

The node joining (192.168.105.52) is the fourth node in a three node cluster where one of the nodes (192.168.105.48) was killed but not removed from the cluster for testing purposes.

A node join should not fail when one of the nodes in the cluster crashes, we should fail over and contact the other nodes in the cluster.

As a consequence our cluster ends up in an inconsistent state:

core_cluster_members: holds original three nodes including the crashed one.
dqlite: has 4 members including our new one as spare.

```
sudo /snap/k8s/current/bin/k8sd sql   --state-dir /var/snap/k8s/common/var/lib/k8sd/state   "SELECT name, role FROM core_cluster_members"
[[green spare] [yellow voter] [red voter]] 

which is 192.168.105.48, 192.168.105.50, 192.168.105.51

sudo cat /var/snap/k8s/common/var/lib/k8sd/state/database/cluster.yaml
- ID: 3297041220608546238
  Address: 192.168.105.48:6400
  Role: 0
- ID: 2600311347157639893
  Address: 192.168.105.50:6400
  Role: 0
- ID: 17391627959315830676
  Address: 192.168.105.51:6400
  Role: 0
- ID: 4234120578868796636
  Address: 192.168.105.52:6400
  Role: 2
```

### DIFF
--- a/example/test/main.sh
+++ b/example/test/main.sh
@@ -176,17 +176,256 @@ test_recover() {
   shutdown_systems
 }
 
+test_join_token_after_cluster_formed() {
+  # Test node join succeeds when original control plane is down
+  # Token generated AFTER the 3-node cluster is formed
+  # Based on k8s-snap test: test_node_join_succeeds_when_original_control_plane_is_down
+  
+  echo "Starting join test - token generated after 3-node cluster formed"
+  
+  new_systems 4 --heartbeat 2s
+  
+  # Bootstrap initial cluster with c1 as original control plane
+  microctl --state-dir "${test_dir}/c1" init "c1" 127.0.0.1:9001 --bootstrap
+  
+  # Get join tokens for c2 and c3 while c1 is available
+  token_c2=$(microctl --state-dir "${test_dir}/c1" tokens add "c2")
+  token_c3=$(microctl --state-dir "${test_dir}/c1" tokens add "c3")
+  
+  # Join c2 and c3 to form 3-node cluster
+  microctl --state-dir "${test_dir}/c2" init "c2" 127.0.0.1:9002 --token "${token_c2}"
+  microctl --state-dir "${test_dir}/c3" init "c3" 127.0.0.1:9003 --token "${token_c3}"
+  
+  # Wait for cluster to stabilize
+  while [[ -n "$(microctl --state-dir "${test_dir}/c1" cluster list -f yaml | yq '.[] | select(.role == "PENDING")')" ]]; do
+    sleep 2
+  done
+  
+  echo "Initial 3-node cluster formed:"
+  microctl --state-dir "${test_dir}/c1" cluster list
+  
+  # Verify all nodes are voters in the 3-node cluster
+  [[ $(microctl --state-dir "${test_dir}/c1" cluster list -f yaml | yq '.[] | select(.clustermemberlocal.name == "c1").role') == "voter" ]]
+  [[ $(microctl --state-dir "${test_dir}/c1" cluster list -f yaml | yq '.[] | select(.clustermemberlocal.name == "c2").role') == "voter" ]]
+  [[ $(microctl --state-dir "${test_dir}/c1" cluster list -f yaml | yq '.[] | select(.clustermemberlocal.name == "c3").role') == "voter" ]]
+  
+  # Get join token for c4 while c1 is still available
+  token_c4=$(microctl --state-dir "${test_dir}/c1" tokens add "c4")
+  
+  echo "Simulating original control plane (c1) failure..."
+  
+  # Kill c1 to simulate original control plane failure
+  c1_pid=$(jobs -p | head -1)  # Get first background job (should be c1)
+  kill -9 "${c1_pid}" 2>/dev/null || true
+  
+  sleep 2
+  
+  echo "Attempting to join c4 while c1 is down (this tests fault tolerance)..."
+  
+  # This is the critical test: can c4 join while c1 is down?
+  # This should work with proper fault tolerance implementation
+  microctl --state-dir "${test_dir}/c4" init "c4" 127.0.0.1:9004 --token "${token_c4}"
+  
+  echo "Verifying cluster state from surviving nodes..."
+  
+  # Verify cluster from c2's perspective (c1 should be unreachable, c2/c3/c4 should be online)
+  microctl --state-dir "${test_dir}/c2" cluster list
+  
+  # Wait for roles to stabilize after c1 failure and c4 join
+  sleep 5
+  
+  # Count online nodes from c2's perspective
+  online_count=$(microctl --state-dir "${test_dir}/c2" cluster list -f yaml | yq '[.[] | select(.status == "ONLINE")] | length')
+  echo "Online nodes count: ${online_count}"
+  
+  # Should have 3 online nodes (c2, c3, c4) even though c1 is down
+  [[ "${online_count}" -eq 3 ]] || {
+    echo "ERROR: Expected 3 online nodes, got ${online_count}"
+    microctl --state-dir "${test_dir}/c2" cluster list
+    return 1
+  }
+  
+  # Verify c4 successfully joined
+  c4_status=$(microctl --state-dir "${test_dir}/c2" cluster list -f yaml | yq '.[] | select(.clustermemberlocal.name == "c4").status')
+  [[ "${c4_status}" == "ONLINE" ]] || {
+    echo "ERROR: c4 should be ONLINE, got ${c4_status}"
+    return 1
+  }
+  
+  # Verify c4 is a voter, not just PENDING
+  # Add retry logic to wait for role promotion
+  echo "Waiting for c4 to be promoted from PENDING to voter..."
+  retry_count=0
+  max_retries=10
+  while [[ "${retry_count}" -lt "${max_retries}" ]]; do
+    c4_role=$(microctl --state-dir "${test_dir}/c2" cluster list -f yaml | yq '.[] | select(.clustermemberlocal.name == "c4").role')
+    if [[ "${c4_role}" == "voter" ]]; then
+      break
+    fi
+    echo "c4 role is still ${c4_role}, waiting for promotion... (attempt $((retry_count + 1))/${max_retries})"
+    sleep 2
+    retry_count=$((retry_count + 1))
+  done
+  
+  [[ "${c4_role}" == "voter" ]] || {
+    echo "ERROR: c4 should be voter, got ${c4_role} after ${max_retries} attempts"
+    microctl --state-dir "${test_dir}/c2" cluster list
+    return 1
+  }
+
+  # Verify dqlite cluster.yaml shows 4 members (low-level dqlite validation)
+  dqlite_cluster_count=$(yq '. | length' "${test_dir}/c2/database/cluster.yaml")
+  [[ "${dqlite_cluster_count}" -eq 4 ]] || {
+    echo "ERROR: Expected exactly 4 members in dqlite cluster.yaml, got ${dqlite_cluster_count}"
+    echo "Dqlite cluster.yaml contents:"
+    cat "${test_dir}/c2/database/cluster.yaml"
+    return 1
+  }
+
+  echo "SUCCESS: Node c4 successfully joined cluster while original control plane c1 was down"
+  echo "SUCCESS: Node c4 is ONLINE and has voter role"
+  echo "Final cluster state:"
+  microctl --state-dir "${test_dir}/c2" cluster list
+  echo "Dqlite cluster members:"
+  cat "${test_dir}/c2/database/cluster.yaml"
+  
+  shutdown_systems
+}
+
+test_join_token_before_cluster_formed() {
+  # Test node join succeeds when original control plane is down
+  # Token generated BEFORE other nodes join (while c1 is solo)
+  # Based on k8s-snap test: test_node_join_succeeds_when_original_control_plane_is_down
+  
+  echo "Starting join test - token generated before cluster formed"
+  
+  new_systems 4 --heartbeat 2s
+  
+  # Bootstrap initial cluster with c1 as original control plane
+  microctl --state-dir "${test_dir}/c1" init "c1" 127.0.0.1:9001 --bootstrap
+  
+  # Get join tokens for c2 and c3 while c1 is available
+  token_c2=$(microctl --state-dir "${test_dir}/c1" tokens add "c2")
+  token_c3=$(microctl --state-dir "${test_dir}/c1" tokens add "c3")
+  # Get join token for c4 while c1 is the only node up
+  token_c4=$(microctl --state-dir "${test_dir}/c1" tokens add "c4")
+  
+  # Join c2 and c3 to form 3-node cluster
+  microctl --state-dir "${test_dir}/c2" init "c2" 127.0.0.1:9002 --token "${token_c2}"
+  microctl --state-dir "${test_dir}/c3" init "c3" 127.0.0.1:9003 --token "${token_c3}"
+  
+  # Wait for cluster to stabilize
+  while [[ -n "$(microctl --state-dir "${test_dir}/c1" cluster list -f yaml | yq '.[] | select(.role == "PENDING")')" ]]; do
+    sleep 2
+  done
+  
+  echo "Initial 3-node cluster formed:"
+  microctl --state-dir "${test_dir}/c1" cluster list
+  
+  # Verify all nodes are voters in the 3-node cluster
+  [[ $(microctl --state-dir "${test_dir}/c1" cluster list -f yaml | yq '.[] | select(.clustermemberlocal.name == "c1").role') == "voter" ]]
+  [[ $(microctl --state-dir "${test_dir}/c1" cluster list -f yaml | yq '.[] | select(.clustermemberlocal.name == "c2").role') == "voter" ]]
+  [[ $(microctl --state-dir "${test_dir}/c1" cluster list -f yaml | yq '.[] | select(.clustermemberlocal.name == "c3").role') == "voter" ]]
+  
+  echo "Simulating original control plane (c1) failure..."
+  
+  # Kill c1 to simulate original control plane failure
+  c1_pid=$(jobs -p | head -1)  # Get first background job (should be c1)
+  kill -9 "${c1_pid}" 2>/dev/null || true
+  
+  sleep 2
+  
+  echo "Attempting to join c4 while c1 is down (this should fail cleanly)..."
+  
+  # This is the critical test: c4 should FAIL to join when c1 is down
+  # and the token was generated before cluster formation
+  # The join should fail cleanly without creating partial state
+  ! microctl --state-dir "${test_dir}/c4" init "c4" 127.0.0.1:9004 --token "${token_c4}" || {
+    echo "ERROR: c4 join should have failed but succeeded"
+    return 1
+  }
+  
+  echo "c4 join failed as expected. Verifying clean failure..."
+  
+  # Verify cluster from c2's perspective (should only show c1 as unreachable, c2/c3 as online)
+  microctl --state-dir "${test_dir}/c2" cluster list
+  
+  # Wait a bit to see if c4 appears in cluster or fails cleanly
+  sleep 5
+  
+  # Count online nodes from c2's perspective - should only be c2 and c3
+  online_count=$(microctl --state-dir "${test_dir}/c2" cluster list -f yaml | yq '[.[] | select(.status == "ONLINE")] | length')
+  echo "Online nodes count: ${online_count}"
+  
+  # Should have only 2 online nodes (c2, c3) since c1 is down and c4 should fail to join
+  [[ "${online_count}" -eq 2 ]] || {
+    echo "ERROR: Expected 2 online nodes (c2, c3), got ${online_count}"
+    microctl --state-dir "${test_dir}/c2" cluster list
+    return 1
+  }
+  
+  # Verify c4 is NOT in the cluster members table (clean failure)
+  c4_exists=$(microctl --state-dir "${test_dir}/c2" cluster list -f yaml | yq '.[] | select(.clustermemberlocal.name == "c4")' | wc -l)
+  [[ "${c4_exists}" -eq 0 ]] || {
+    echo "ERROR: c4 should not appear in cluster members table (partial join detected)"
+    echo "c4 state in cluster:"
+    microctl --state-dir "${test_dir}/c2" cluster list -f yaml | yq '.[] | select(.clustermemberlocal.name == "c4")'
+    return 1
+  }
+  
+  # Verify that only c1, c2, c3 exist in cluster (no c4)
+  member_count=$(microctl --state-dir "${test_dir}/c2" cluster list -f yaml | yq '. | length')
+  [[ "${member_count}" -eq 3 ]] || {
+    echo "ERROR: Expected exactly 3 cluster members (c1, c2, c3), got ${member_count}"
+    microctl --state-dir "${test_dir}/c2" cluster list
+    return 1
+  }
+  
+  # Verify dqlite cluster.yaml shows only 3 members (low-level dqlite validation)
+  dqlite_cluster_count=$(yq '. | length' "${test_dir}/c2/database/cluster.yaml")
+  [[ "${dqlite_cluster_count}" -eq 3 ]] || {
+    echo "ERROR: Expected exactly 3 members in dqlite cluster.yaml, got ${dqlite_cluster_count}"
+    echo "Dqlite cluster.yaml contents:"
+    cat "${test_dir}/c2/database/cluster.yaml"
+    return 1
+  }
+  
+  # Verify c4 (127.0.0.1:9004) is NOT in dqlite cluster.yaml
+  c4_in_dqlite_yaml=$(yq '.[] | select(.Address == "127.0.0.1:9004")' "${test_dir}/c2/database/cluster.yaml" | wc -l)
+  [[ "${c4_in_dqlite_yaml}" -eq 0 ]] || {
+    echo "ERROR: c4 found in dqlite cluster.yaml (partial join detected at dqlite level)"
+    echo "Dqlite cluster.yaml contents:"
+    cat "${test_dir}/c2/database/cluster.yaml"
+    return 1
+  }
+  
+  echo "SUCCESS: Node c4 failed to join cleanly - no partial join state detected"
+  echo "SUCCESS: Verified at both microcluster API and go-dqlite cluster members"
+  echo "Final cluster state (c4 should not appear):"
+  microctl --state-dir "${test_dir}/c2" cluster list
+  echo "Dqlite cluster members:"
+  cat "${test_dir}/c2/database/cluster.yaml"
+  
+  shutdown_systems
+}
+
 # allow for running a specific set of tests
 if [ "${1:-"all"}" = "all" ] || [ "${1}" = "" ]; then
   test_misc
   test_tokens
   test_recover
+  test_join_token_after_cluster_formed
+  test_join_token_before_cluster_formed
 elif [ "${1}" = "recover" ]; then
   test_recover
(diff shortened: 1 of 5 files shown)


internal/daemon/daemon.go
```diff
@@ -655,16 +655,17 @@ func (d *Daemon) StartAPI(ctx context.Context, bootstrap bool, initConfig map[st
...
 			// At this point the joiner is only trusted on the node that was leader at the time,
-			// so find it and have it instruct all dqlite members to trust this system now that it is functional.
-			if !clusterConfirmation {
-				err := internalClient.AddTrustStoreEntry(ctx, &c.Client, localMemberInfo)
-				if err != nil {
-					lastErr = err
-				} else {
-					clusterConfirmation = true
-				}
+			// so propagate trust to all reachable cluster members for fault tolerance.
```
roosterfish: I see that the comment is misleading, as we are not trying to find any leader.

However don't we already reach out to all existing cluster members here? Because `cluster.Query` is concurrently (`concurrent=true`) running `internalClient.AddTrustStoreEntry` for each of the existing cluster members (except ourself).
It seems a bit racy to me though as its unclear when `clusterConfirmation` is set to true. But I would argue most of the times it will run it for each of the members.

I agree let's get rid of the `if` condition.

internal/state/state.go
```diff
@@ -165,35 +165,28 @@ func (s *InternalState) HasExtension(ext string) bool {
...
-	for _, clusterMember := range clusterMembers {
-		if s.Address().URL.Host == clusterMember.Address.String() {
-			continue
-		}
-
-		publicKey, err := s.ClusterCert().PublicKeyX509()
-		if err != nil {
-			return nil, err
+	// Filter out ourselves from the client list
+	clients := make(client.Cluster, 0, len(allClients))
```
roosterfish: `len(allClients)-1` as we are filtering out ourselves?

internal/state/state.go
```diff
@@ -165,35 +165,28 @@ func (s *InternalState) HasExtension(ext string) bool {
...
+	for _, client := range allClients {
+		if s.Address().URL.Host != client.URL().URL.Host {
+			clients = append(clients, client)
 		}
-
-		url := api.NewURL().Scheme("https").Host(clusterMember.Address.String())
-		c, err := internalClient.New(*url, s.ServerCert(), publicKey, isNotification)
```
roosterfish, line 191: I like the fact we can use `remotes.Cluster` in order to create the list of clients to omit duplicating the code here a second time.

internal/trust/remotes.go
```diff
@@ -270,6 +272,12 @@ func (r *Remotes) Addresses() map[string]types.AddrPort {
 func (r *Remotes) Cluster(isNotification bool, serverCert *shared.CertInfo, publicKey *x509.Certificate) (client.Cluster, error) {
 	cluster := make(client.Cluster, 0, r.Count()-1)
 	for _, addr := range r.Addresses() {
+		// Filter out unreachable nodes with a quick connectivity check
```
roosterfish: I am not too sure about this one. 
You can call `.Query` on the returned `client.Cluster` to concurrently run a func for each of the cluster members. You can also provide a context to deal with offline cluster members.

Can you please explain in which situation you see issues with this?

roosterfish (changes requested):
Thanks for this! I like the fact that we can make the inter cluster communication more bulletproof. 

Please check my comment on the `isReachable` func.

roosterfish:
Also please sign-off your commits and prefix the commit messages with the respective Go package name. Furthmore please split the changes into their own commits.
See those commits from one of the last merged PRs as reference https://github.com/canonical/microcluster/pull/472/commits.

internal/daemon/daemon.go
louiseschmidtgen: Agreed, this section is not great and racy. 
Ideally, you would have something like raft that handles this truststore change. Right now we have the heartbeats that sync the truststore on every beat. 
We could possibly overwrite the "new member" if it wasn't persisted on the node sending out the heartbeats. Right now this is a best effort attempt to get the truststore entry added.
It would be a bigger fix to handle the race conditions. What do you think?

roosterfish: Yes I think what you currently have in the commit is fine: Remove the condition to ensure we always do it on all known members.

internal/trust/remotes.go
louiseschmidtgen: Yes, you're right here. I went back in the code and this function gets used in the heartbeat logic as well where it should figure out itself that the member is unreachable. As a safer means, I've made a second ReachableCluster function that can be called in the places where we need it.

Let me know your thoughts!

internal/daemon/daemon.go
louiseschmidtgen: Yes, I see now that the logic is a bit spaghetti. 
We call the maybe add truststore entry function from all known node clients then the trust post call has each one of those clients reach out to the cluster leader to make the change. 
So at the end of the day all clients will tell the leader to update the truststore and that change gets sent out on the heartbeat to all nodes.
So as long as one of these calls succeeds we're actually fine.

internal/daemon/daemon.go
```diff
@@ -630,7 +630,7 @@ func (d *Daemon) StartAPI(ctx context.Context, bootstrap bool, initConfig map[st
 		return err
 	}
 
-	cluster, err := d.trustStore.Remotes().Cluster(false, d.ServerCert(), publicKey)
+	cluster, err := d.trustStore.Remotes().ReachableCluster(false, d.ServerCert(), publicKey)
```
roosterfish: Can you please add the commit that introduces `ReachableCluster` before using it in this commit? This makes it easier to follow through the commits as you only reference stuff which already exists at this point.

internal/db/db.go
```diff
@@ -27,7 +27,9 @@ import (
 // Open opens the dqlite database and loads the schema.
 // Returns true if we need to wait for other nodes to catch up to our version.
 func (db *DqliteDB) Open(ext extensions.Extensions, bootstrap bool) error {
-	ctx, cancel := context.WithTimeout(db.ctx, 30*time.Second)
+	// Allow dqlite up to 2 minutes to become ready when starting up.
+	// This is to allow for unready/dead nodes to time out.
+	ctx, cancel := context.WithTimeout(db.ctx, 120*time.Second)
```
roosterfish, line 32: Why is this a problem if we return early at this stage? I don't understand this line `This is to allow for unready/dead nodes to time out.`.

Is it problematic to return early here or is this just a safety measure?

internal/db/db.go
```diff
@@ -41,6 +43,13 @@ func (db *DqliteDB) Open(ext extensions.Extensions, bootstrap bool) error {
 		db.statusLock.Lock()
 		db.status = types.DatabaseOffline
 		db.statusLock.Unlock()
+		// close the db if any of the following steps fail
+		closeErr := db.dqlite.Close()
```
roosterfish, line 47: Why try closing if we haven't yet called `db.dqlite.Open`?

But we should definitely move this section before `err = db.waitUpgrade(bootstrap, ext)` as this might error out and leave it open? What do you think?

internal/state/state.go
```diff
@@ -165,35 +168,53 @@ func (s *InternalState) HasExtension(ext string) bool {
...
+	// Use trust store instead of database - it's updated on heartbeats
+	// and is more likely to reflect current reachable cluster state
+	remotes := s.Remotes()
+	allReachableClients, err := remotes.ReachableCluster(isNotification, s.ServerCert(), publicKey)
+	if err != nil {
+		return nil, err
+	}
 
-		clients = append(clients, client.Client{Client: *c})
+	// Filter out ourselves from the client list
```
roosterfish: Can we please have another unexported func e.g. `func (s *InternalState) filterClients` that is called by both `ReachableCluster` and `Cluster` to prevent duplicating this filtering?

internal/rest/resources/control.go
```diff
@@ -183,12 +185,40 @@ func controlPost(state state.State, r *http.Request) response.Response {
...
 func joinWithToken(state state.State, r *http.Request, req *internalTypes.Control) (*internalTypes.TokenResponse, *trust.Remote, error) {
 	token, err := internalTypes.DecodeToken(req.JoinToken)
 	if err != nil {
 		return nil, nil, err
 	}
 
+	// If none of the addresses are reachable, fail fast.
+	reachableJoinAddresses := filterReachableJoinAddresses(token.JoinAddresses)
+	if len(reachableJoinAddresses) == 0 {
+		return nil, nil, fmt.Errorf("none of the provided join token addresses are reachable")
```
roosterfish: Please start the error with uppercase.

internal/rest/resources/truststore.go
```diff
@@ -52,7 +52,7 @@ func trustPost(s state.State, r *http.Request) response.Response {
 	defer cancel()
 
 	if !client.IsNotification(r) {
-		cluster, err := s.Cluster(true)
+		cluster, err := s.ReachableCluster(true)
```
roosterfish: What is the risk of this? 
I understand that offline cluster members won't receive a notification about new cluster members so they won't update their trust store until the next heartbeat in which they are online again?

example/test/main.sh
```diff
@@ -176,17 +176,245 @@ test_recover() {
   shutdown_systems
 }
 
+test_join_token_after_cluster_formed() {
```
roosterfish, line 179: Can you please move this commit at the end of the PR? 

We don't squash when merge so this commit will appear in the log before the actual fixes.

internal/state/state.go
```diff
@@ -165,35 +168,53 @@ func (s *InternalState) HasExtension(ext string) bool {
 // this one.
 // All requests made by the client will have the UserAgentNotifier header set
 // if isNotification is true.
+// Uses the trust store instead of database for better fault tolerance -
+// trust store is updated on heartbeats and shouldn't contain crashed nodes.
 func (s *InternalState) Cluster(isNotification bool) (client.Cluster, error) {
-	c, err := s.Leader()
+	publicKey, err := s.ClusterCert().PublicKeyX509()
```
roosterfish, line 171: Please move the modifications to `Cluster` out into its own commit.

internal/rest/resources/truststore.go
roosterfish: This should be the same as today if I am not mistaken just with the twist that we don't even try to reach out to offline members.

internal/trust/remotes.go
```diff
@@ -282,6 +284,38 @@ func (r *Remotes) Cluster(isNotification bool, serverCert *shared.CertInfo, publ
 	return cluster, nil
 }
 
+// ReachableCluster returns a set of clients for every reachable remote, which can be concurrently queried.
+func (r *Remotes) ReachableCluster(isNotification bool, serverCert *shared.CertInfo, publicKey *x509.Certificate) (client.Cluster, error) {
```
roosterfish: Please land this func in its own commit before using it.

internal/trust/remotes.go
```diff
@@ -282,6 +284,38 @@ func (r *Remotes) Cluster(isNotification bool, serverCert *shared.CertInfo, publ
...
+	cluster := make(client.Cluster, 0, r.Count()-1)
+	for _, addr := range r.Addresses() {
+		if !isReachable(addr) {
+			continue
+		}
+
+		url := api.NewURL().Scheme("https").Host(addr.String())
```
roosterfish: Mhh, the only difference in this func is the check for `isReachable`. 
Should this rather be added to the existing `Cluster` func by adding another argument `reachable bool` to the func? 

`Cluster` is currently only has two callers so this change shouldn't be too invasive.

roosterfish (changes requested):
I have now looked more deeper into the code to understand the root cause of the issues you are seeing.

Whilst I recognize we have to put up some fixes in order to allow cluster joins with offline members, I am still not confident we should do the connection check upfront (see [this](https://github.com/canonical/microcluster/pull/487#discussion_r2410046464) comment) to get a potential delta of cluster members which are reachable. Let me explain my reasoning behind this. 

Right now the problem is that we are too forceful when it comes to errors during trust propagation (and hook execution, see following comment).
The loop which is currently in place works fine, however it exits prematurely if some members are offline and we are not able to propagate the new member as pointed out in your PR description.
We can fix this by instead of returning an error simply log the error (as we do in other places during the join). So change this line to instead log an error https://github.com/canonical/microcluster/blob/v3/internal/rest/resources/truststore.go#L66.
Now there is one more place where we have to log an error instead of returning it. This is when calling the `on-new-member` hook. Also log an error [here](https://github.com/canonical/microcluster/blob/v3/internal/daemon/daemon.go#L710). This is in line with what we do when failing to send the [upgrade notification](https://github.com/canonical/microcluster/blob/v3/internal/daemon/daemon.go#L690).

By being a bit more careful on those two errors, we can already address the join issue. But in addition let's also propagate trust to all members regardless whether or not someone already set `clusterConfirmation` (as you currently have in your commit to make this section less racy).

The reason I don't like the upfront connection check is because right now we proactively try to execute an action, and if it fails due to connection issues we continue on. 
However when we first aggregate a list, we don't know if the members are still online when actually performing the action so whilst this check seems ok at first sight, it doesn't accommodate all scenarios. Whereas the current approach simply continues on until the operation was successful (when putting in the mentioned error logging fixes).

Overall there should not be the need for a `ReachableCluster` func, but instead the caller should carefully handle errors when they occur and either return them to its caller or log them.

internal/db/db.go
louiseschmidtgen: This context of 30 seconds did not allow Dqlite enough time to perform the db.Ready() call when one member is down. Without this change the db.Ready call is cancelled before it completes.

internal/db/db.go
louiseschmidtgen: The new node is already present in the dqlite cluster at this point (since the call to dqlite.New), if we don't clean up the db here if db.Ready or db.Open fail the new member is still in the dqlite cluster and does not get cleaned up.

internal/trust/remotes.go
louiseschmidtgen: No longer applicable as I removed Reachable Cluster function.

internal/trust/remotes.go
louiseschmidtgen: No longer applicable as I removed Reachable Cluster function.

internal/state/state.go
louiseschmidtgen: No longer applicable as I removed Reachable Cluster function.

internal/rest/resources/truststore.go
louiseschmidtgen: No longer applicable as I removed Reachable Cluster function.

internal/state/state.go
louiseschmidtgen: No longer applicable as I removed Reachable Cluster function.

internal/daemon/daemon.go
louiseschmidtgen: No longer applicable as I removed Reachable Cluster function.

internal/trust/remotes.go
louiseschmidtgen: No longer applicable as I removed Reachable Cluster function.

internal/daemon/daemon.go
```diff
@@ -707,16 +717,24 @@ func (d *Daemon) StartAPI(ctx context.Context, bootstrap bool, initConfig map[st
...
 			err = internalClient.RunNewMemberHook(ctx, c.Client.UseTarget(remote.Name), internalTypes.HookNewMemberOptions{NewMember: localMemberInfo})
 			if err != nil && !api.StatusErrorCheck(err, http.StatusServiceUnavailable) {
-				return err
+				// log error but continue with other nodes
+				logger.Warn("Failed running OnNewMember hook on node", logger.Ctx{"node": c.URL().URL.Host, "error": err})
+				return nil
 			}
 		}
 
+		successCount++
```
roosterfish, line 729: Do we require a lock here?

internal/daemon/daemon.go
```diff
@@ -707,16 +717,24 @@ func (d *Daemon) StartAPI(ctx context.Context, bootstrap bool, initConfig map[st
...
 		return nil
 	})
 	if err != nil {
 		return err
 	}
 
+	// Only fail if we attempted to notify other nodes but all failed
+	if attemptCount > 0 && successCount == 0 {
+		return fmt.Errorf("Failed to notify any existing cluster member at %q", strings.Join(joinAddresses, ", "))
```
roosterfish, line 739: Can we ever reach this error? When trying to join a cluster but all the members are offline, will it actually be responsive?

internal/db/db.go
roosterfish: Is this caused by another 30s timeout we configure at a different place? I am thinking about something like https://github.com/canonical/microcluster/pull/512.

internal/db/db.go
roosterfish: Got you, it's done right before calling this `Open` func. Do we also have to `db.db.Close()` or is this done by `db.dqlite.Close()`? Just so that we don't accidentally leak something which is still open.

internal/rest/resources/control.go
```diff
@@ -183,12 +185,40 @@ func controlPost(state state.State, r *http.Request) response.Response {
...
+}
+
 func joinWithToken(state state.State, r *http.Request, req *internalTypes.Control) (*internalTypes.TokenResponse, *trust.Remote, error) {
 	token, err := internalTypes.DecodeToken(req.JoinToken)
 	if err != nil {
 		return nil, nil, err
 	}
 
+	// If none of the addresses are reachable, fail fast.
+	reachableJoinAddresses := filterReachableJoinAddresses(token.JoinAddresses)
```
roosterfish: I would rather like to be relying on the context and not do the upfront connection check (see my reasoning here https://github.com/canonical/microcluster/pull/487#pullrequestreview-3322034297).
This would be the only place in which we would perform such a check.

Instead rather set some concrete context deadlines in the `for _, addr := range token.JoinAddresses {` loop below. Let's go with the 30s we always use. And if this collides with any context set by the caller (client func), let's approach it the same way as in https://github.com/canonical/microcluster/pull/512/commits/0c6b339a8ec1d1c2960abbffd78cebcaf775d878.

internal/rest/resources/truststore.go
```diff
@@ -57,17 +58,37 @@ func trustPost(s state.State, r *http.Request) response.Response {
...
+			successCount++
+			return nil
 		})
 		if err != nil {
 			return response.SmartError(err)
 		}
+
+		// Only fail if we attempted to propagate to other nodes but all failed
+		if attemptCount > 0 && successCount == 0 {
+			return response.SmartError(fmt.Errorf("Failed adding truststore entry to any cluster node"))
```
roosterfish, line 96: Same comment as before, can we ever reach this statement with a still responsive cluster?

internal/rest/resources/truststore.go
```diff
@@ -57,17 +58,37 @@ func trustPost(s state.State, r *http.Request) response.Response {
...
+		// Try to add the truststore entry to all other nodes in the cluster.
+		// We don't fail the entire operation if some nodes are unreachable.
 		err = cluster.Query(ctx, true, func(ctx context.Context, c *client.Client) error {
 			// No need to send a request to ourselves, or to the node we are adding.
 			if s.Address().URL.Host == c.URL().URL.Host || req.Address.String() == c.URL().URL.Host {
 				return nil
 			}
 
-			return internalClient.AddTrustStoreEntry(ctx, &c.Client, req)
+			attemptCount++
```
roosterfish, line 75: `cluster.Query` is marked to run the funcs in parallel so this would require locking.

roosterfish (changes requested):

internal/daemon/daemon.go
louiseschmidtgen: I added this check in case attempts = 0 and successCount = 0 which happens when we bootstrap the node.

For the error message, it feels unsafe not to check that at least one of the queries succeeded considering we only log the error before continuing with the logic. If you don't think that is necessary I can remove it.

internal/db/db.go
louiseschmidtgen: Yes, `db.db` would be `db.dqlite` so yes that is the same. I am calling `db.dqlite` since `db.db` might not be assigned yet.

internal/db/db.go
louiseschmidtgen: In this call, we cancel the context prematurely because dqlite's Open function needs more time when one member is down. By moving the db.Close command we at least properly revert state. There are timeouts on Dqlite's side which can take place when we allow for a longer context. This allows dqlite's open function to succeed.

I think it makes sense to leave the increased timeout in this place close to the Open call which requires it. Let me know your thoughts please.

internal/rest/resources/truststore.go
louiseschmidtgen: Added mutex around counters marking as resolved.

internal/daemon/daemon.go
louiseschmidtgen: Added mutex around counters marking as resolved.

internal/db/db.go
roosterfish: All right, thanks.

internal/db/db.go
roosterfish: Can we do something like 60s instead? This would allow 30s for the `Ready` and another 30s for the actual `Open`.

internal/daemon/daemon.go
roosterfish: I see thanks.

internal/daemon/daemon.go
```diff
@@ -723,18 +725,24 @@ func (d *Daemon) StartAPI(ctx context.Context, bootstrap bool, initConfig map[st
...
 		successCount++
+		counterMu.Unlock()
 		return nil
 	})
 	if err != nil {
 		return err
 	}
 
 	// Only fail if we attempted to notify other nodes but all failed
+	counterMu.Lock()
```
roosterfish: I think we don't need any lock here as we have exclusive access to `attemptCount` and `successCount` if `cluster.Query` has returned.

internal/rest/resources/truststore.go
```diff
@@ -78,17 +82,23 @@ func trustPost(s state.State, r *http.Request) response.Response {
...
 			successCount++
+			counterMu.Unlock()
 			return nil
 		})
 		if err != nil {
 			return response.SmartError(err)
 		}
 
 		// Only fail if we attempted to propagate to other nodes but all failed
+		counterMu.Lock()
```
roosterfish: Same here, at this stage we have exclusive access to `attemptCount` and `successCount`.

internal/daemon/daemon.go
```diff
@@ -683,8 +683,8 @@ func (d *Daemon) StartAPI(ctx context.Context, bootstrap bool, initConfig map[st
 	// We don't fail the entire operation if some nodes are unreachable.
```
roosterfish, line 683: Please squash the two mutex commits into their respective original commit where you add the counters. 
We don't squash commits during merge.

internal/rest/resources/control.go
roosterfish: Can we please don't have `isReachable`? Is this feasible?

roosterfish (changes requested):

internal/db/db.go
louiseschmidtgen: The `dqlite.Open()` call already has a 60 second sleep/timeout in its retry loop.

Then the `dqlite.Ready()` call performs a couple of things that it waits for: new node added, new node promoted from spare-> voter, things can take a bit longer when one node in the cluster is unresponsive. 

I think it makes sense to allow for a generous timeout to allow dqlite to start properly even under sub-par conditions: node down/slow network/leader elections etc.

internal/rest/resources/control.go
louiseschmidtgen: From my microcluster test case it looks possible. I want to run the k8s e2e scenario again to confirm it doesn't cause problems.

internal/daemon/daemon.go
louiseschmidtgen: Of course, done.

roosterfish (approved):
LGTM! Thanks.

internal/db/db.go
roosterfish, line 47: @louiseschmidtgen This caused a regression and broke the schema upgrade procedure, see bug report https://github.com/canonical/microcluster/issues/544.

`Open` is called in a retry loop by the `Join` func to wait until all cluster members upgrade their schema versions. When doing `dqlite.Close()` here, any subsequent attempts to `Open` will fail and cause a panic.
We have to `db.db.Close()` I guess. 

Shouldn't the reverter be added right after the connection attempt? Like this:

```go
if db.db == nil {
		db.db, err = db.dqlite.Open(db.ctx, db.dbName)
		if err != nil {
			return fmt.Errorf("Open dqlite: %w", err)
		}

		reverter.Add(func() {
			closeErr := db.db.Close()
			if closeErr != nil {
				logger.Error("Failed to close database", logger.Ctx{"address": db.listenAddr.String(), "error": closeErr})
			}

			db.db = nil
		})
	}
```

louiseschmidtgen: Hi @roosterfish thanks for looking into the bug.

Did you run my e2e tests from this PR?

IIRC this was necessary for proper failed join cleanup. AFAIK, the moment we call the [Dqlite app](https://github.com/canonical/microcluster/blob/v3/internal/db/dqlite.go#L153) the DB exists and won't be properly closed/removed unless the reverter takes care of it.

I need to look at the join path maybe we can be more precise in the cleanup. Wdyt?

roosterfish: What might work is adding an extra reverter to `Join` which jumps in when the `for { .. }` loop exits and then runs `dqlite.Close`. 

The main problem I see is that we are closing something inside `Open` which was created outside (in `Join`). 
It likely was a good call to move the reverter inside `Open` a few lines up before `waitUpgrade` but let's only add the revert if we actually performed the `db.dqlite.Open` call.

louiseschmidtgen: I think you're spot on for adding an extra reverter for `Join`. As I mentioned async dqlite.Close immediately stops all connections and removes the Dqlite app closing all Raft connections. Dqlite essentially thinks this node is no longer part of the cluster.
We don't want that for a normal `Open` as apparent in this schemaUpdate logic.
Here we do want the db.db.Close() which is a graceful shutdown of db connections. We don't need to rip it our of Dqlite.

louiseschmidtgen: Linking PR: https://github.com/canonical/microcluster/pull/547

Issue #282, opened 2024-11-06
Grow our own `response` package as in LXD


roosterfish:
Recently we ran into trouble with the dependency to the `lxd/response` package downstreams have to use when defining their microcluster API handlers.
If they the version used for defining the handler diverges from the one used internally by microcluster you might end up with build errors. 

To overcome this limitation microcluster should have its own implementation of `response` which can also be more lightweight as we don't require all the different response types offered by LXD.

roosterfish:
Fixed in `v3` with the latest refactorings.

PR #502, opened 2025-10-10
Remove from Microcluster the dependency on lxd/response and lxd/request

mseralessandri:
Part of https://github.com/canonical/microcluster/issues/503
Refers also to https://github.com/canonical/microcluster/issues/282
LXD-3636

### DIFF
--- a/rest/response/response.go
+++ b/rest/response/response.go
@@ -1,24 +1,177 @@
 package response
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
 
+	"github.com/canonical/lxd/lxd/util"
 	"github.com/canonical/lxd/shared/api"
 	"github.com/canonical/lxd/shared/logger"
 )
 
-// ParseResponse takes a http response, parses it and returns the extracted result.
+// Init registers smart error mappings.
+func Init(smartErrors map[int][]error) {
+	for code, additionalErrors := range smartErrors {
+		existingErrs, ok := httpResponseErrors[code]
+		if ok {
+			httpResponseErrors[code] = append(existingErrs, additionalErrors...)
+			continue
+		}
+
+		httpResponseErrors[code] = additionalErrors
+	}
+}
+
+// Response represents an API response.
+type Response interface {
+	Render(w http.ResponseWriter, r *http.Request) error
+	String() string
+}
+
+// Sync response.
+type syncResponse struct {
+	success  bool
+	metadata any
+}
+
+// EmptySyncResponse represents an empty success response.
+var EmptySyncResponse = &syncResponse{success: true, metadata: make(map[string]any)}
+
+// SyncResponse returns a new syncResponse with the success and metadata fields set.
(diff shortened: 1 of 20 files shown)


rest/request/request.go
```diff
@@ -0,0 +1,20 @@
...
+type ctxKey string
+
+const (
+	// ctxConnection is the key used to store connection information in request context.
+	ctxConnection ctxKey = "connection"
+)
+
+// SaveConnectionInContext is a ConnContext function that saves the connection
+// in the request context for later retrieval.
+func SaveConnectionInContext(ctx context.Context, c net.Conn) context.Context {
```
roosterfish: Mhh, I wonder if we even use this in Microcluster? Because `ctxConnection` seems to be only used here.

rest/response/response.go
```diff
@@ -1,32 +1,205 @@
...
 	"github.com/canonical/lxd/shared/api"
 	"github.com/canonical/lxd/shared/logger"
+	"github.com/canonical/lxd/lxd/util"
 )
 
-// ParseResponse takes a http response, parses it and returns the extracted result.
+var debug bool
+
+// Init sets the debug variable and registers smart error mappings.
+func Init(d bool, smartErrors map[int][]error) {
```
roosterfish: Looks like we can also get rid of the `debug` option. We only ever call `Init` once during the daemon's `init()` using `debug=false`.

roosterfish (changes requested):
Thanks for this! There seem to be some golangci-lint errors regarding wrong ordering of `import` statements. Also I think we can simply drop some stuff, please check my comments.

rest/response/response.go
```diff
@@ -1,32 +1,205 @@
...
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	w.Header().Set("X-Content-Type-Options", "nosniff")
+
+	if w.Header().Get("Connection") != "keep-alive" {
+		w.WriteHeader(r.code)
+	}
+
+	_, err = fmt.Fprintln(w, buf.String())
```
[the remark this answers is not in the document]
roosterfish: I think this is actually a good catch! We should certainly not change this in v2 to not modify the output behavior of the API, but we can start using this in v3.

go.mod
```diff
@@ -20,6 +20,7 @@ require (
 require (
 	github.com/Rican7/retry v0.3.1 // indirect
 	github.com/armon/go-proxyproto v0.1.0 // indirect
+	github.com/canonical/go-dqlite/v2 v2.0.1 // indirect
```
roosterfish, line 23: Whoops, wondered why this was missed, found the issue, fixed in https://github.com/canonical/microcluster/pull/511.

roosterfish: It's merged, you can now rebase this PR.

roosterfish: Or simply drop the first commit :)

mseralessandri: Done!

roosterfish (approved):

PR #512, opened 2025-10-14
internal/rest/resources: Cleanup lost node as best-effort when `--force`d

bschimke95:
Removing a lost node from a cluster fails right now because the it tries to run the pre-remove hook on the lost node and eventually times out after 30 seconds.

This fix will try to reach the node that is about to be removed before running the pre-remove hook. If it cannot be reached the removal will either fail or continue without running the `pre-remove` hook if forced.

### DIFF
--- a/internal/rest/client/cluster.go
+++ b/internal/rest/client/cluster.go
@@ -10,9 +10,19 @@ import (
 	"github.com/canonical/microcluster/v3/rest/types"
 )
 
+// withTimeoutIfUnset returns a context with a 30s timeout only if the parent context has no deadline set.
+func withTimeoutIfUnset(ctx context.Context) (context.Context, context.CancelFunc) {
+	_, ok := ctx.Deadline()
+	if ok {
+		return ctx, func() {}
+	}
+
+	return context.WithTimeout(ctx, 30*time.Second)
+}
+
 // AddClusterMember records a new cluster member in the trust store of each current cluster member.
 func AddClusterMember(ctx context.Context, c *Client, args types.ClusterMember) (*internalTypes.TokenResponse, error) {
-	queryCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	queryCtx, cancel := withTimeoutIfUnset(ctx)
 	defer cancel()
 
 	tokenResponse := internalTypes.TokenResponse{}
@@ -26,7 +36,7 @@ func AddClusterMember(ctx context.Context, c *Client, args types.ClusterMember) 
 
 // ResetClusterMember clears the state directory of the cluster member, and re-execs its daemon.
 func ResetClusterMember(ctx context.Context, c *Client, name string, force bool) error {
-	queryCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	queryCtx, cancel := withTimeoutIfUnset(ctx)
 	defer cancel()
 
 	endpoint := api.NewURL().Path("cluster", name)
@@ -39,7 +49,7 @@ func ResetClusterMember(ctx context.Context, c *Client, name string, force bool)
 
 // GetClusterMembers returns the database record of cluster members.
 func (c *Client) GetClusterMembers(ctx context.Context) ([]types.ClusterMember, error) {
-	queryCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	queryCtx, cancel := withTimeoutIfUnset(ctx)
 	defer cancel()
 
 	clusterMembers := []types.ClusterMember{}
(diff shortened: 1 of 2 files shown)


internal/rest/resources/cluster.go
```diff
@@ -602,6 +604,7 @@ func clusterMemberDelete(s state.State, r *http.Request) response.Response {
 	}
 
 	// Tell the cluster member to run its PreRemove hook and return.
+	c.Timeout = time.Second * 2
```
roosterfish: Let's rather increase the timeout of the overall operation. Otherwise we only allow the hook to be running for 2 seconds as `Timeout` also includes the time until reading the response body. And this would be a breaking change.
But this also requires making the context on the client func (which calls this API) to be more flexible and not enforce the 30s timeout.

We now came across the rather "fixed" timeout of 30s quite often.
I would suggest for the case of cluster member removal, let's set [this](https://github.com/canonical/microcluster/blob/v3/internal/rest/resources/cluster.go#L417) timeout to 60s. This would still allow the hook to be running for at least 30s (as it's today) to not cause any breaking changes (as we surely want to backport this).

In addition let's modify the [client funcs](https://github.com/canonical/microcluster/blob/v3/internal/rest/client/cluster.go) to only set the 30s timeout if no deadline is already configured on the context. This won't change existing behavior but adds more flexibility to the caller. 

What do you think?

roosterfish (changes requested):

roosterfish:
Posting it here too for context:

Thanks for reporting. I suspect this only happens in case you run into a timeout. If you immediately get something like connection refused, it won't run into the context deadline and consider the `force=true` flag and continue on. 
The latter you can test easily when you bring up multiple Microcluster members on the same node on different ports. Then it works as expected.

internal/rest/resources/cluster.go
```diff
@@ -613,10 +614,9 @@ func clusterMemberDelete(s state.State, r *http.Request) response.Response {
 	err = s.Database().Transaction(r.Context(), func(ctx context.Context, tx *sql.Tx) error {
 		return cluster.DeleteCoreClusterMember(ctx, tx, remote.Address.String())
 	})
-	if err != nil {
+	if err != nil && !force {
```
bschimke95, line 616: For `truststore` and `internalClient.ResetClusterMember` below we already ignore error if `force` is set. I add this here to support the case where the database is cleaned up correctly but the truststore etc not yet. (Related to #487)

roosterfish: Makes sense to me thanks. Please move it into its own commit for clarity.

roosterfish (changes requested):

internal/rest/resources/cluster.go
bschimke95: Done

internal/rest/resources/cluster.go
bschimke95: Makes sense, I've put the client funcs update into a separate commit and applied your suggestion.

internal/rest/client/cluster.go
```diff
@@ -10,9 +10,18 @@ import (
 	"github.com/canonical/microcluster/v3/rest/types"
 )
 
+// withTimeoutIfUnset returns a context with a 30s timeout only if the parent context has no deadline set.
+func withTimeoutIfUnset(ctx context.Context) (context.Context, context.CancelFunc) {
+	if _, ok := ctx.Deadline(); ok {
```
roosterfish: Can you please write the assignment in its own line? We usually do this for readability across all the code bases. Unfortunately the check isn't yet in place for Microcluster repo:

```go
_, ok := ctx.Deadline()
if ok {

}
```

roosterfish (changes requested):
Looks good, please have a look at my last comment.

internal/rest/client/cluster.go
roosterfish: Ah, seems like it is.

roosterfish: https://github.com/canonical/microcluster/actions/runs/18498481592/job/52708657253?pr=512

bschimke95: meh, I forget this every time I want to commit to this repository :(

roosterfish (approved):

bschimke95:
Thanks for merging this @roosterfish 
We would like to backport this. Do you see any issues/breaking changes with this? Otherwise, I'd go ahead and create the backport.

roosterfish:
> Thanks for merging this @roosterfish We would like to backport this. Do you see any issues/breaking changes with this? Otherwise, I'd go ahead and create the backport.

Yes please!

Issue #503, opened 2025-10-10
Remove LXD dependency from Microcluster


mseralessandri:
# Remove LXD dependency from microcluster

**Goal**  
Gradually remove all dependencies on github.com/canonical/lxd/lxd/*, keeping changes independent.  
Each pull request focuses on a single dependency removal. Import from LXD only what is needed in Microcluster.

## Planned Work
 - [x]   **Remove dependency on `lxd/request`**  
```
./internal/daemon/daemon.go`
```

 - [x]   **Remove dependency on` lxd/response`**  
 ```
./example/api/extended.go  
./internal/daemon/daemon.go  
./internal/rest/resources/api_1.0.go  
./internal/rest/resources/certificates.go  
./internal/rest/resources/cluster.go  
./internal/rest/resources/control.go  
./internal/rest/resources/daemon.go  
./internal/rest/resources/database.go  
./internal/rest/resources/heartbeat.go  
./internal/rest/resources/hooks.go  
./internal/rest/resources/ready.go  
./internal/rest/resources/shutdown.go  
./internal/rest/resources/sql.go  
 ./internal/rest/resources/tokens.go  
 ./internal/rest/resources/truststore.go  
 ./internal/rest/rest.go  
 ./rest/access/handlers.go  
 ./rest/rest.go
```

 - [x]   **Remove dependency on` lxd/cluster/request`**  
```
./client/client.go  
./internal/rest/client/client.go
```

 - [x]   **Remove dependency on `lxd/endpoints/listeners`**  
``` 
./internal/endpoints/network.go`
```

 - [x]   **Remove dependency on `lxd/util`**  
```
./internal/daemon/daemon.go  
./internal/daemon/daemon_test.go  
./internal/endpoints/network.go  
./internal/rest/resources/control.go  
./rest/access/handlers.go
```

 - [x]   **Remove dependency on `lxd/db/query`**  
```
./cluster/cluster_members.mapper.go  
./cluster/token_records.mapper.go  
./example/database/extended.mapper.go  
./internal/db/db.go  
./internal/db/db_test.go  
./internal/db/update/cluster.go  
./internal/db/update/schema.go  
./internal/db/update/update_test.go  
./internal/rest/resources/sql.go
```

 - [x]   **Remove dependency on `lxd/db/schema`**  
```
./example/database/extended_schema.go  
./internal/daemon/daemon.go  
./internal/db/db_test.go  
./internal/db/dqlite.go  
./internal/db/update/schema.go  
[...]

PR #513, opened 2025-10-14
internal/endpoints: removed dependency from LXD FancyTLSListener

mseralessandri:
Refers to https://github.com/canonical/microcluster/issues/503
LXD-3654

### DIFF
--- a/internal/endpoints/mutable_tls_listener.go
+++ b/internal/endpoints/mutable_tls_listener.go
@@ -1,0 +1,61 @@
+package endpoints
+
+import (
+	"crypto/tls"
+	"net"
+	"sync"
+
+	"github.com/canonical/lxd/lxd/util"
+	"github.com/canonical/lxd/shared"
+)
+
+// mutableTLSListener is a variation of the standard tls.Listener that supports
+// atomically swapping the underlying TLS configuration.
+// Requests served before the swap will continue using the old configuration.
+// This implementation matches LXD's FancyTLSListener but excludes:
+//   - trustedProxy []net.IP field (proxy protocol support not needed)
+//   - TrustedProxy() method for setting proxy IPs
+//   - isProxy() function and proxy protocol wrapping logic
+//   - proxyproto.NewConn() calls in Accept()
+type mutableTLSListener struct {
+	net.Listener
+	mu     sync.RWMutex
+	config *tls.Config
+}
+
+// newMutableTLSListener creates a new mutableTLSListener.
+func newMutableTLSListener(inner net.Listener, cert *shared.CertInfo) *mutableTLSListener {
+	listener := &mutableTLSListener{
+		Listener: inner,
+	}
+
+	listener.Config(cert)
+	return listener
+}
+
+// Accept waits for and returns the next incoming TLS connection then use the
+// current TLS configuration to handle it.
+// EXCLUDED from LXD: proxy protocol check and wrapping:
+//   - if isProxy(c.RemoteAddr().String(), l.trustedProxy) { c = proxyproto.NewConn(c, 0) }
+func (l *mutableTLSListener) Accept() (net.Conn, error) {
+	c, err := l.Listener.Accept()
+	if err != nil {
+		return nil, err
+	}
+
+	l.mu.RLock()
+	defer l.mu.RUnlock()
+	config := l.config
+
+	return tls.Server(c, config), nil
+}
+
+// Config safely swaps the underlying TLS configuration.
+func (l *mutableTLSListener) Config(cert *shared.CertInfo) {
+	config := util.ServerTLSConfig(cert)
+
+	l.mu.Lock()
+	defer l.mu.Unlock()
+
+	l.config = config
+}
--- a/internal/endpoints/mutable_tls_listener_test.go
+++ b/internal/endpoints/mutable_tls_listener_test.go
@@ -1,0 +1,25 @@
+package endpoints
+
+import (
+	"testing"
+
+	"github.com/canonical/lxd/shared"
+	"github.com/stretchr/testify/require"
+)
+
+// Test_mutableTLSListener tests immutability and concurrent access.
+func Test_mutableTLSListener(t *testing.T) {
+	listener := &mutableTLSListener{}
+	cert1 := shared.TestingKeyPair()
+	cert2 := shared.TestingAltKeyPair()
+
+	// Test immutability: different certs create different config objects
+	listener.Config(cert1)
+	config1 := listener.config
+	require.NotNil(t, config1)
+
+	listener.Config(cert2)
+	config2 := listener.config
+	require.NotNil(t, config2)
+	require.NotSame(t, config1, config2, "Different certs should create different config objects")
+}
--- a/internal/endpoints/network.go
+++ b/internal/endpoints/network.go
@@ -10,7 +10,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/canonical/lxd/lxd/endpoints/listeners"
 	"github.com/canonical/lxd/lxd/util"
 	"github.com/canonical/lxd/shared"
 	"github.com/canonical/lxd/shared/api"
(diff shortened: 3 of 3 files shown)


internal/endpoints/network.go
```diff
@@ -74,21 +74,31 @@ func (n *Network) Listen() error {
 		return fmt.Errorf("Failed to listen on https socket: %w", err)
 	}
 
-	n.listener = listeners.NewFancyTLSListener(listener, n.cert)
+	// Create TLS configuration with dynamic certificate loading
+	config := &tls.Config{
+		GetCertificate: func(*tls.ClientHelloInfo) (*tls.Certificate, error) {
```
roosterfish: Interesting approach, if this flies well we could also think about doing the same in LXD (I think it was originally copied from there).

However it currently causes a failure in the pipeline.

internal/endpoints/network.go
```diff
@@ -74,21 +74,31 @@ func (n *Network) Listen() error {
...
+			// Read current certificate from network struct
+			n.certMu.RLock()
+			defer n.certMu.RUnlock()
+			keypair := n.cert.KeyPair()
+			return &keypair, nil
+		},
+		MinVersion: tls.VersionTLS12,
```
roosterfish: What is the background behind this? Can you please add a comment. What is the default value?

roosterfish (changes requested):

internal/endpoints/network.go
mseralessandri: Switched from Go stdlib-only approach to the `FancyTLSListener` re-implementation in Microcluster. It turns out we still need LXD's `util.ServerTLSConfig` for TLS features that stdlib lacks.

internal/endpoints/fancytls.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"crypto/tls"
+	"net"
+	"sync"
+
+	"github.com/canonical/lxd/lxd/util"
+	"github.com/canonical/lxd/shared"
+)
+
+// fancyTLSListener is a variation of the standard tls.Listener that supports
```
roosterfish: Based on the discussion from yesterday, what about `mutableTLSListener`?

internal/endpoints/fancytls.go
```diff
@@ -0,0 +1,64 @@
...
+// This implementation matches LXD's FancyTLSListener but excludes:
+//   - trustedProxy []net.IP field (proxy protocol support not needed)
+//   - TrustedProxy() method for setting proxy IPs
+//   - isProxy() function and proxy protocol wrapping logic
+//   - proxyproto.NewConn() calls in Accept()
+type fancyTLSListener struct {
+	net.Listener
+	mu     sync.RWMutex
+	config *tls.Config
+	// EXCLUDED from LXD: trustedProxy []net.IP (proxy protocol not needed by MicroCluster)
```
roosterfish: Isn't this already included in the list above?

roosterfish (changes requested):

internal/endpoints/mutable_tls_listener.go
```diff
@@ -0,0 +1,60 @@
...
+//   - if isProxy(c.RemoteAddr().String(), l.trustedProxy) { c = proxyproto.NewConn(c, 0) }
+func (l *mutableTLSListener) Accept() (net.Conn, error) {
+	c, err := l.Listener.Accept()
+	if err != nil {
+		return nil, err
+	}
+
+	l.mu.RLock()
+	defer l.mu.RUnlock() // Use defer pattern like LXD for thread safety
+	config := l.config
```
roosterfish: Mhh, this section made me thinking. Isn't this causing a race? Because all we do is assign the pointer val `l.config` to `config` so even after releasing the lock we are still accessing the same underlying struct. See this example in which both pointers `%p %p` are identical:

```go
type mutableTLSListener struct {
	net.Listener
	mu     sync.RWMutex
	config *tls.Config
}

func main() {
	t := time.Unix(0, 0)

	l := mutableTLSListener{
		config: &tls.Config{
			Time: func() time.Time {
				return t
			},
		},
	}

	l.mu.RLock()
	copyConfig := l.config
	l.mu.RUnlock()

	fmt.Printf("%p %p\n", l.config, copyConfig)
	fmt.Println(l.config.Time())
	fmt.Println(copyConfig.Time())

	t = time.Unix(1, 0)

	fmt.Println(l.config.Time())
	fmt.Println(copyConfig.Time())
}
```

roosterfish (changes requested):

internal/endpoints/mutable_tls_listener.go
roosterfish: Looks this also affects LXD.

mseralessandri: There is race condition in your example, I tried out the code.

In our case though, I think we're safe because `util.ServerTLSConfig()` always creates new config objects.

Every time we update TLS, we create a completely new config object and swap the pointer. Old connections keep using the old object, new connections get the new one. I'll add a test to be on the safe side.

roosterfish: But won't we swap out the config of existing accepted connections?

Once we update `l.config` inside `Config` we also modify the TLS config of already accepted connections because we simply copied the pointer before.

internal/endpoints/mutable_tls_listener_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+	// Test immutability: different certs create different config objects
+	listener.Config(cert1)
+	config1 := listener.config
+	require.NotNil(t, config1)
+
+	listener.Config(cert2)
+	config2 := listener.config
+	require.NotNil(t, config2)
+	require.NotEqual(t, config1, config2, "Different certs should create different config objects")
```
roosterfish: The docs on https://pkg.go.dev/github.com/stretchr/testify/require#NotEqual say this:

> Pointer variable equality is determined based on the equality of the referenced values (as opposed to the memory addresses). 

So it doesn't check for equality of the pointer but verifies the struct's values?
Maybe rather use https://pkg.go.dev/github.com/stretchr/testify/require#NotSame:

> NotSame asserts that two pointers do not reference the same object.

roosterfish (changes requested):

internal/endpoints/mutable_tls_listener_test.go
roosterfish: When using `NotSame` I assume the test will fail as it points to the same underlying object `listener.config`.

mseralessandri: Thanks for the thorough review! The test isn't failing though, let me check what the correct behavior should be.

internal/endpoints/mutable_tls_listener.go
roosterfish: I think the essential piece missing here is something like this https://pkg.go.dev/crypto/tls#Config.Clone. 
Because if we clone the config, we have our own copy which we can safely pass down to `tls.Server` so we can modify it again freely using our own `.Config`.

roosterfish: Ok nvm, I was confusing things. What you are doing is correct. We don't copy the pointer but the pointer value. So once `config := l.config` is assigned, we can freely change the value of `l.config`. And your test is checking this successfully. 

I think either `NotSame` or `NotEqual` should be fine.

internal/endpoints/mutable_tls_listener_test.go
roosterfish: Should be ok as is, see https://github.com/canonical/microcluster/pull/513#discussion_r2445147168.

internal/endpoints/mutable_tls_listener_test.go
```diff
@@ -0,0 +1,39 @@
...
+	config1 := listener.config
+	require.NotNil(t, config1)
+
+	listener.Config(cert2)
+	config2 := listener.config
+	require.NotNil(t, config2)
+	require.NotSame(t, config1, config2, "Different certs should create different config objects")
+
+	// Test concurrent access doesn't cause race conditions
+	var wg sync.WaitGroup
```
roosterfish: Thanks for the test, but I am not too sure about this section. What specifically are you trying to test here?

roosterfish (changes requested):

internal/endpoints/mutable_tls_listener_test.go
roosterfish: We can never ensure the order in which the routines get executed, if they even run in parallel as we start them one after another and they only run a short amount of time.

roosterfish (approved):
Thanks!

PR #515, opened 2025-10-17
fix: add consistency checks across core_cluster_members, truststore, and dqlite

louiseschmidtgen:
## Problem
Microcluster can enter inconsistent states where core_cluster_members (database), truststore, and dqlite cluster configuration become out of sync during partial failures. This leads to failed operations and difficult recovery scenarios.

## Solution
Implements membership consistency validation before critical operations:

Validates before operations: Checks all three sources match before joins, removals, and token generation
Clear error messages: Shows differences between sources when inconsistencies detected making it possible for admins to recover their cluster before worse things happen.

## Changes Made

- [state.go](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) - Core consistency checking logic with CheckMembershipConsistency()
- [cluster.go](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) - Added checks before join/remove operations
- [tokens.go](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) - Added checks before token generation
- [main.sh](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) - Integration test simulating inconsistent state and verifying blocked operations, parallel join test showing that join operations started concurrently do not fail

## Testing

```
./example/test/main.sh membership  # Test membership consistency 
./example/test/main.sh parallel-join  # Test parallel join operations
```

### DIFF
--- a/example/test/main.sh
+++ b/example/test/main.sh
@@ -68,7 +68,10 @@ shutdown_systems() {
 
   # The cluster doesn't always shut down right away; we've given it a chance
   for job_pid in $(jobs -p); do
-    kill -9 "${job_pid}"
+    # Check if process still exists before trying to kill it
+    if kill -0 "${job_pid}" 2>/dev/null; then
+      kill -9 "${job_pid}" 2>/dev/null || true
+    fi
   done
 }
 
--- a/example/test/main.sh
+++ b/example/test/main.sh
@@ -403,6 +403,140 @@ test_join_token_before_cluster_formed() {
   shutdown_systems
 }
 
+test_membership_consistency() {
+  echo "Testing membership consistency checks"
+  
+  new_systems 4 --heartbeat 2s
+  
+  # Bootstrap first member (daemon already running from new_systems)
+  microctl --state-dir "${test_dir}/c1" init "c1" 127.0.0.1:9001 --bootstrap
+  
+  # Join second member (daemon already running)
+  token_c2=$(microctl --state-dir "${test_dir}/c1" tokens add "c2")
+  microctl --state-dir "${test_dir}/c2" init "c2" 127.0.0.1:9002 --token "${token_c2}"
+
+  # Start third member and join cluster
+  token_c3=$(microctl --state-dir "${test_dir}/c1" tokens add "c3")
+  microctl --state-dir "${test_dir}/c3" init "c3" 127.0.0.1:9003 --token "${token_c3}"
+  
+  # Fetch join token for c4
+  token_c4=$(microctl --state-dir "${test_dir}/c1" tokens add "c4")
+
+  # Wait for cluster to stabilize
+  echo "  -> Waiting for cluster members to exit PENDING state"
+  
+  # Wait for all members to be promoted from PENDING
+  retry_count=0
+  max_retries=10
+  while [[ -n "$(microctl --state-dir "${test_dir}/c1" cluster list -f yaml | yq '.[] | select(.role == "PENDING")')" ]] && [[ ${retry_count} -lt ${max_retries} ]]; do
+    echo "  -> Still waiting for members to exit PENDING state..."
+    sleep 2
+    retry_count=$((retry_count + 1))
+  done
+  
+  echo "  -> Cluster established successfully"
+  
+  # Verify cluster is healthy
+  cluster_size=$(microctl --state-dir "${test_dir}/c1" cluster list -f yaml | yq '. | length')
+  if [ "${cluster_size}" != "3" ]; then
+    echo "ERROR: Expected cluster size 3, got ${cluster_size}"
+    exit 1
+  fi
+  
+  # Simulate inconsistent state by directly manipulating the database
+  # while keeping dqlite/truststore intact
+  echo "  -> Simulating inconsistent membership state"
+
+  # Remove c2's membership from core_cluster_members (simulating partial remove failure)
+  microctl --state-dir "${test_dir}/c1" sql "DELETE FROM core_cluster_members WHERE name = 'c2'"
+  
+  echo "  -> Created inconsistent state (c2 removed from database but still in truststore until heartbeat timeout)"
+
+  # Test member removal with inconsistent state
+  echo "  -> Testing member removal with inconsistent state"
+  if microctl --state-dir "${test_dir}/c1" cluster remove c2 2>/tmp/remove_error; then
+    echo "ERROR: Member removal should have failed due to inconsistent state"
+    cat /tmp/remove_error
+    exit 1
+  else
+    echo "  -> Member removal correctly blocked due to membership inconsistency"
+    cat /tmp/remove_error
+  fi
+  
+  # Try to join a new member - this should fail due to inconsistency
+  echo "  -> Testing join of new member c4 with inconsistent state"
+  if microctl --state-dir "${test_dir}/c4" init "c4" 127.0.0.1:9004 --token "${token_c4}" 2>/tmp/join_error; then
+    echo "ERROR: Member c4 should not have been able to join due to inconsistent state"
+    cat /tmp/join_error
+    exit 1
+  else
+    echo "  -> Membership inconsistency correctly detected, c4 join blocked"
+    cat /tmp/join_error
+  fi
+  
+  # Attempt to generate token should fail
+  echo "  -> Testing token generation with inconsistent state"
+  if microctl --state-dir "${test_dir}/c1" tokens add c5 2>/tmp/token_error; then
+    echo "ERROR: Token generation should have failed due to inconsistent state"
+    cat /tmp/token_error
+    exit 1
+  else
+    echo "  -> Token generation correctly blocked due to membership inconsistency"
+    cat /tmp/token_error
+  fi
+
+  echo "  -> Membership consistency checks working as expected"
+  
+  shutdown_systems
+}
+
+test_parallel_joins() {
+  echo "Testing parallel joins"
+
+  new_systems 4 --heartbeat 2s
+
+  # Bootstrap first member
+  microctl --state-dir "${test_dir}/c1" init "c1" 127.0.0.1:9001 --bootstrap
+
+  # Prepare tokens for remaining members
+  token_c2=$(microctl --state-dir "${test_dir}/c1" tokens add "c2")
+  token_c3=$(microctl --state-dir "${test_dir}/c1" tokens add "c3")
+  token_c4=$(microctl --state-dir "${test_dir}/c1" tokens add "c4")
+
+  # Kick off joins in parallel and collect PIDs
+  microctl --state-dir "${test_dir}/c2" init "c2" 127.0.0.1:9002 --token "${token_c2}" &
+  pids=($!)
+  microctl --state-dir "${test_dir}/c3" init "c3" 127.0.0.1:9003 --token "${token_c3}" &
+  pids+=($!)
+  microctl --state-dir "${test_dir}/c4" init "c4" 127.0.0.1:9004 --token "${token_c4}" &
+  pids+=($!)
+
+  for pid in "${pids[@]}"; do
+    if ! wait "${pid}"; then
+      echo "ERROR: parallel join failed (pid ${pid})"
+      return 1
+    fi
+  done
+
+  # Wait for cluster to stabilize
+  retry_count=0
+  max_retries=10
+  while [[ -n "$(microctl --state-dir "${test_dir}/c1" cluster list -f yaml | yq '.[] | select(.role == "PENDING")')" ]] && [[ ${retry_count} -lt ${max_retries} ]]; do
+    sleep 2
+    retry_count=$((retry_count + 1))
+  done
+
+  cluster_size=$(microctl --state-dir "${test_dir}/c1" cluster list -f yaml | yq '. | length')
+  if [ "${cluster_size}" != "4" ]; then
+    echo "ERROR: Expected cluster size 4 after parallel joins, got ${cluster_size}"
+    return 1
+  fi
+
+  echo "SUCCESS: parallel joins completed without failures"
+
+  shutdown_systems
+}
+
 test_extended_endpoints() {
   new_systems 4 --heartbeat 2s
 
@@ -442,6 +576,8 @@ if [ "${1:-"all"}" = "all" ] || [ "${1}" = "" ]; then
   test_join_token_after_cluster_formed
   test_join_token_before_cluster_formed
   test_extended_endpoints
+  test_membership_consistency
+  test_parallel_joins
 elif [ "${1}" = "recover" ]; then
   test_recover
 elif [ "${1}" = "tokens" ]; then
@@ -454,6 +590,10 @@ elif [ "${1}" = "join-before" ]; then
   test_join_token_before_cluster_formed
 elif [ "${1}" = "extended" ]; then
   test_extended_endpoints
+elif [ "${1}" = "membership" ]; then
+  test_membership_consistency
+elif [ "${1}" = "parallel-join" ]; then
+  test_parallel_joins
 else
   echo "Unknown test ${1}"
 fi
--- a/internal/rest/resources/cluster.go
+++ b/internal/rest/resources/cluster.go
@@ -100,6 +100,18 @@ func clusterPost(s state.State, r *http.Request) response.Response {
 		return response.SmartError(fmt.Errorf("Remote with address %q exists", req.Address.String()))
 	}
 
+	// Check cluster membership consistency before allowing joins
+	// This ensures core_cluster_members, truststore, and dqlite are all in sync
+	intState, err := internalState.ToInternal(s)
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	err = intState.CheckMembershipConsistency(ctx)
+	if err != nil {
+		return response.SmartError(err)
+	}
+
 	// Forward request to leader.
 	if leaderInfo.Address != s.Address().URL.Host {
 		client, err := s.Connect().Leader(false)
@@ -113,11 +125,6 @@ func clusterPost(s state.State, r *http.Request) response.Response {
 		}
 
 		return response.SyncResponse(true, tokenResponse)
-	}
-
-	intState, err := internalState.ToInternal(s)
-	if err != nil {
-		return response.SmartError(err)
 	}
 
 	// Check if the joining node's extensions are compatible with the leader's.
@@ -425,6 +432,20 @@ func clusterMemberDelete(s state.State, r *http.Request) response.Response {
 
 	ctx, cancel := context.WithTimeout(r.Context(), time.Second*60)
 	defer cancel()
+
+	// Check cluster membership consistency before allowing removals (unless forced)
+	// This ensures core_cluster_members, truststore, and dqlite are all in sync
+	if !force {
+		intState, err := internalState.ToInternal(s)
+		if err != nil {
+			return response.SmartError(err)
+		}
+
+		err = intState.CheckMembershipConsistency(ctx)
+		if err != nil {
+			return response.SmartError(err)
+		}
+	}
 
 	leader, err := s.Database().Leader(ctx)
 	if err != nil {
--- a/internal/rest/resources/cluster.go
+++ b/internal/rest/resources/cluster.go
@@ -76,8 +76,7 @@ func clusterPost(s state.State, r *http.Request) response.Response {
 		return response.BadRequest(err)
 	}
 
-	ctx, cancel := context.WithTimeout(r.Context(), time.Second*30)
-	defer cancel()
+	ctx := r.Context()
 
 	leaderClient, err := s.Database().Leader(ctx)
 	if err != nil {
@@ -119,7 +118,7 @@ func clusterPost(s state.State, r *http.Request) response.Response {
 			return response.SmartError(err)
 		}
 
-		tokenResponse, err := internalClient.AddClusterMember(r.Context(), client, req)
+		tokenResponse, err := internalClient.AddClusterMember(ctx, client, req)
 		if err != nil {
 			return response.SmartError(err)
 		}
@@ -430,8 +429,7 @@ func clusterMemberDelete(s state.State, r *http.Request) response.Response {
 		return response.SmartError(fmt.Errorf("No remote exists with the given name %q", name))
 	}
 
-	ctx, cancel := context.WithTimeout(r.Context(), time.Second*60)
-	defer cancel()
+	ctx := r.Context()
 
 	// Check cluster membership consistency before allowing removals (unless forced)
 	// This ensures core_cluster_members, truststore, and dqlite are all in sync
@@ -485,7 +483,7 @@ func clusterMemberDelete(s state.State, r *http.Request) response.Response {
 			return response.SmartError(err)
 		}
 
-		err = internalClient.DeleteClusterMember(r.Context(), client, name, force)
+		err = internalClient.DeleteClusterMember(ctx, client, name, force)
 		if err != nil {
 			return response.SmartError(err)
 		}
@@ -507,7 +505,7 @@ func clusterMemberDelete(s state.State, r *http.Request) response.Response {
 		})
 	}
 
-	info, err := leader.Cluster(r.Context())
+	info, err := leader.Cluster(ctx)
 	if err != nil {
 		return response.SmartError(err)
 	}
@@ -526,7 +524,7 @@ func clusterMemberDelete(s state.State, r *http.Request) response.Response {
 	}
 
 	var clusterMembers []cluster.CoreClusterMember
-	err = s.Database().Transaction(r.Context(), func(ctx context.Context, tx *sql.Tx) error {
+	err = s.Database().Transaction(ctx, func(ctx context.Context, tx *sql.Tx) error {
 		var err error
 		clusterMembers, err = cluster.GetCoreClusterMembers(ctx, tx)
 
@@ -564,7 +562,7 @@ func clusterMemberDelete(s state.State, r *http.Request) response.Response {
 	}
 
 	// Refresh members information since we may have changed roles.
-	info, err = leader.Cluster(r.Context())
+	info, err = leader.Cluster(ctx)
 	if err != nil {
 		return response.SmartError(err)
 	}
@@ -608,7 +606,7 @@ func clusterMemberDelete(s state.State, r *http.Request) response.Response {
 			clusterDisableMu.Unlock()
 		}()
 
-		err = internalClient.DeleteClusterMember(r.Context(), client, name, force)
+		err = internalClient.DeleteClusterMember(ctx, client, name, force)
 		if err != nil {
 			return response.SmartError(err)
 		}
@@ -648,7 +646,7 @@ func clusterMemberDelete(s state.State, r *http.Request) response.Response {
 	}
 
 	// Remove the cluster member from the database.
-	err = s.Database().Transaction(r.Context(), func(ctx context.Context, tx *sql.Tx) error {
+	err = s.Database().Transaction(ctx, func(ctx context.Context, tx *sql.Tx) error {
 		return cluster.DeleteCoreClusterMember(ctx, tx, remote.Address.String())
 	})
 	if err != nil && !force {
@@ -657,7 +655,7 @@ func clusterMemberDelete(s state.State, r *http.Request) response.Response {
 
 	// Remove the node from dqlite, if it has a record there.
 	if index >= 0 {
-		err = leader.Remove(r.Context(), info[index].ID)
+		err = leader.Remove(ctx, info[index].ID)
 		if err != nil {
 			return response.SmartError(err)
 		}
@@ -683,7 +681,7 @@ func clusterMemberDelete(s state.State, r *http.Request) response.Response {
 		return response.SmartError(err)
 	}
 
-	err = internalClient.ResetClusterMember(r.Context(), client, name, force)
+	err = internalClient.ResetClusterMember(ctx, client, name, force)
 	if err != nil && !force {
 		return response.SmartError(err)
 	}
@@ -694,7 +692,7 @@ func clusterMemberDelete(s state.State, r *http.Request) response.Response {
 	}
 
 	// Run the PostRemove hook locally.
-	hookCtx, hookCancel := context.WithCancel(r.Context())
+	hookCtx, hookCancel := context.WithCancel(ctx)
 	err = intState.Hooks.PostRemove(hookCtx, s, force)
 	hookCancel()
 	if err != nil {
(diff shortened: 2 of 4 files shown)


internal/rest/resources/tokens.go
```diff
@@ -50,6 +51,13 @@ func tokensPost(state state.State, r *http.Request) response.Response {
 		return response.SmartError(fmt.Errorf("Token name %q is not a valid FQDN: %w", req.Name, err))
 	}
 
+	// Check cluster membership consistency before allowing token creation
```
roosterfish, line 53: Why do we have to perform this check for token generation? Is it because the token might contain an address which is only present in one of the three sources but not in all of them?

internal/membership/membership.go
```diff
@@ -0,0 +1,99 @@
+package membership
```
roosterfish: I wonder if we have to put this into a new package? From what I see `CheckMembershipConsistency` is only ever called from inside an API handler. 

So should we rather move all of the `membership.*` funcs onto the `InternalState` struct inside `internal/state`? I think the name would also be fitting. You could leave the `CheckMembershipConsistency` exported and the others unexported as they are.

example/test/main.sh
```diff
@@ -409,13 +412,120 @@ test_join_token_before_cluster_formed() {
...
+  microctl --state-dir "${test_dir}/c3" init "c3" 127.0.0.1:9003 --token "${token_c3}"
+  
+  # Fetch join token for c4
+  token_c4=$(microctl --state-dir "${test_dir}/c1" tokens add "c4")
+
+  # Wait for cluster to stabilize
+  echo "  -> Waiting for cluster members to exit PENDING state"
+  
+  # Wait for all members to be promoted from PENDING
+  while [[ -n "$(microctl --state-dir "${test_dir}/c1" cluster list -f yaml | yq '.[] | select(.role == "PENDING")')" ]]; do
```
roosterfish: Can we please put a max retries check to this loop? Otherwise if this section fails the runner test job will never fail but instead run into a timeout after hours.

roosterfish: Same as the `max_retries` in `test_join_token_after_cluster_formed`.

example/test/main.sh
```diff
@@ -409,13 +412,120 @@ test_join_token_before_cluster_formed() {
...
+  fi
+  
+  # Simulate inconsistent state by directly manipulating the database
+  # while keeping dqlite/truststore intact
+  echo "  -> Simulating inconsistent membership state"
+  
+  # Stop c2 to manipulate its state safely
+  microctl --state-dir "${test_dir}/c2" shutdown || true
+  
+  # Remove c2 from c1's database directly (simulating partial failure)
```
roosterfish: `from cluster's database`? At this stage the DB is represented by c1 and c3.

roosterfish (changes requested):
Thanks raising awareness that we can ultimately end up in a state where those three data sources are out of sync. Also the corresponding test looks great.

I wonder if you could please share a reproducer how we might end up in such an inconsistent state in a production cluster?
From the test I understand that you can easily reproduce this when manually deleting DB entries (which we cannot protect against but it's something you should never really do). The same applies for manually editing the trust store.

Instead of checking for an already broken state when trying to add/delete members, what about performing the check elsewhere so we don't end up in this state overall? 
If it's easy to end up in such a state in a prod deployment, I wonder if Microcluster doesn't properly clean up certain actions in case of failure?

internal/rest/resources/tokens.go
louiseschmidtgen: The thought here is that we tell the user "hey something is wrong with your cluster" before they start the membership operation in the first place. 
Of course, if you prefer we can tell the user only at the later stage when they have the key in hand to join their new node to the cluster.
In other words, the reasoning is to detect the issue earlier.

example/test/main.sh
louiseschmidtgen: Good point 👍🏼

example/test/main.sh
louiseschmidtgen: Updated the comments

louiseschmidtgen:
> share a reproducer

As discussed at the sprint, the reverter logic only works as long as the node does not crash while the revert is ongoing.
The revert does not get persisted and is picked up when the node restart, meaning that we can remain in such a state where the membership's are out of sync.

roosterfish:
> As discussed at the sprint, the reverter logic only works as long as the node does not crash while the revert is ongoing. The revert does not get persisted and is picked up when the node restart, meaning that we can remain in such a state where the membership's are out of sync.

Hi @louiseschmidtgen, I appreciate the discussions at the sprint to better understand where you are heading with this.
As you mentioned the reverter pattern can be a single point of failure in case the member which executes the revert dies somewhere in the middle. 

With the changes proposed in this PR, it can still happen but there would now be an error telling the user that DB/dqlite/truststore are out of sync when performing actions like creating a token, adding and removing a cluster member.

In the presumably rare case a member running a revert dies, it should rather be the admin that figures out that the current action he is executing (e.g. adding a member) ran into a failure. When joining a member `m2` via existing member `m1` and `m1` dies during the process (interrupting the revert), the user will see an error as the connection to `m1` gets interrupted. The user then has to perform further checks/validations and ultimately retry the operation.
Therefore we already support a `--force` option during removal to be able to clean up and recover in case things are left partially. So if we see further issues like https://github.com/canonical/microcluster/pull/512, those have to be addressed to ensure a robust cleanup.

But as mentioned at the sprint, we don't want to put in checks at certain places to check if something is already broken. Rather fix it when its happening. And if something from the outside is unexpectedly killing a Microcluster member, this is certainly not under our control and we should not come up with patterns in Microcluster to resolve this.

louiseschmidtgen:
@roosterfish I understand that we do not want to add any unnecessary checks and rather iron out the faults in the repo. However, we are not going to fix the reverter logic any time soon.

From a UX perspective expecting `the user to perform further checks/validations and ultimately retry the operation` is not great. Why not log the error and not expect the user to figure out where to look for the inconsistencies. We can't expect the user to look in the microcluster's internal logic, neither (I admit) is bubbling up these internal details great. Without a solid reverter logic and the possibility of getting your cluster into an unrecoverable case, I would still advocate for this safeguard.

Let's get another opinion from @bschimke95.

roosterfish:
> @roosterfish I understand that we do not want to add any unnecessary checks and rather iron out the faults in the repo. However, we are not going to fix the reverter logic any time soon.

After https://github.com/canonical/microcluster/pull/512 and https://github.com/canonical/microcluster/pull/487 do we know what else is currently not working in case the reverter kicks in?

louiseschmidtgen:
> After #512 and #487 do we know what else is currently not working in case the reverter kicks in?

We will need to do more exploratory testing to find further cases, race conditions etc.

bschimke95:
Hey folks,

I generally agree with both of you.

* if the reverter operation fails because of a lost node I don't have a good way as a user to detect that, without knowing the internals of microcluster and check manually in a couple of places - this is arguably not a great UX

* As @louiseschmidtgen pointed out, it is also not great to bubble up internal state up to the user and tell them about microclusters internal inconsistency (especially in places where they would not expect it (e.g. when creating a token))

In both cases, we'd leak abstractions and users would need to know the internals of microcluster.

I'd propose two things:

a) If I understand the issue correctly then the main problem (that we know of) right now is that the reverter logic might not run into completion if the node dies mid-way. What about just making this reverter logic persistent? Presumably, this should not be too much work and would solve the issue at hand.

b) Introduce a `consistency` API (or integrate into an existing one) that exposes some information about the membership state that consumers then can use to integrate in their workflows.

louiseschmidtgen:
@bschimke95 AFAIK, @roosterfish and his team have experimented with an options to persist the change by storing it in Dqlite and making use of Raft's fault-tolerance. However, this was plagued with performance issues IIRC.
Finding a good solution here will likely be a non-trivial task.

bschimke95:
I'm not super deep in the Raft-game but I'd be surprised if there are any significant performance issues.
You only require a reverter for tasks like joining/removing etc. During those steps, you may store some small data in dqlite but you are also not (normally) concerned too much with performance during those steps. After the clustering step succeeded, you may cleanup those entries and there are no left-overs for the actual runtime.

That is just my understanding of the system/problem. I probably miss something.

louiseschmidtgen:
> I'm not super deep in the Raft-game but I'd be surprised if there are any significant performance issues. You only require a reverter for tasks like joining/removing etc. During those steps, you may store some small data in dqlite but you are also not (normally) concerned too much with performance during those steps. After the clustering step succeeded, you may cleanup those entries and there are no left-overs for the actual runtime.
> 
> That is just my understanding of the system/problem. I probably miss something.

I'm afraid it's never trivial with distributed systems, especially when you start to think about all the things that go wrong. Say we store the information in Dqlite `new node in core_cluster_members`, `new node in truststore`, ` new node in dqlite`. 

One example: assume my "new node" fails somewhere in the join process and crashes half way through the reverter and the new node's not coming back to clean up its state any time soon. Now the other nodes need to "deal with the mess" before doing any further membership operations. Dealing with the mess is going to require some more thought through logic and all of the sudden this is not a quick fix.

bschimke95:
I don't think that this is a quick fix but rather was surprised by the claim that this affects the performance of the system. I totally agree that thinks can get messy in distributed systems.

Anyway, @roosterfish and @louiseschmidtgen  how do you want to proceed with this PR? I think the direction where we want to go with this is a bit unclear right now and I don't see this PR to land anytime soon. 

My proposal would be to close this PR, regroup and discuss next steps in a better setting (e.g. spec) to agree on a aligned way forward.

roosterfish:
> Anyway, @roosterfish and @louiseschmidtgen how do you want to proceed with this PR?

Hey guys, sorry for the delay, this is not a priority for us at the moment as the goal of this PR is to circumvent possible behavior and not fixing an actual bug. The current answer of Microcluster is to `--force` remove a member in case you end up in a situation where this might be needed. 

I see the point that there can potentially be various different inconsistencies, for example a member entry only present in the truststore. As this is a hypothesis, I cannot present any reproducer to end up in this state, but in this case you can also delete the member using `--force`, Microcluster has logic for this already, see https://github.com/canonical/microcluster/blob/v3/internal/rest/resources/cluster.go#L504.
But it looks we should also have tests to perform this from various different cluster sizes as I see there are checks below which might hinder us from performing a cleanup.

@louiseschmidtgen in your previous message you mentioned user experience, so we should not simply notify the user but rather allow cleaning up in a "controlled way". What do you think about also extending the deletion endpoint to cope with the different scenarios? This would allow us to cover the entire workflow, from error to solution.

roosterfish:
@louiseschmidtgen before we continue with this, can you please have a look at https://github.com/canonical/microcluster/issues/544? There seems to be a regression I just found with the recent cluster join fixes.

louiseschmidtgen:
> > Anyway, @roosterfish and @louiseschmidtgen how do you want to proceed with this PR?
> 
> Hey guys, sorry for the delay, this is not a priority for us at the moment as the goal of this PR is to circumvent possible behavior and not fixing an actual bug. The current answer of Microcluster is to `--force` remove a member in case you end up in a situation where this might be needed.
> 
> I see the point that there can potentially be various different inconsistencies, for example a member entry only present in the truststore. As this is a hypothesis, I cannot present any reproducer to end up in this state, but in this case you can also delete the member using `--force`, Microcluster has logic for this already, see https://github.com/canonical/microcluster/blob/v3/internal/rest/resources/cluster.go#L504. But it looks we should also have tests to perform this from various different cluster sizes as I see there are checks below which might hinder us from performing a cleanup.
> 
> @louiseschmidtgen in your previous message you mentioned user experience, so we should not simply notify the user but rather allow cleaning up in a "controlled way". What do you think about also extending the deletion endpoint to cope with the different scenarios? This would allow us to cover the entire workflow, from error to solution.

@roosterfish I like your idea of doing automated clean-up. I will extend the PR to include this.

louiseschmidtgen:
> @louiseschmidtgen in your previous message you mentioned user experience, so we should not simply notify the user but rather allow cleaning up in a "controlled way". What do you think about also extending the deletion endpoint to cope with the different scenarios? This would allow us to cover the entire workflow, from error to solution.

Giving this some more thought:

The auto-resolution strategy should be:
- `dqlite` = source of truth 
  - it's the actual cluster, 
  - choosing `core_cluster_members` as source of truth could lead us to automatically remove a `dqlite` member and loose quorum 
- ~~Add missing entries to `core_cluster_members`/`truststore` if they exist in `dqlite`~~
  - **we can't add** the members back without the full member information (certificate, schema versions, etc.). We can only remove orphaned entries.
- Remove extra entries in `core_cluster_members`/ `truststore` if they don't exist in `dqlite` 
  - Potentially **problematic if it runs in parallel** with a node joining

@roosterfish / @bschimke95 What do you think? I don't see a good/quick way to do the clean-up without putting the cluster at risk.

internal/state/state.go
```diff
@@ -55,6 +60,9 @@ type State interface {
 
 	// ExtensionServers returns an immutable list of the daemon's additional listeners.
 	ExtensionServers() []string
+
+	// CheckMembershipConsistency verifies that all parts have consistent membership information.
+	CheckMembershipConsistency(ctx context.Context) error
```
roosterfish: As this is something internal, please move it to the [`InternalState`](https://github.com/canonical/microcluster/blob/v3/internal/state/state.go#L61) struct directly. With all the refactoring which is currently going on, we try to make the `State` interface as small as possible.

This would be the process on the calling side:

```go
intState, err := internalState.ToInternal(state)
if err != nil {
}

...

err = intState.CheckMembershipConsistency(ctx)
```

roosterfish: I see you already defined it on the `InternalState`. So just use this pattern to call it but please remove the func from the interface.

internal/state/state.go
```diff
@@ -229,3 +237,87 @@ func ToInternal(s State) (*InternalState, error) {
...
+	var dqliteAddresses []string
+	for _, node := range dqliteNodes {
+		dqliteAddresses = append(dqliteAddresses, node.Address)
+	}
+
+	sort.Strings(dqliteAddresses)
+
+	// Check if all three slices are equal
+	if !slices.Equal(coreClusterAddresses, trustAddresses) || !slices.Equal(coreClusterAddresses, dqliteAddresses) {
+		return fmt.Errorf("Microcluster node membership is inconsistent across core_cluster_members, truststore, and dqlite: CoreClusterMembersTable=%v, Truststore=%v, Dqlite=%v",
```
roosterfish: Can we rather do something like this:

```go
return fmt.Errorf("Microcluster node membership is inconsistent across core_cluster_members (%v), truststore (%v) and dqlite (%v)", coreClusterAddresses, trustAddresses, dqliteAddresses)
```

I find it odd that we first name it `core_cluster_members` and then `CoreClusterMembersTable` in the same sentence.

roosterfish: What do you think?

internal/rest/resources/tokens.go
roosterfish: Ok thanks.

example/test/main.sh
```diff
@@ -68,7 +68,10 @@ shutdown_systems() {
 
   # The cluster doesn't always shut down right away; we've given it a chance
   for job_pid in $(jobs -p); do
-    kill -9 "${job_pid}"
+    # Check if process still exists before trying to kill it
```
roosterfish, line 71: Please add this change into a new commit as it's not directly related with the new `test_membership_consistency` test suite.

example/test/main.sh
```diff
@@ -403,6 +406,115 @@ test_join_token_before_cluster_formed() {
   shutdown_systems
 }
 
+test_membership_consistency() {
+  echo "====> Testing membership consistency checks"
```
roosterfish: The other tests don't use `====>`. Can you please make this consistent?

example/test/main.sh
```diff
@@ -403,6 +406,115 @@ test_join_token_before_cluster_formed() {
...
+  
+  new_systems 4 --heartbeat 2s
+  
+  # Start first member
+  microd --state-dir "${test_dir}/c1" "${cluster_flags[@]}" &
```
roosterfish: The daemons are already started by `new_systems`.

roosterfish (changes requested):
Thanks, please have a look at my few additional comments.

As I wrote in the other thread, let's include this as the checks itself are not very expensive but can be helpful.
Once we get a reproducer to end up in such a state, we should address this directly by modifying either the member join or remove logic.

Another aspect I just thought about is whether or not we can recover from all of the different inconsistencies. In [this comment](https://github.com/canonical/microcluster/pull/515#issuecomment-3656639468) you mentioned that we might miss some data when restoring `core_cluster_members` so I suspect this is a manual operation anyway. 
Have you already tested this for `core_cluster_members`, truststore and dqlite?

internal/state/state.go
louiseschmidtgen: Sounds good to me 👍🏼

louiseschmidtgen:
> Thanks, please have a look at my few additional comments.
> 
> As I wrote in the other thread, let's include this as the checks itself are not very expensive but can be helpful. Once we get a reproducer to end up in such a state, we should address this directly by modifying either the member join or remove logic.
> 
> Another aspect I just thought about is whether or not we can recover from all of the different inconsistencies. In [this comment](https://github.com/canonical/microcluster/pull/515#issuecomment-3656639468) you mentioned that we might miss some data when restoring `core_cluster_members` so I suspect this is a manual operation anyway. Have you already tested this for `core_cluster_members`, truststore and dqlite?

+1 When we get a reproducer we can fix it in the future. In the meantime, this PR will help us / users find and define those issues. 
On the testing question yes, I tried implementing the automated recovery but based on my findings this won't be possible for some cases and is not safe to do in others.

internal/state/state.go
```diff
@@ -234,3 +234,86 @@ func ToInternal(s State) (*InternalState, error) {
...
+func (s *InternalState) CheckMembershipConsistency(ctx context.Context) error {
+	// Get data from all sources
+	coreClusterMembers, truststoreRemotes, dqliteNodes, err := s.getMembershipData(ctx)
+	if err != nil {
+		return fmt.Errorf("Failed to gather membership data for consistency check: %w", err)
+	}
+
+	// Check for inconsistencies across all three sources
+	err = s.checkMembershipConsistency(coreClusterMembers, truststoreRemotes, dqliteNodes)
+
```
roosterfish, line 319: Just a nit, we usually don't add newlines in between the check for `err`.

example/test/main.sh
```diff
@@ -403,6 +406,115 @@ test_join_token_before_cluster_formed() {
...
+  microctl --state-dir "${test_dir}/c2" shutdown || true
+
+  # Remove c2's membership from core_cluster_members (simulating partial remove failure)
+  microctl --state-dir "${test_dir}/c1" sql "DELETE FROM core_cluster_members WHERE name = 'c2'"
+  
+  echo "  -> Created inconsistent state (c2 removed from database but still in truststore)"
+  
+  # Try to join a new member - this should fail due to inconsistency
+  echo "  -> Testing join of new member c4 with inconsistent state"
+  microd --state-dir "${test_dir}/c4" "${cluster_flags[@]}" &
```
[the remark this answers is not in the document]
roosterfish: I guess this comment is still valid as you run `new_systems 4 --heartbeat 2s` in the first few lines of the test.

roosterfish (changes requested):
Almost ready, pls check the suggestion from Copilot and one small nit.

internal/state/state.go
```diff
@@ -229,3 +234,86 @@ func ToInternal(s State) (*InternalState, error) {
...
+	}
+
+	return coreClusterMembers, truststoreRemotes, dqliteNodes, nil
+}
+
+// checkMembershipConsistency checks consistency across all three membership sources using addresses.
+func (s *InternalState) checkMembershipConsistency(coreClusterMembers []cluster.CoreClusterMember, truststoreRemotes map[string]trust.Remote, dqliteNodes []dqliteClient.NodeInfo) error {
+	// Collect addresses from each source into sorted slices
+	var coreClusterAddresses []string
+	for _, member := range coreClusterMembers {
```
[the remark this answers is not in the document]
roosterfish: @louiseschmidtgen mhh is this something we have to explicitly check for? AFAIK the join process itself isn't governed by any higher level lock. So if we run two parallel joins, does this fail the consistency check if the first joined wasn't yet able to enter the member info in all three "stores" (table, trust and dqlite)?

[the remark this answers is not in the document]
roosterfish: Currently you can do concurrent joins and we should keep this behavior.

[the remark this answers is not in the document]
louiseschmidtgen: You can start the join operations concurrently but Dqlite will only allow you to add one node at a time to the configuration. In that sense, you've already got a serialized operation.

internal/state/state.go
louiseschmidtgen: My bad :)

internal/state/state.go
```diff
@@ -229,3 +234,86 @@ func ToInternal(s State) (*InternalState, error) {
...
+	}
+
+	// Check for inconsistencies across all three sources
+	err = s.checkMembershipConsistency(coreClusterMembers, truststoreRemotes, dqliteNodes)
+
+	if err != nil {
+		return err
+	}
+
+	return nil
```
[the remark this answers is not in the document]
louiseschmidtgen: I quite prefer having this return thank you.

internal/state/state.go
```diff
@@ -229,3 +234,86 @@ func ToInternal(s State) (*InternalState, error) {
...
+	var dqliteAddresses []string
+	for _, node := range dqliteNodes {
+		dqliteAddresses = append(dqliteAddresses, node.Address)
+	}
+
+	sort.Strings(dqliteAddresses)
+
+	// Check if all three slices are equal
+	if !slices.Equal(coreClusterAddresses, trustAddresses) || !slices.Equal(coreClusterAddresses, dqliteAddresses) {
+		return fmt.Errorf("Microcluster node membership is inconsistent across core_cluster_members (%v), truststore (%v) and dqlite (%v)", coreClusterAddresses, trustAddresses, dqliteAddresses)
```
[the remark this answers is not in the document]
louiseschmidtgen: I don't think there is a problem:
This renders quite alright:
Microcluster node membership is inconsistent across core_cluster_members ([127.0.0.1:9001 127.0.0.1:9003]), truststore ([127.0.0.1:9001 127.0.0.1:9002 127.0.0.1:9003]) and dqlite ([127.0.0.1:9001 127.0.0.1:9002 127.0.0.1:9003])"

internal/state/state.go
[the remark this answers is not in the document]
roosterfish: I see thanks. I was fearing this might lock concurrent join requests against the Microcluster API. But as long as it's serialized properly under the hood then we are good.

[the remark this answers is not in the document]
roosterfish: Just ran a few of those:

```
...
microctl --state-dir /tmp/m1 init m1 127.0.0.1:9001 --bootstrap
...
microctl --state-dir /tmp/m2 init m2 127.0.0.1:9002 --token "$token2" &
microctl --state-dir /tmp/m3 init m3 127.0.0.1:9003 --token "$token3" &
microctl --state-dir /tmp/m4 init m4 127.0.0.1:9004 --token "$token4" &
...
```

and got this in the log which confirms the suspision:

`time=2026-01-14T09:58:52.610+01:00 level=ERROR msg="Unable to complete cluster join request" address=127.0.0.1:9001 error="Microcluster node membership is inconsistent across core_cluster_members ([127.0.0.1:9001 127.0.0.1:9004]), truststore ([127.0.0.1:9001 127.0.0.1:9004]) and dqlite ([127.0.0.1:9001])"
Error: 1 join attempts were unsuccessful. Last error: Microcluster node membership is inconsistent across core_cluster_members ([127.0.0.1:9001 127.0.0.1:9004]), truststore ([127.0.0.1:9001 127.0.0.1:9004]) and dqlite ([127.0.0.1:9001])
`

internal/state/state.go
```diff
@@ -239,20 +239,33 @@ func ToInternal(s State) (*InternalState, error) {
 // all have consistent membership information. This should be called before any member
 // add/remove operations or join token generation to ensure the cluster is in a healthy state.
 func (s *InternalState) CheckMembershipConsistency(ctx context.Context) error {
-	// Get data from all sources
-	coreClusterMembers, truststoreRemotes, dqliteNodes, err := s.getMembershipData(ctx)
-	if err != nil {
-		return fmt.Errorf("Failed to gather membership data for consistency check: %w", err)
-	}
+	const maxRetries = 10
```
roosterfish: Can we just use the provided `ctx` as the upper limit? Then we don't have to define an arbitrary number of max retries but can just rely on the provided context.

As long as the consistency check succeeds within the context's deadline, we are fine and can continue on:

```go
for {
    err = s.checkMembershipConsistency
    ...
    
    select {
			case <-ctx.Done():
				return ctx.Err()
			case <-time.After(retryInterval):
				continue
    }
}
```

roosterfish: Maybe as an additional safety mechanism we can add a deadline if none is already set on the context?

The `clusterPost` handler currently already sets a deadline of 30s on the context which is used here so let's use 30s as an upper limit if no deadline is set on the context.

roosterfish (changes requested):

internal/state/state.go
roosterfish: The extra newline is still in there.

internal/state/state.go
louiseschmidtgen: Good idea thanks, I've implemented your suggestion.

internal/state/state.go
```diff
@@ -239,20 +239,27 @@ func ToInternal(s State) (*InternalState, error) {
 // all have consistent membership information. This should be called before any member
 // add/remove operations or join token generation to ensure the cluster is in a healthy state.
 func (s *InternalState) CheckMembershipConsistency(ctx context.Context) error {
-	// Get data from all sources
-	coreClusterMembers, truststoreRemotes, dqliteNodes, err := s.getMembershipData(ctx)
-	if err != nil {
-		return fmt.Errorf("Failed to gather membership data for consistency check: %w", err)
-	}
+	checkCtx, cancel := context.WithDeadline(ctx, time.Now().Add(30*time.Second))
```
roosterfish: 30 seconds should be the guardrail in case no deadline is configured to ensure we don't end up in a dead lock. But if the caller decides to wait for 60s we should still allow it, see [this pattern](https://github.com/canonical/microcluster/blob/v3/internal/rest/client/client.go#L197-L202) for reference.

roosterfish: Also please squash this commit into the other one in which you add the `CheckMembershipConsistency` func to not add and remove stuff in the same PR.

roosterfish (changes requested):

internal/state/state.go
louiseschmidtgen: Should be gone :)

internal/rest/resources/cluster.go
```diff
@@ -77,8 +77,7 @@ func clusterPost(s state.State, r *http.Request) response.Response {
 		return response.BadRequest(err)
 	}
 
-	ctx, cancel := context.WithTimeout(r.Context(), time.Second*30)
-	defer cancel()
+	ctx := r.Context()
```
roosterfish, line 79: I am afraid we cannot simply remove the deadline as we want the API handler `clusterPost` to return at some point if the operation wasn't successful. Otherwise it might run indefinitely. 

In the chat message you mentioned you see this error `Error: Delete "http://control.socket/core/1.0/cluster/c2": context deadline exceeded
` which ultimately is produced on the client side. 

My suggestion was to instead for returning the generic context deadline exceeded error you can return the actual error from the consistency check. 
However this doesn't apply here because they client can always set a custom context deadline the server isn't even aware of. And in this case the client exits before the context on the server runs into a deadline.
So the only way to overcome this would be to lower the deadline on the server to ensure it always returns before the client's context exceeds or increase the deadline on the client.
But this doesn't feel like the right approach here.

Let's just log the error on the server side. This way we don't loose the information and we can still backport this into v2 as we don't modify the fixed timeouts.

internal/rest/resources/cluster.go
```diff
@@ -424,8 +430,21 @@ func clusterMemberDelete(s state.State, r *http.Request) response.Response {
 		return response.SmartError(fmt.Errorf("No remote exists with the given name %q", name))
 	}
 
-	ctx, cancel := context.WithTimeout(r.Context(), time.Second*60)
-	defer cancel()
+	ctx := r.Context()
```
roosterfish, line 432: Same here as with the other API handler.

roosterfish (changes requested):

internal/rest/resources/cluster.go
louiseschmidtgen: I see your concern but that doesn't make sense if we use [Go contexts](https://pkg.go.dev/context) the way they are intended to work.
The parent context is responsible for cancelling the context. Where it is created it is also where it is supposed to be closed. That way nothing hangs indefinitely.

The way we handle it now swallows valuable information not just my error message. Setting these extra contexts has led to hard to debug bugs before. Why don't we try to follow Go conventions?

roosterfish: My concern is all about it being "backportable". 
I am happy to accept the removal of the context deadline inside the api handler in `v3` only, but then please add it to an additional commit so that it doesn't get backported together with the consistency checks.
In `v2` we should not change those fixed limits as it will fundamentally change how the join/remove endpoints behave.

Also we have many API endpoints where this pattern is used, so I would rather do this in a separate PR for all the endpoints altogether. Let's not make this PR bigger as it has to be?

roosterfish:
Please rebase with `main` because there is now a conflict after merging https://github.com/canonical/microcluster/pull/582.

internal/rest/resources/cluster.go
louiseschmidtgen: I split the context related changes into their separate commit 👍🏼

internal/rest/resources/cluster.go
```diff
@@ -100,6 +100,21 @@ func clusterPost(s state.State, r *http.Request) response.Response {
 		return response.SmartError(fmt.Errorf("Remote with address %q exists", req.Address.String()))
 	}
 
+	ctx, cancel := context.WithTimeout(r.Context(), time.Second*60)
```
roosterfish: I guess this is just a leftover from rebasing? The deadline gets set already a few lines before in 79.

roosterfish: Most probably as you remove it in the new commit that replaces the contexts.

roosterfish (changes requested):
Thanks! Only one potential leftover from rebasing.

internal/rest/resources/cluster.go
louiseschmidtgen: Ah, thanks for catching that! It's in the one commit and then removed in the ctx commit. I'll clean that up.

louiseschmidtgen: Done :)

roosterfish (approved):
LGTM!

PR #518, opened 2025-10-21
Remove the dependency from lxd/db/query and lxd/schema

mseralessandri:
Refers to #503


internal/db/query/retry.go
```diff
@@ -0,0 +1,105 @@
...
+// jitterDeviation adds random jitter to retry timing.
+// This implementation matches LXD's jitterDeviation function.
+func jitterDeviation(factor float64, duration time.Duration) time.Duration {
+	floor := int64(math.Floor(float64(duration) * (1 - factor)))
+	ceil := int64(math.Ceil(float64(duration) * (1 + factor)))
+	return time.Duration(rand.Int64N(ceil-floor) + floor)
+}
+
+// IsRetriableError returns true if the given error might be transient and the
+// interaction can be safely retried. This implementation matches LXD's IsRetriableError function.
```
roosterfish: Minor nit, but all other doc strings use a new line for the "This implementation ..." sentence.

internal/db/query/objects.go
```diff
@@ -0,0 +1,54 @@
...
+		}
+	}
+
+	return rows.Err()
+}
+
+// Dest is a function that is expected to return the objects to pass to the
+// 'dest' argument of sql.Rows.Scan(). It is invoked by SelectObjects once per
+// yielded row, and it will be passed the index of the row being scanned.
+type Dest func(scan func(dest ...any) error) error
```
roosterfish: Can we please specify types at the top of the file?

internal/db/query/slices.go
```diff
@@ -0,0 +1,83 @@
...
+	err = rows.Err()
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Function to scan a single row.
+type scanFunc func(*sql.Rows) error
```
roosterfish: Please move to the top of the file.

internal/db/query/dump.go
```diff
@@ -0,0 +1,171 @@
+// Package query dump functions that match LXD's dump functionality.
```
roosterfish, line 1: This and the other commit messages are missing the package name as prefix.

internal/db/query/dump.go
```diff
@@ -0,0 +1,171 @@
...
+	"database/sql"
+	"fmt"
+	"strconv"
+	"strings"
+	"time"
+)
+
+// Dump returns a SQL text dump of all rows across all tables, similar to
+// sqlite3's dump feature. This implementation matches LXD's Dump function.
+func Dump(ctx context.Context, tx *sql.Tx, schemaOnly bool) (string, error) {
```
roosterfish, line 16: Can you please also add the unit tests for the respective files under `db/query`? This helps keeping a good score in Tiobe as we are now having the code duplicated in the Microcluster repo.

internal/db/query/dump.go
```diff
@@ -0,0 +1,171 @@
...
+	"context"
+	"database/sql"
+	"fmt"
+	"strconv"
+	"strings"
+	"time"
+)
+
+// Dump returns a SQL text dump of all rows across all tables, similar to
+// sqlite3's dump feature. This implementation matches LXD's Dump function.
```
roosterfish: Newline for the "This implementation .." sentence.

internal/db/schema/schema.go
```diff
@@ -0,0 +1,20 @@
...
+	"database/sql"
+)
+
+// Update represents a database schema update function.
+// It takes a context and transaction and applies the schema changes.
+type Update func(ctx context.Context, tx *sql.Tx) error
+
+// Hook is a function that gets executed during schema updates.
+// The version parameter indicates which update version is being processed.
+// A version of -1 indicates a special hook execution.
```
roosterfish: What does special mean?

internal/db/query/slices_test.go
```diff
@@ -0,0 +1,84 @@
+package query_test
```
roosterfish, line 1: Only saw later that the tests are there. Maybe you can keep those commits together? I think only the `objects_test.go` is missing.

cluster/cluster_members.mapper.go
```diff
@@ -10,7 +10,7 @@ import (
 	"net/http"
 	"strings"
 
-	"github.com/canonical/lxd/lxd/db/query"
+	"github.com/canonical/microcluster/v3/internal/db/query"
```
roosterfish: Have you generated this file using `lxd-generate`? Do we also need to backport this tool as it might imply certain package import paths?

roosterfish (changes requested):
Thanks. Please have a look at my few comments. The biggest question I have is around `lxd-generate` and if we can still use this tool to generate the mapper files given that we now use our own DB funcs?

cluster/cluster_members.mapper.go
roosterfish: As discussed yesterday let's get rid of the "generated by" headers and don't mind about the tool as we anyway only have a few of those mapper files.

internal/db/schema/schema.go
mseralessandri: I removed the confusing reference to version -1 from the comment (in LXD -1 is passed to the `Hook` in the `File()` method to execute queries from a file, but Microcluster doesn't implement this functionality)

cluster/cluster_members.mapper.go
```diff
@@ -1,7 +1,5 @@
 package cluster
```
roosterfish, line 1: I think it's fine if we keep the files named as `.mapper.go` as it still identifies the files as being once generated.

roosterfish (approved):
LGTM!

PR #520, opened 2025-10-27
Replace LXD's logger with Go's standard logger

roosterfish:
Go's new `log/slog` logger which is part of the std. library has the concept of a handler (interface) which allows configuring log output and level as well as other options. Such a handler can now be passed into Microcluster for log fine tuning.

Furthermore internally Microcluster now uses `log/slog` for each log message by deriving the central logger instance from the daemon's context which is shared into every branch of the daemon as well as the request contexts of the API handlers.

As an effect of those changes we also drop the `--verbose` flag from the commands inside the `example` package.

I have committed the changes in phases:
1. Adding utils required for the new logging approach
2. Passing the actual logger into the daemon from Microcluster's `app` construct
3. Further passing it into the various internal (and some public) packages (most of the changes in this PR)
4. Using/materializing it in the `example` package and demonstrating the use also inside the hooks (by deriving the logger from the context)


roosterfish:
@markylaing added another `Logger` func on the app construct for the `example` package. This allows accessing the logger from outside without having to care about the context. Therefore we keep this idea of passing the logger around in the context also as something we use internally.

internal/log/logger.go
```diff
@@ -0,0 +1,22 @@
+package log
+
+import (
+	"context"
+	"errors"
+	"log/slog"
+)
+
+type CtxKey string
```
markylaing: Why export this type?

rest/access/handlers.go
```diff
@@ -83,7 +88,7 @@ func Authenticate(state state.State, r *http.Request, hostAddress string, truste
 			for _, cert := range r.TLS.PeerCertificates {
 				trusted, fingerprint := util.CheckMutualTLS(*cert, trustedCerts)
 				if trusted {
-					logger.Debugf("Trusting HTTP request to %q from %q with fingerprint %q", r.URL.String(), r.RemoteAddr, fingerprint)
+					logger.Debug(fmt.Sprintf("Trusting HTTP request to %q from %q with fingerprint %q", r.URL.String(), r.RemoteAddr, fingerprint))
```
markylaing: Can we make these structured fields while we're at it? E.g. 
```go
logger.Debug("Authenticated request", slog.String("origin", r.RemoteAddr), slog.String("destination", r.URL.String()), slog.String("fingerprint", fingerprint))
```

example/cmd/microd/main.go
```diff
@@ -87,14 +88,13 @@ func (c *cmdDaemon) run(cmd *cobra.Command, args []string) error {
 	dargs.Hooks = &state.Hooks{
 		// PostBootstrap is run after the daemon is initialized and bootstrapped.
 		PostBootstrap: func(ctx context.Context, s state.State, initConfig map[string]string) error {
-			logCtx := logger.Ctx{}
-			for k, v := range initConfig {
-				logCtx[k] = v
-			}
+			// We can derive the logger using our custom handler from the app.
+			logger := m.Logger()
```
markylaing: It looks like we have two different ways of accessing a logger. One is via the context, and the other is via `(Microcluster).Logger`. Can we only have the context way? Here we could have something like:
```go
logger, err = microcluster.LoggerFromCtx(ctx)
if err != nil {
    ...
}
```

markylaing (changes requested):
Overall I think this looks great. Just a few questions and nitpicks:
1. Did we want to allow using another logger that is not slog? It might be worth asking other teams what they require.
2. It's kind of annoying to have to handle an error when getting a logger. E.g.:
```go
logger, err := microcluster.LoggerFromContext(ctx)
if err != nil {
    return fmt.Errorf("Failed to get a logger to log details about the thing I was doing")
}
```

I'm wondering if we should just panic if the logger isn't present. It means something really bad or stupid has happened and we want to catch it as early as possible in unit or e2e tests.
3. Can we do away with the `internal/log` package and instead just use exported/unexported types and methods in the `microcluster` package?

roosterfish:
> 1. Did we want to allow using another logger that is not slog? It might be worth asking other teams what they require.

I stumbled upon [this](https://github.com/golang/example/blob/master/slog-handler-guide/README.md#introduction) and can recommend quickly going over it. It's linked from here https://pkg.go.dev/log/slog#hdr-Writing_a_handler.

Generally we have to implement some "frontend" framework which offers us the different logging methods like `Debug` and `Error`. Those are fixed as we have to work with something we know. 
However how those logs get rendered or logged (maybe even sent to an external system) should be out of our control. Therefore users can pass a custom handler (or any of the standard ones) in order to control this essential piece. 

Also I found that there are a lot of log sinks (handlers) implemented for various backends so to me it looked like this is the most flexible way to go.
If we start accepting different types of loggers, we might not have all the different log level funcs etc available. As an example the old style `log` package which just has the funcs like `Print`.

tomponline:
> If we start accepting different types of loggers, we might not have all the different log level funcs etc available. As an example the old style `log` package which just has the funcs like `Print`.

If we can get away without accepting multiple logger implementations that would be good.

internal/log/logger.go
roosterfish: No need to export, thanks fixed.

rest/access/handlers.go
roosterfish: Fixed. Also changed one more `logger.Error` where this made sense (in inotify code).

example/cmd/microd/main.go
roosterfish: I went with the app's `Logger` so that a user doesn't have to deal with `LoggerFromCtx` and perform the error check for something which should be present from the users perspective in any case.
It's not populated by the user but internally by Microcluster.

So this means a user would always use `microcluster.Logger` when wanting to log something, but internally where we care about the context and properly populate the logger. In addition we can also access it from there.

There are essentially two places a user might want to use logging: hooks and API handlers.
And both are defined after initializing the Microcluster app as you have to pass them when starting the daemon, not when initializing the Microcluster app struct.
So you can always rely on `app.Logger` to get a logging instance which is using the handler provided during initialization.

roosterfish:
Just updated this PR as merging https://github.com/canonical/microcluster/pull/518 caused conflicts. Also ensured that the main context is now passed further down into the `internal/db` package so that the DB funcs can also consume the logger.

example/cmd/microd/main.go
markylaing: Ok fine for now. Ultimately I think we should always have the logger coming from context as it means a) we always have to pass a context around, which is generally good practice, and b) Microcluster can populate log fields such as the caller address and certificate fingerprint

markylaing (dismissed):
Overall I think this is fine. I like the usage of context to store the logger and I'm fine with making the base logger (with configured handler) available via `(Microcluster).Logger`. We can make it available from context to end users if requested, this could include fields added when requests are authenticated.

example/cmd/microd/main.go
roosterfish: Let's do it right from the beginning. As the context is already passed down to both hooks and API handlers, we can have this func defined on the app construct:

```go
// LoggerFromCtx returns a logger instance using the provided logging handler.
// A default handler is used if none is provided.
// If the context doesn't contain a logger, a new one is returned instead.
func (m *MicroCluster) LoggerFromCtx(ctx context.Context) *slog.Logger {
	logger, err := log.LoggerFromCtx(ctx)
	if err != nil {
		// If the logger cannot be retrieved from the context, return a new logger with the defined logging handler.
		return slog.New(m.args.LogHandler)
	}

	return logger
}

```

It has the benefit that we can fallback to the logging handler set on the receiver `m` in case we were not able to find a logger on the context. This way a user never has to care about whether or not a logger can be returned from the context. This also helps in scenarios where a user passes a custom context in which case we still return a logger using the right handler but just some of the log fields you mentioned would be missing.

What do you think?

roosterfish: This also gives us more flexibility as we can freely add logging fields on the fly and the user doesn't need to switch to another method to get the "right" logger (from the ctx).

markylaing: Sounds good :) 

Can we log an error if there is no logger available in the context? e.g.
```go
// LoggerFromCtx returns a logger instance using the provided logging handler.
// A default handler is used if none is provided.
// If the context doesn't contain a logger, a new one is returned instead.
func (m *MicroCluster) LoggerFromCtx(ctx context.Context) *slog.Logger {
	logger, err := log.LoggerFromCtx(ctx)
	if err != nil {
		// If the logger cannot be retrieved from the context, return a new logger with the defined logging handler.
		logger = slog.New(m.args.LogHandler)
		logger.Error("Failed to get a logger from context", /* fields & error + maybe stacktrace? we should never get here IMO */)
		return logger
	}

	return logger
}
```

Can we also please rename to `LoggerFromContext`?

roosterfish: Pushed the changes now :) Only a few lines though.

roosterfish: Fixed, also renamed the internal `log.LoggerFromCtx` to `LoggerFromContext`.

markylaing (approved):
LGTM :rocket:

mseralessandri:
refers to https://github.com/canonical/microcluster/issues/503

PR #527, opened 2025-10-31
Don't require LXD's `shared` URL type

roosterfish:
A cosmetic followup on https://github.com/canonical/microcluster/issues/503.

In order to use the exposed client funcs (e.g. `Query` and `Websocket`), the URL required to use LXD's shared `api.URL` type with helpers around the standard `url.URL`. 
As it's in the `shared` package reuse is desired but let's also drop this dependency to be fully decoupled from a users perspective of Microcluster as this single type doesn't justify importing the LXD dep.

To not require any user of Microcluster to import `github.com/canonical/lxd` just to use the query functions, we fallback to the standard `url.URL`.
But this doesn't prevent us internally from using LXD's "extended" struct as we can always pass the underlying struct when requested as it's embedded by `api.URL`.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -3,8 +3,8 @@ package client
 import (
 	"context"
 	"net/http"
+	"net/url"
 
-	"github.com/canonical/lxd/shared/api"
 	"github.com/gorilla/websocket"
 
 	"github.com/canonical/microcluster/v3/internal/rest/client"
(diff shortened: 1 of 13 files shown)


client/client.go
```diff
@@ -23,7 +23,7 @@ func IsNotification(r *http.Request) bool {
 
 // Query is a helper for initiating a request on any endpoints defined external to microcluster. This function should be used for all client
 // methods defined externally from microcluster.
-func (c *Client) Query(ctx context.Context, method string, prefix types.EndpointPrefix, path *api.URL, in any, out any) error {
```
markylaing, line 26: I'm assuming this won't be backported to v2 since it is a breaking change?

markylaing (approved):
LGTM great work!

client/client.go
roosterfish: Jup, only v3.

PR #552, opened 2025-12-02
Restructure (2): Consolidate packages and move to `microcluster`

roosterfish:
This is the second PR taken from a single commit in https://github.com/canonical/microcluster/pull/534.

In this PR types which might end up on the user's end through the various channels (client funcs, app funcs, API handlers) are now moved to the central `microcluster/*` package which is the regular entry point for Microcluster users to derive an app construct.


kadinsayani (commented):
The new package structure LGTM overall. Shall we also update the Makefile `update-schema` target in this PR?

```
❯ make update-schema
go generate ./cluster/...
pattern ./cluster/...: lstat ./cluster/: no such file or directory
make: *** [Makefile:66: update-schema] Error 1
```

Apologies, I initially posted this comment on #534 by accident.

roosterfish:
> Apologies, I initially posted this comment on #534 by accident.

Left a comment there, let me add another commit to this PR which drops this.

kadinsayani (dismissed):

roosterfish:
@kadinsayani missed two more lines in code comments.

kadinsayani (commented):
Shall we also drop the generation directives from `internal/cluster`?

kadinsayani (approved):
LGTM, thanks!

PR #562, opened 2025-12-09
Sys: Fix missing hosts in server certificate

roosterfish:
A regression was introduced in https://github.com/canonical/microcluster/pull/526/commits/fd2bc7656d079f127d5572f031806f84a1677643 when trying to remove dependencies to LXD's util package.

The original util.LoadServerCert func also calls shared.KeyPairAndCA but with CertOptions{AddHosts:true}. This was missed when running shared.KeyPairAndCA directly which causes the Microcluster server certificate not being populated with the respective SAN.

This only affects Microcluster v3/edge as it was never backported and was identified when testing MicroCloud v3 using latest Microcluster.

### DIFF
--- a/internal/sys/os.go
+++ b/internal/sys/os.go
@@ -118,7 +118,8 @@ func (s *OS) DatabasePath() string {
 
 // ServerCert gets the local server certificate from the state directory.
 func (s *OS) ServerCert() (*shared.CertInfo, error) {
-	cert, err := shared.KeyPairAndCA(s.stateDir, string(types.ServerCertificateName), shared.CertServer, shared.CertOptions{})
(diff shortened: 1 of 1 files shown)


roosterfish:
Currently running the pipeline on https://github.com/canonical/microcloud/pull/1110 with those fixes to verify.

roosterfish:
Confirmed this fixes the issue regarding `Failed to authenticate using mTLS` in MicroCloud due to a Microcluster service certificate without SAN.

tugbataluy (approved):
I'm still getting familiar with the codebase, so I can't say I fully understand the issue, but based on the PR description, the change seem reasonable.

PR #563, opened 2025-12-09
Response: Remove odd response body discard reader

roosterfish:
The ParseResponse function always reads from the response which invalidates the purpose of the body discard reader.
It's good Go practice to always read the response and then close the body but not anymore required in this case.

As a neat side effect we remove the logging statement from the client func which is odd anyway as such an error should rather be returned directly and not logged as this func is not only used as part of the daemon (which has an initialized logger) but can also be called by clients directly.

### DIFF
--- a/microcluster/types/response.go
+++ b/microcluster/types/response.go
@@ -5,7 +5,6 @@ import (
 	"database/sql"
 	"encoding/json"
 	"errors"
-	"fmt"
 	"net/http"
 	"os"
 
@@ -235,10 +234,6 @@ func ParseResponse(resp *http.Response) (*api.Response, error) {
 
 	err := decoder.Decode(&response)
 	if err != nil {
-		if resp.StatusCode != http.StatusOK {
-			return nil, fmt.Errorf("Failed to fetch %q: %q", resp.Request.URL.String(), resp.Status)
-		}
-
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


tugbataluy (dismissed):
Approved. As fat as I understand, this is a focused cleanup that eliminates unnecessary logic, improves architectural separation, and ensures errors are handled properly at the caller level rather than being logged internally.

roosterfish:
I'll leave it open as I want to verify it further. Also I want to investigate why this was added in the first place.
As it's not causing any issues rn there is no pressure right now to get it merged. Just something I found along the way.

microcluster/rest/response/response.go
markylaing: Can we also remove this? Or at least allow a range of response codes 200-299? It seems weird that if someone adds a microcluster endpoint like an operation returning a 202 then this function will behave differently in the error scenario

markylaing (dismissed):
Good to double check where this came from though I suspect it was just copied from LXD or elsewhere. 

Good to get rid of this cruft though, any errors that result from reading the response body would be caught during the JSON decoding.

microcluster/rest/response/response.go
roosterfish: Good point, this entire section was landed in the [initial blob](https://github.com/canonical/microcluster/commit/2f97da8b47340d4e68f22c3b366a9a05742611a0), so I suspect we can try to live without it. Removed.

markylaing (approved):
Smashing thanks 👍

PR #578, opened 2026-01-06
Fix SelectRandom panic on empty clusters (v3 Breaking Change)

### LINKED ISSUE #432: `SelectRandom` panics on single-node deployment.
[The function that should return client for one of the cluster members](https://github.com/canonical/microcluster/blob/82058a9853b143a0f85739aea4f87f7d4820af97/client/cluster.go#L12-L15) panics if the deployment has only one member.

It seems that on single-node deployments, the size of the "cluster" (`c`) is 0, causing it to call `rand.Intn` function with an invalid argument `0`.

Stacktrace:
```
2025/07/14 15:40:51 http: panic serving @: invalid argument to Intn
goroutine 155 [running]:
net/http.(*conn).serve.func1()
/snap/go/10828/src/net/http/server.go:1903 +0xbe      
panic({0x9d55c0?, 0xbafdb0?})
/snap/go/10828/src/runtime/panic.go:770 +0x132
math/rand.(*Rand).Intn(0xab45f0?, 0xc000444660?)
/snap/go/10828/src/math/rand/rand.go:180 +0x4c        
math/rand.Intn(0x0)
/snap/go/10828/src/math/rand/rand.go:453 +0x25        
github.com/canonical/microcluster/v2/client.Cluster.SelectRandom({0xff8980, 0x0, 0x140?})
/root/go/pkg/mod/github.com/canonical/microcluster/v2@v2.0.5/client/cluster.go:14 +0x5b
```

I don't have live code with this bug anywhere, but this snippet roughly shows the reproducer.

```go
func ovnCentralIpsUpdated(ctx context.Context, s state.State, key string, _ string) error {
	errMsgPrefix := fmt.Sprintf("handling of '%s' config failed.", key)
	cluster, err := s.Cluster(false)
	if err != nil {
		logger.Errorf("failed to trigger OVN environment refresh. %v", err)
		errMsg := fmt.Sprintf(
			"%s Failed to trigger environment update in the OVN cluster. "+
				"Config value was not successfully applied. Please see logs for more details.",
			errMsgPrefix,
		)
		return fmt.Errorf("%s", errMsg)
	}

	client := cluster.SelectRandom()
        ...
```

tugbataluy:
## Problem

The existing SelectRandom() function panics when called on empty clusters due to rand.Intn(0) being invoked ([as mentioned in](https://github.com/canonical/microcluster/issues/432)). This occurs in three distinct cluster states:

    Fresh/uninitialized cluster (not bootstrapped, not joined) - len(c) == 0

    Single-node cluster - len(c) == 1 (works but unnecessary randomization)

    Multi-node cluster - len(c) > 1 (works as intended)

## Root Cause
SelectRandom() doesn't differentiate between cluster states and attempts random selection even when no clients are available, causing a runtime panic.
 
## Solution
Solution - v3 Breaking Change Approach
Fixes #432

This PR introduces a breaking change by modifying the SelectRandom() function signature to handle all cluster states gracefully with proper error handling:

**Updated Function Signature**
```
// Before (v2):
func (c Cluster) SelectRandom() Client

// After (v3):
func (c Cluster) SelectRandom() (*Client, error)
```
**Behavior Changes**
-Empty cluster: Returns (nil, error) with descriptive message instead of panicking
-Single-node cluster: Returns [(&client, nil)](vscode-file://vscode-app/snap/code/215/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) directly (optimized - no randomization needed)
-Multi-node cluster: Returns (&randomClient, nil) 

## Changes Made

-Modified SelectRandom() signature to return (*Client, error)
-Added proper error handling for uninitialized clusters
-Implemented state-aware logic with switch statement for optimal handling

## Migration Required
```
// Old v2 code:
client := cluster.SelectRandom() // Could panic

// New v3 code:
client, err := cluster.SelectRandom()
if err != nil {
    log.Printf("Failed to select client: %v", err)
    return err
}
// Use client safely...

```

## Next Steps
v2 compatible version: A separate PR will be created for v2 that maintains backward compatibility using the SelectRandomClient() approach. Which is [here](https://github.com/canonical/microcluster/pull/579)

### DIFF
--- a/client/cluster.go
+++ b/client/cluster.go
@@ -2,6 +2,7 @@ package client
 
 import (
 	"context"
+	"fmt"
 	"math/rand"
 	"sync"
 )
@@ -10,8 +11,18 @@ import (
 type Cluster []Client
 
 // SelectRandom returns a randomly selected client.
-func (c Cluster) SelectRandom() Client {
-	return c[rand.Intn(len(c))]
+func (c Cluster) SelectRandom() (*Client, error) {
+	switch len(c) {
+	case 0:
+		// Returns an error if the cluster is uninitialized (not bootstrapped, not joined).
+		return nil, fmt.Errorf("Cluster is uninitialized or has no members")
+	case 1:
+		// Returns the only available client if cluster size is 1.
+		return &c[0], nil
+	default:
+		// Returns a randomly selected client for clusters with multiple members.
+		return &c[rand.Intn(len(c))], nil
+	}
 }
 
 // Query executes the given hook across all members of the cluster.
--- a/internal/state/state.go
+++ b/internal/state/state.go
@@ -188,6 +188,11 @@ func (s *InternalState) Cluster(isNotification bool) (client.Cluster, error) {
 		}
 	}
 
+	// Return error if no other cluster members exist.
+	if len(clients) == 0 {
+		return nil, fmt.Errorf("No other cluster members available.")
+	}
+
 	return clients, nil
 }
 


client/cluster.go
```diff
@@ -10,8 +11,29 @@ import (
 type Cluster []Client
 
 // SelectRandom returns a randomly selected client.
+// Deprecated: This function panics on empty clusters. Use SelectRandomClient instead.
 func (c Cluster) SelectRandom() Client {
```
markylaing: I'm not sure I'm a fan of just deprecating the function. Perhaps this can be the solution in the v2 branch, but in v3 I think we should change the function signature to return a `(*Client, error)`.

client/cluster.go
```diff
@@ -10,8 +11,29 @@ import (
...
+	client, err := c.SelectRandomClient()
+	if err != nil {
+		panic(err)
+	}
+
+	return client
+}
+
+// SelectRandomClient returns a randomly selected client.
+func (c Cluster) SelectRandomClient() (Client, error) {
```
markylaing: Can this please return a `(*Client, error)`? It's generally convention throughout LXD/micro* that if something returns a (result, error) tuple then the result is nullable.

client/cluster.go
```diff
@@ -10,8 +11,29 @@ import (
 type Cluster []Client
```
markylaing, line 11: Can we please have a look at where this type is initialized by microcluster and how it is possible to get a `Cluster` of length zero in the first place?

It looks like if we change this function https://github.com/canonical/microcluster/blob/5d64e30142fa35a55cf34401331b1df051919fd3/internal/state/state.go#L169 to return an error if there are no remotes other than ourselves then it shouldn't be possible to get into this situation without manually instantiating the Cluster type.

markylaing (changes requested):

client/cluster.go
tugbataluy: Thank you for your help and guidance, @markylaing . I will first open separate PRs for v2 and v3 (signature change for v3 and defining the deprecated function + new function for v2).

Based on your comment here, I understood this from what you said:
If we look at the number of cluster members when the cluster is first initialized, wouldn't you say there's no need to change SelectRandom() at all? I want to ask that.

markylaing: Yes, please search for where `client.Cluster` is initialized within microcluster and, if possible, return an error if there are no members other than the current one (that is executing the initialization).

IMO `SelectRandom` should still be changed because it should never be possible to panic (especially if not gated by any Go convention such as "Must").

tugbataluy: I examined the code and as far as I understand and estimate first initialization happens the line you pin pointed.

**Where client.Cluster is initialized**: The main initialization happens in state.go in the Cluster() method (line 169).

**How it works:**

1. s.Cluster() calls remotes.Cluster() in remotes.go
2. remotes.Cluster() creates a slice with capacity r.Count()-1 (expecting to exclude the current node)
3. It iterates through r.Addresses() to create clients for each remote
4. Back in state.go, it filters out the current node from the list

**How zero-length clusters are possible:**

- Single-node cluster: Right after bootstrap, before any other nodes join.
- All other nodes left: When other cluster members have been removed or failed.
- Network partitions: During certain failure scenarios where truststore doesn't have reachable members.

**Fix**:
I did what you have suggested already. I added a control statement for cluster member length in state.go

internal/state/state.go
```diff
@@ -188,6 +188,11 @@ func (s *InternalState) Cluster(isNotification bool) (client.Cluster, error) {
 		}
 	}
 
+	// Return error if no other cluster members exist.
+	if len(clients) == 0 {
+		return nil, fmt.Errorf("No other cluster members available.")
```
roosterfish, line 193: I think this change is fine, but we cannot backport it as it changes the behavior of the `Cluster()` func which can be in use by downstream projects.

So lets keep it just in `v3`.

client/cluster.go
```diff
@@ -10,8 +11,18 @@ import (
...
 
 // SelectRandom returns a randomly selected client.
-func (c Cluster) SelectRandom() Client {
-	return c[rand.Intn(len(c))]
+func (c Cluster) SelectRandom() (*Client, error) {
```
roosterfish, line 14: Usually the process is as follows:

We fix something in the edge branch (in this case `v3`) and then backport the fix to `v2`.
I was about to suggest to please first add a ("backportable") commit which doesn't change the signature of `SelectRandom` and just returns a `nil` client. 
However this would still cause a panic on the callers end if the return value isn't checked so I would consider this func and its current behavior broken anyway. 
So let's accept this and fix it by changing the func signature as you have suggested in this commit.

client/cluster.go
```diff
@@ -10,8 +11,18 @@ import (
...
+	switch len(c) {
+	case 0:
+		// Returns an error if the cluster is uninitialized (not bootstrapped, not joined).
+		return nil, fmt.Errorf("Cluster is uninitialized or has no members.")
```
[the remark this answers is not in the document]
roosterfish: Can you please apply this suggestion?

client/cluster.go
roosterfish: So let's then backport this commit as is.
AFAIK `SelectRandom` isn't used by any of the Micro* projects (and k8s) so it should be safe to change anyway. But worth double checking.

roosterfish, line 14: Also the change to return a `Client` pointer seems reasonable as the other `Query` func also supplies a pointer to the provided callback func. This makes it more consistent.

roosterfish (changes requested):

client/cluster.go
roosterfish: Agree. We could add such a header when backporting the fix commit. However as the func is ultimately broken anyway, we should be fine changing it's signature. There is no point in keeping something "stable" which is already broken :)

client/cluster.go
```diff
@@ -10,8 +11,18 @@ import (
 type Cluster []Client
 
 // SelectRandom returns a randomly selected client.
-func (c Cluster) SelectRandom() Client {
-	return c[rand.Intn(len(c))]
+func (c Cluster) SelectRandom() (*Client, error) {
+	switch len(c) {
+	case 0:
+		// Returns an error if the cluster is uninitialized (not bootstrapped, not joined).
+		return nil, fmt.Errorf("cluster is uninitialized or has no members")
```
roosterfish: Please start the error with uppercase.

roosterfish (changes requested):

roosterfish (approved):
LGTM!

Have you had a change to check the remaining `SelectRandom` funcs?

markylaing (approved):

tugbataluy:
> LGTM!
> 
> Have you had a change to check the remaining `SelectRandom` funcs?

No, unfortunately not yet. I'll be checking them soon, yesterday I couldn't focus on that.

PR #579, opened 2026-01-06
Fix: SelectRandom panic on empty clusters with backward compatibility (v2)

tugbataluy:
## Problem

The existing SelectRandom() function panics when called on empty clusters due to rand.Intn(0) being invoked (https://github.com/canonical/microcluster/issues/432). This occurs in three distinct cluster states:

Fresh/uninitialized cluster (not bootstrapped, not joined) - len(c) == 0

Single-node cluster - len(c) == 1 (works but unnecessary randomization)

Multi-node cluster - len(c) > 1 (works as intended)

Root Cause

SelectRandom() doesn't differentiate between cluster states and attempts random selection even when no clients are available, causing a runtime panic.
Solution

Solution - v2 Backward Compatible Approach
Fixes https://github.com/canonical/microcluster/issues/432

This PR introduces a backward-compatible solution by adding a new safe function while preserving existing behavior for legacy code:

**New Safe Function**

```
// New safe function with proper error handling
func (c Cluster) SelectRandomClient() (*Client, error)
```

**Maintained Compatibility**
```
// Original function - unchanged signature, preserved panic behavior
func (c Cluster) SelectRandom() Client
```

**Implementation Strategy**

- Added SelectRandomClient(): New function with proper error handling for all cluster states
- Refactored SelectRandom(): Now calls SelectRandomClient() internally but maintains identical behavior
- Zero breaking changes: Existing code continues to work unchanged

## Changes Made
**New SelectRandomClient() function:**

- Returns (*Client, error) for safe error handling
- Handles empty cluster: Returns (nil, error) instead of panicking
- Optimizes single-node: Returns client directly without randomization
- Multi-node behavior: Returns randomly selected client

**Backward-compatible SelectRandom() refactor:**

- Maintains exact same signature: func (c Cluster) SelectRandom() Client
- Preserves panic behavior for empty clusters (strict backward compatibility)
- Added deprecation warning with migration guidance
- Internal implementation uses new safe function

### DIFF
--- a/client/cluster.go
+++ b/client/cluster.go
@@ -2,6 +2,7 @@ package client
 
 import (
 	"context"
+	"fmt"
 	"math/rand"
 	"sync"
 )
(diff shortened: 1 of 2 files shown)


client/cluster.go
```diff
@@ -10,8 +11,29 @@ import (
...
+	client, err := c.SelectRandomClient()
+	if err != nil {
+		panic(err)
+	}
+
+	return *client
+}
+
+// SelectRandomClient returns a randomly selected client.
+func (c Cluster) SelectRandomClient() (*Client, error) {
```
roosterfish, line 25: Let's not overcomplicate it. We can just backport the one commit from the other PR which fixes `SelectRandom` which is already broken. 

Ultimately we don't want to cause a panic anymore, so the only option we have is to fix what is broken, even if this breaks the function signature.

tugbataluy: As we discussed I'm closing that PR.

PR #582, opened 2026-01-12
Restructure (4): HTTP client interface with cluster wide queries

roosterfish:
This is the fourth PR taken from a single commit in https://github.com/canonical/microcluster/pull/534.

In this PR two new `Client` and `Connector` interfaces are added to address all use cases when it comes to cluster wide HTTP connections.
In addition we now have app level funcs for all user facing operations which previously required going to a separate client package.
Furthermore the specific query functions are now moved to `internal/` as they should not be exposed directly to the end user.


microcluster/types/client.go
```diff
@@ -0,0 +1,95 @@
...
+			}
+		}(client)
+	}
+
+	// Wait for all queries to complete and check for any errors.
+	wg.Wait()
+	for _, err := range errors {
+		if err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
roosterfish: The code was just moved but looks using an errgroup is more reasonable here.

example/cmd/microctl/cluster_members.go
```diff
@@ -104,26 +103,7 @@ func (c *cmdClusterMembersList) run(cmd *cobra.Command, args []string) error {
...
-			return err
-		}
-	}
-
-	return c.listClusterMembers(cmd.Context(), client)
-}
-
-func (c *cmdClusterMembersList) listClusterMembers(ctx context.Context, client *client.Client) error {
-	clusterMembers, err := client.GetClusterMembers(ctx)
+	clusterMembers, err := m.GetClusterMembers(context.TODO())
```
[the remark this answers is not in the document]
roosterfish: It's the CLI and example package. Nothing to fix here.

internal/state/state.go
```diff
@@ -217,12 +220,56 @@ func (s *InternalState) Leader() (*client.Client, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	clusterClientNum := len(clusterClients)
+
+	switch clusterClientNum {
+	case 0:
+		// Returns an error if the cluster is uninitialized (not bootstrapped, not joined).
+		return nil, fmt.Errorf("Cluster is uninitialized or has no members")
```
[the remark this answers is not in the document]
roosterfish: This is a good to keep defensive guard rail to not end up causing a panic. A slice of clients could also potentially be crafted manually. Nothing to fix here.

internal/state/state.go
```diff
@@ -217,12 +220,56 @@ func (s *InternalState) Leader() (*client.Client, error) {
...
+	c, err := internalClient.New(*url, s.ServerCert(), publicKey, isNotification)
 	if err != nil {
 		return nil, err
 	}
 
-	return &client.Client{Client: *c}, nil
+	return c, nil
+}
+
+func (s *InternalState) Member(url *url.URL, isNotification bool, cert *x509.Certificate) (types.Client, error) {
```
markylaing, line 233: Comment missing

example/cmd/microctl/cluster_members.go
tugbataluy, line 47: I believe the other example files (example/api and example/client) contain comment lines explaining the functions. However, these comments are not present for the functions in this file. Was this done intentionally? Or was it overlooked?

internal/rest/client/cluster.go
```diff
@@ -19,46 +19,46 @@ func withTimeoutIfUnset(ctx context.Context) (context.Context, context.CancelFun
...
+		endpoint = endpoint.WithQuery("force", "1")
 	}
 
-	return &tokenResponse, nil
+	return c.Query(queryCtx, "PUT", types.InternalEndpoint, &endpoint.URL, nil, nil)
 }
 
-// ResetClusterMember clears the state directory of the cluster member, and re-execs its daemon.
-func ResetClusterMember(ctx context.Context, c *Client, name string, force bool) error {
+// AddClusterMember records a new cluster member in the trust store of each current cluster member.
```
markylaing, line 22: Why did these functions get switched around?

internal/rest/client/tokens.go
```diff
@@ -10,34 +10,34 @@ import (
...
 	var token string
 	tokenRecord := types.TokenRequest{Name: name, ExpireAfter: expireAfter}
-	err := c.QueryStruct(queryCtx, "POST", types.ControlEndpoint, &api.NewURL().Path("tokens").URL, tokenRecord, &token)
+	err := c.Query(queryCtx, "POST", types.ControlEndpoint, &api.NewURL().Path("tokens").URL, tokenRecord, &token)
 
 	return token, err
 }
 
 // DeleteTokenRecord deletes the toekn record.
-func (c *Client) DeleteTokenRecord(ctx context.Context, name string) error {
```
markylaing, line 25: Can we please have a consistent argument ordering? E.g. always start with ctx, client, ...

markylaing (changes requested):
Just a couple of nits.

Overall I think this is a good simplification - exemplified by the example package!

internal/rest/client/cluster.go
roosterfish: I don't anymore remember the exact reasoning, but I usually tend to define those "CRUD" funcs in order, meaning start by create, then list, update, and finish with delete. 
However this looks to be scrambled up here. Let me fix this.

example/cmd/microctl/cluster_members.go
roosterfish: Putting comments on lowercase/unexported funcs is optional. 
However the linter doesn't pick this rule up consistently, it failed at least for one occasion in this PR, see https://github.com/canonical/microcluster/pull/582#discussion_r2685875491.

internal/rest/client/tokens.go
roosterfish: Yes this looks odd! Thanks for catching.

internal/state/state.go
roosterfish: Fixed.

internal/rest/client/cluster.go
roosterfish: Fixed.

internal/rest/client/tokens.go
roosterfish: Fixed. Also went through some of the other client files in this package as some where also scrambled up.
This also made me realize we require the two missing funcs `UpdateServers` and `UpdateCertificates` on the app as both client methods are now internal only. This is the same approach as with all the other funcs (e.g. delete a cluster member).

markylaing (approved):
Looks great, thanks :)

PR #590, opened 2026-01-20
Restructure (6): `Store` interface for local truststore access

roosterfish:
This is the sixth PR taken from a single commit in https://github.com/canonical/microcluster/pull/534 (even though in this case the actual changes are very different from the original commit).

In this PR the `Remotes` func from the `State` interface is replaced with an interface to not anymore return an implementation of the truststore with types located inside the `internal/` package.
Furthermore the naming is made a bit more consistent:

* `Store` represents a truststore
* `Remote` represents an entry in the truststore

The initial idea was to just hide the truststore access. However it's a very inexpensive way of querying for the existence and details of cluster members as the truststore is already located in memory.

### DIFF
--- a/internal/rest/rest.go
+++ b/internal/rest/rest.go
@@ -273,7 +273,7 @@ func HandleEndpoint(state state.State, mux *mux.Router, version string, e rest.E
 			handleRequest = handleDatabaseRequest
 		}
 
-		trusted, err := access.Authenticate(state, r, state.Address().Host, state.Remotes().CertificatesNative())
+		trusted, err := access.Authenticate(state, r, state.Address().Host, state.Truststore().RemoteCertificatesNative())
 		if err != nil && !errors.As(err, &access.ErrInvalidHost{}) {
 			resp = response.Forbidden(fmt.Errorf("Failed to authenticate request: %w", err))
 		} else {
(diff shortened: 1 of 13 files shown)


tugbataluy (approved):
I think your implementation is solid, looks great as-is, ready to ship!

internal/rest/rest.go
```diff
@@ -273,7 +273,7 @@ func HandleEndpoint(state state.State, mux *mux.Router, version string, e rest.E
 			handleRequest = handleDatabaseRequest
 		}
 
-		trusted, err := access.Authenticate(state, r, state.Address().Host, state.Remotes().CertificatesNative())
+		trusted, err := access.Authenticate(state, r, state.Address().Host, state.Truststore().RemoteCertificatesNative())
```
markylaing, line 276: Not for this PR but we should revisit this authentication function at some point. We shouldn't need to copy the whole trust store for every API call. Instead we should look at the peer certificates any retrieve exactly one matching certificate by fingerprint, then perform the full byte-for-byte comparison.

markylaing: Created an issue for this.

markylaing (approved):
Great work 🚀

internal/rest/rest.go
roosterfish: Nice thanks, I was just about to request this.

PR #591, opened 2026-01-20
Signal Handling Refactor: Improve Flexibility and Control

### LINKED ISSUE #564: Don't subscribe to signals by default
When starting a Microcluster daemon we automatically subscribe to signals to cancel the daemons context which might not be desired by all importers of Microcluster.
Instead leave it up to the importer to subscribe to specific signals.

For ease of use we should offer another func on the app struct which takes care of this. Maybe even with a default set of signals.

tugbataluy:
## Overview
This PR adds built-in signal handling to MicroCluster with a simple and intuitive API. Users can now specify which signals should trigger graceful daemon shutdown directly in their DaemonArgs.

Fixes #564 

## Motivation
Previously, MicroCluster had hardcoded signal handling (SIGPWR, SIGTERM, SIGINT, SIGQUIT) that couldn't be customized. This made it difficult for users who wanted different signal behavior, such as:
- Only responding to SIGINT for development.
- Custom signal combinations for different deployment scenarios.
- No signal handling when managing the lifecycle externally.

## Solution
Add a `ShutdownSignals []os.signal` field to `daemon.Args` that allowa users to specify exactly which signals should trigger graceful shutdown. The implementation uses Go's standard `signal.NotifyContext()` for clean, efficient signal handling.

## Changes Made
 **Core Implementation**
- `daemon.go`: Add `ShutdownSignals []os.Signal` fields to `Args` struct.
- `app.go`: Implement signal handling using `signal.NotifyContext()` when signals are configured.
 
**Example Usage**
- `main.go` : Demonstrate usage with common shutdown signals.

**Documentation**
- `README.md`: Add Signal Handling section with usage examples.

**Basic Usage**
```
// Standard shutdown signals
dargs.ShutdownSignals = []os.Signal{unix.SIGPWR, unix.SIGTERM, unix.SIGINT, unix.SIGQUIT}

// Development,  only Ctrl+C
dargs.ShutdownSignals = []os.Signal{unix.SIGINT}

// No signal handling (manage externally)
dargs.ShutdownSignals = nil // or omit field
```
This adds a clean, flexible signal handling mechanism that covers the most common use cases while keeping the API simple and intuitive.

### DIFF
--- a/example/cmd/microd/main.go
+++ b/example/cmd/microd/main.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/spf13/cobra"
+	"golang.org/x/sys/unix"
 
 	"github.com/canonical/microcluster/v3/example/api"
 	"github.com/canonical/microcluster/v3/example/database"
@@ -82,6 +83,9 @@ func (c *cmdDaemon) run(cmd *cobra.Command, args []string) error {
 		ExtensionsSchema: database.SchemaExtensions,
 		APIExtensions:    api.Extensions(),
 		ExtensionServers: api.Servers,
+
+		// Set up signal handling with custom signals, in this example SIGPWR is omitted.
+		ShutdownSignals: []os.Signal{unix.SIGTERM, unix.SIGINT, unix.SIGQUIT},
 	}
 
 	// exampleHooks are some example post-action hooks that can be run by MicroCluster.
--- a/microcluster/app.go
+++ b/microcluster/app.go
@@ -76,6 +76,9 @@ func App(args Args) (*MicroCluster, error) {
 
 // Start starts up a brand new MicroCluster daemon. Only the local control socket will be available at this stage, no
 // database exists yet. Any api or schema extensions can be applied here.
+// If daemonArgs.ShutdownSignals is provided, the daemon will automatically handle those signals for graceful shutdown.
+// If nil, default signals (SIGPWR, SIGTERM, SIGINT, SIGQUIT) will be used for graceful shutdown.
+// If set to an empty slice, no signal handling will be performed.
 func (m *MicroCluster) Start(ctx context.Context, daemonArgs DaemonArgs) error {
 	logger := m.LoggerFromContext(ctx)
 
@@ -83,11 +86,18 @@ func (m *MicroCluster) Start(ctx context.Context, daemonArgs DaemonArgs) error {
 	defer logger.Info("Daemon stopped")
 	d := daemon.NewDaemon()
 
-	chIgnore := make(chan os.Signal, 1)
-	signal.Notify(chIgnore, unix.SIGHUP)
+	// Use default signals if none provided (nil slice).
+	// If an empty but not-nil slice is provided, no signal handling will be performed.
+	if daemonArgs.ShutdownSignals == nil {
+		daemonArgs.ShutdownSignals = []os.Signal{unix.SIGPWR, unix.SIGTERM, unix.SIGINT, unix.SIGQUIT}
+	}
 
-	ctx, cancel := signal.NotifyContext(ctx, unix.SIGPWR, unix.SIGTERM, unix.SIGINT, unix.SIGQUIT)
-	defer cancel()
+	// Handle shutdown signals if any are configured.
+	if len(daemonArgs.ShutdownSignals) > 0 {
+		var cancel context.CancelFunc
+		ctx, cancel = signal.NotifyContext(ctx, daemonArgs.ShutdownSignals...)
+		defer cancel()
+	}
 
 	// Attach the logger to the parent context.
 	ctx = context.WithValue(ctx, log.CtxLogger, logger)
--- a/README.md
+++ b/README.md
@@ -119,6 +119,42 @@ for i, result := range batch.Results {
 
 Learn more about the Dqlite database in [doc/database.md](doc/database.md).
 
+### Signal Handling
+
+MicroCluster provides built-in signal handling for graceful daemon shutdown. By default, MicroCluster handles `SIGPWR`, `SIGTERM`, `SIGINT`, and `SIGQUIT` signals. You can customize this behavior by setting `ShutdownSignals` in your `DaemonArgs`:
+
+```go
+import "golang.org/x/sys/unix"
+
+// Use default signals (SIGPWR, SIGTERM, SIGINT, SIGQUIT)
+dargs := microcluster.DaemonArgs{
+    // ShutdownSignals not set (nil), defaults will be used.
+}
+
+// Or customize the signals:
+dargs := microcluster.DaemonArgs{
+    ShutdownSignals: []os.Signal{unix.SIGTERM, unix.SIGINT, unix.SIGQUIT}, // Omit SIGPWR.
+    // other configuration...
+}
+
+// Disable signal handling entirely:
+dargs := microcluster.DaemonArgs{
+    ShutdownSignals: []os.Signal{}, // Empty slice, no signal handling.
+}
+```
+
+For simpler cases, you can handle only specific signals:
+
+```go
+// Shutdown only on SIGINT (Ctrl+C)
+dargs := microcluster.DaemonArgs{
+    ShutdownSignals: []os.Signal{unix.SIGINT},
+}
+```
+
+When `ShutdownSignals` is not set (nil), MicroCluster uses default signals for graceful shutdown. When set to an empty slice, no signal handling is performed.
+
+
 ### Create your own API endpoints
 
 ```go
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -73,6 +73,11 @@ type Args struct {
 	// DrainConnectionsTimeout is the amount of time to allow for all core server connections to drain when shutting down.
 	// If it's 0, the connections are not drained when shutting down.
 	DrainConnectionsTimeout time.Duration
+
+	// ShutdownSignals are the signals that will trigger daemon shutdown.
+	// If nil, default signals (SIGPWR, SIGTERM, SIGINT, SIGQUIT) will be used.
+	// If set to an empty slice, no signal handling will be performed.
+	ShutdownSignals []os.Signal
 }
 
 // Daemon holds information for the microcluster daemon.


microcluster/app.go
```diff
@@ -74,21 +72,26 @@ func App(args Args) (*MicroCluster, error) {
 	}, nil
 }
 
+// Logger returns a logger instance using the MicroCluster's configured logging handler.
+func (m *MicroCluster) Logger() *slog.Logger {
```
roosterfish: Why don't you use `LoggerFromContext` which already exists on `m *Microcluster`? 
You can call it with an arbitrary context, if there is no logger defined on it, it will derive a new one.

microcluster/signals.go
```diff
@@ -0,0 +1,148 @@
...
+		}
+	}
+
+	return false
+}
+
+// StartWithConfig starts the MicroCluster daemon with the provided signal configuration.
+// If config is nil, no signal handling will be configured (equivalent to NoSignalConfig).
+// If config.ShutdownSignals is empty and config.SignalHandler is nil, the daemon will start without any signal handling.
+func (m *MicroCluster) StartWithConfig(ctx context.Context, daemonArgs DaemonArgs, config *SignalConfig) error {
```
roosterfish: Can we please make the `SignalConfig` another field in the `DaemonArgs`? This way we always start a Microcluster daemon with the `Start` func but can optionally set a config for the signals.

This is something we also did in the past when adding further functionality as we can freely extend the args.

microcluster/signals.go
```diff
@@ -0,0 +1,148 @@
...
+	"context"
+	"log/slog"
+	"os"
+	"os/signal"
+
+	"golang.org/x/sys/unix"
+)
+
+// SignalConfig defines signal handling configuration for the MicroCluster daemon.
+type SignalConfig struct {
```
roosterfish: Can you please define this type in the `types` package?

microcluster/signals.go
```diff
@@ -0,0 +1,148 @@
...
+	ShutdownSignals []os.Signal
+
+	// IgnoredSignals are signals that will be caught but not acted upon.
+	IgnoredSignals []os.Signal
+
+	// SignalHandler is a custom handler for signals. If provided, it will be called instead of the default behavior (calling cancel).
+	// The handler receives the signal and the context cancel function, allowing custom shutdown logic.
+	SignalHandler func(os.Signal, context.CancelFunc)
+
+	// StopOnFirstSignal determines whether the signal handler should exit after processing the first shutdown signal. Default is true.
```
roosterfish: > Default is true.

When you derive a new instance of the `SignalConfig`, wouldn't `StopOnFirstSignal` be `false` by default as long as you don't set something else? The comment seems misleading as you probably refer to what is set by default in the `DefaultSignalConfig`.

microcluster/signals.go
```diff
@@ -0,0 +1,148 @@
...
+type SignalConfig struct {
+	// ShutdownSignals are the signals that will trigger daemon shutdown.
+	ShutdownSignals []os.Signal
+
+	// IgnoredSignals are signals that will be caught but not acted upon.
+	IgnoredSignals []os.Signal
+
+	// SignalHandler is a custom handler for signals. If provided, it will be called instead of the default behavior (calling cancel).
+	// The handler receives the signal and the context cancel function, allowing custom shutdown logic.
+	SignalHandler func(os.Signal, context.CancelFunc)
```
roosterfish: If you define the `SignalHandler`, does this make setting `ShutdownSignals` and `IgnoredSignals` obsolete?

microcluster/signals.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+// handleSignals is a dedicated signal handler goroutine that processes signals according to the provided configuration. It runs until either a shutdown signal is received (when StopOnFirstSignal is true) or the done channel is closed.
+func (m *MicroCluster) handleSignals(sigChan <-chan os.Signal, config *SignalConfig, cancel context.CancelFunc, done <-chan struct{}) {
+	for {
+		select {
+		case sig := <-sigChan:
+			m.handleSignal(sig, config, cancel)
+			// Exit after first shutdown signal if it's configured.
+			if config.StopOnFirstSignal && !m.isIgnoredSignal(sig, config) {
```
roosterfish: Why does `StopOnFirstSignal` need to be configurable?

roosterfish (changes requested):
Thanks for tackling this.

We usually extend the daemon args (see the PR which added the `DrainConnectionsTimeout`) when adding new features on daemon startup.
This also ensures you don't have to worry about picking the right `Start*` method as you can optionally provide the signal config in case you want to.

microcluster/signals.go
tugbataluy: I thought it might have a good use in theory:
Signal escalation: SIGTERM = graceful shutdown, second SIGINT = faster shutdown
Multi-phase shutdowns: Different signals trigger different shutdown behaviors
Debugging: Log all signals received during shutdown process.

But in practice, it doesn't work that way, meaning it doesn't provide any advantage. It would probably make more sense not to make it configurable.

microcluster/signals.go
tugbataluy: SignalHandler determines how to handle received signals. That is, the user can provide the desired behavior, the Shutdown signal and ignored signals must be defined in the first place.

microcluster/types/signals.go
```diff
@@ -0,0 +1,35 @@
...
+
+// SignalConfig defines signal handling configuration for the MicroCluster daemon.
+type SignalConfig struct {
+	// ShutdownSignals are the signals that will trigger daemon shutdown.
+	ShutdownSignals []os.Signal
+
+	// IgnoredSignals are signals that will be caught but not acted upon.
+	IgnoredSignals []os.Signal
+
+	// SignalHandler is a custom handler for signals. If provided, it will be called instead of the default behavior (calling cancel).
```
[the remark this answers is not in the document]
tugbataluy: It is written correctly.

microcluster/signals.go
tugbataluy: Done :)

microcluster/signals.go
tugbataluy: Done :)

microcluster/types/signals.go
```diff
@@ -0,0 +1,35 @@
...
+	"golang.org/x/sys/unix"
+)
+
+// SignalConfig defines signal handling configuration for the MicroCluster daemon.
+type SignalConfig struct {
+	// ShutdownSignals are the signals that will trigger daemon shutdown.
+	ShutdownSignals []os.Signal
+
+	// IgnoredSignals are signals that will be caught but not acted upon.
+	IgnoredSignals []os.Signal
```
roosterfish: Isn't `IgnoredSignals` just the delta of all signals excluding `ShutdownSignals`? Why do we need to set those too?

If I want to shutdown the daemon only when receiving a `SIGINT`, probably I don't want to specify all the others I am not interested in.

microcluster/app.go
```diff
@@ -76,21 +76,79 @@ func App(args Args) (*MicroCluster, error) {
...
+// If daemonArgs.SignalConfig is provided, the daemon will automatically handle signals according to the configuration.
+// If daemonArgs.SignalConfig is nil, the caller is responsible for managing the context and signal handling.
 func (m *MicroCluster) Start(ctx context.Context, daemonArgs DaemonArgs) error {
 	logger := m.LoggerFromContext(ctx)
 
 	// Start up a daemon with a basic control socket.
 	defer logger.Info("Daemon stopped")
 	d := daemon.NewDaemon()
 
-	chIgnore := make(chan os.Signal, 1)
```
roosterfish, line 86: Do you know why this was originally in there? Seems like a noop to me as we write to the channel when we observe `unix.SIGHUP` but as the channel isn't used anywhere else I wonder what is the purpose of this overall. And probably that caused you to add the to be ignored signals in the signal config?

microcluster/app.go
```diff
@@ -76,21 +76,79 @@ func App(args Args) (*MicroCluster, error) {
...
-	signal.Notify(chIgnore, unix.SIGHUP)
+	// Attach the logger to the parent context.
+	ctx = context.WithValue(ctx, log.CtxLogger, logger)
+
+	// Handle signals if SignalConfig is provided
+	if daemonArgs.SignalConfig != nil {
```
roosterfish: Can we make this a lot simpler by just doing something like the following:

```go
if daemonArgs.SignalConfig != nil {
    ctx, cancel := signal.NotifyContext(ctx, daemonArgs.SignalConfig.Signals...)
}
```

I see you wanted to allow adding an additional handler. But AFAIK you can always also just run `signal.Notify` outside of Microcluster in case you want to subscribe to a specific signal? Because then this logic is up to the user of Microcluster and we don't have to carry it around.

In this case we could think about having just `ShutdownSignals []os.Signal` instead of `SignalConfig` on the daemon args. What do you think?

example/cmd/microd/main.go
```diff
@@ -203,6 +203,10 @@ func (c *cmdDaemon) run(cmd *cobra.Command, args []string) error {
 		},
 	}
 
+	// Set up signal handling with default configuration
+	dargs.SignalConfig = types.DefaultSignalConfig()
```
roosterfish: Good call to add this to the example package.

roosterfish (changes requested):

microcluster/app.go
tugbataluy: I'm not sure the main motive behind that. I tried to examine the commit history and I guess, it was created to capture SIGHUP signals so they wouldn't trigger the default signal handler, but the signals were never actually processed ,they just sat in the channel.

microcluster/app.go
tugbataluy: It makes sense, I'll remove IgnoredSignals and simplify it by just using ShutdownSignals []os.Signal.

microcluster/app.go
```diff
@@ -76,18 +75,20 @@ func App(args Args) (*MicroCluster, error) {
...
 	d := daemon.NewDaemon()
 
-	chIgnore := make(chan os.Signal, 1)
-	signal.Notify(chIgnore, unix.SIGHUP)
-
-	ctx, cancel := signal.NotifyContext(ctx, unix.SIGPWR, unix.SIGTERM, unix.SIGINT, unix.SIGQUIT)
-	defer cancel()
+	// Handle shutdown signals if configured
+	if len(daemonArgs.ShutdownSignals) > 0 {
+		_, cancel := signal.NotifyContext(ctx, daemonArgs.ShutdownSignals...)
```
markylaing:
```suggestion
		ctx, cancel = signal.NotifyContext(ctx, daemonArgs.ShutdownSignals...)
```

We'll need to set `ctx` to the returned value and pass it into `d.Run` (otherwise the run context will not be cancelled when a signal is received)

markylaing (changes requested):

microcluster/app.go
tugbataluy: I realize from this that I don't fully understand context logic. I made adjustments based on your comment, @markylaing :smiley: .Thank you.
Now it's time to do some research on contexts.

markylaing:
If back ported, users of `v2` will suddenly no longer be able to stop microcluster via os signal. 

I think we should keep this for v3, but add a default list of os signals for v2 (to match current behaviour). 

I'd like to get @roosterfish's opinion on that also

example/cmd/microd/main.go
```diff
@@ -82,6 +83,9 @@ func (c *cmdDaemon) run(cmd *cobra.Command, args []string) error {
 		ExtensionsSchema: database.SchemaExtensions,
 		APIExtensions:    api.Extensions(),
 		ExtensionServers: api.Servers,
+
+		// Set up signal handling with default signals for graceful shutdown
+		ShutdownSignals: []os.Signal{unix.SIGPWR, unix.SIGTERM, unix.SIGINT, unix.SIGQUIT},
```
roosterfish: Thanks for keeping the original list of signals.

Before you had something like a `DefaultSignals` func which I found interesting. Now I wonder if we should rather continue setting those four as defaults as a user can now overwrite this behavior using `ShutdownSignals` list? 

Maybe in the example you can show how to customize it and omit e.g. `SIGPWR` and only set the remaining three.

We don't need an extra func for the defaults, but in the `Start` func if `ShutdownSignals` list is empty, we can just assign those four signals. 

That would also allow seamless migration from `v2` to `v3` as you don't have to worry about the signals being changed, but you can now influence them as you wish.

roosterfish (changes requested):
Looking good now as it tries to fix the real problem of being able to influence the list of signals. Only one more remark about the `v2` to `v3` upgrade path.

example/cmd/microd/main.go
roosterfish: Ah yes, @markylaing was faster, just saw your comment but looks we are on the same page then :) https://github.com/canonical/microcluster/pull/591#issuecomment-3783647121

tugbataluy: I think that's a pretty reasonable approach. As you said, I can show how the signals can be customized in the example and keep the ones I kept as default (the 4 signals that were previously subscribed to) as default signals. Without requiring a function, of course :smile:

tugbataluy: I made the relevant changes. I hope I got it right.

microcluster/app.go
```diff
@@ -76,17 +76,22 @@ func App(args Args) (*MicroCluster, error) {
 
 // Start starts up a brand new MicroCluster daemon. Only the local control socket will be available at this stage, no
 // database exists yet. Any api or schema extensions can be applied here.
+// If daemonArgs.ShutdownSignals is provided, the daemon will automatically handle those signals for graceful shutdown.
+// If empty, default signals (SIGPWR, SIGTERM, SIGINT, SIGQUIT) will be used for graceful shutdown.
 func (m *MicroCluster) Start(ctx context.Context, daemonArgs DaemonArgs) error {
 	logger := m.LoggerFromContext(ctx)
-
```
roosterfish, line 81: Can we still keep this line to separate the comment from the line that retrieves the logger?

microcluster/app.go
```diff
@@ -76,17 +76,22 @@ func App(args Args) (*MicroCluster, error) {
...
-	chIgnore := make(chan os.Signal, 1)
-	signal.Notify(chIgnore, unix.SIGHUP)
+	// Use default signals if none provided.
+	if len(daemonArgs.ShutdownSignals) == 0 {
+		daemonArgs.ShutdownSignals = []os.Signal{unix.SIGPWR, unix.SIGTERM, unix.SIGINT, unix.SIGQUIT}
+	}
 
-	ctx, cancel := signal.NotifyContext(ctx, unix.SIGPWR, unix.SIGTERM, unix.SIGINT, unix.SIGQUIT)
+	// Handle shutdown signals.
+	var cancel context.CancelFunc
```
roosterfish: You can just do `ctx, cancel :=` as `cancel` isn't yet defined and then ommit the `var` line above.

roosterfish (changes requested):
Two small nits.

markylaing:
Before merging, make sure we differentiate between a nil slice (default signals) and an empty but not-nil slice (no signal handling)

tugbataluy:
I updated the code based on your comments. I also noted in the relevant places that if the shutdown signal is nil, the default signals are used, and if it is an empty slice, the signal will not be handled :sparkle: .

microcluster/app.go
```diff
@@ -76,18 +76,29 @@ func App(args Args) (*MicroCluster, error) {
...
+	}
 
-	ctx, cancel := signal.NotifyContext(ctx, unix.SIGPWR, unix.SIGTERM, unix.SIGINT, unix.SIGQUIT)
-	defer cancel()
+	// Handle shutdown signals if any are configured.
+	if len(daemonArgs.ShutdownSignals) > 0 {
+		// Use a different variable name to avoid shadowing ctx, then assign it back.
+		signalCtx, cancel := signal.NotifyContext(ctx, daemonArgs.ShutdownSignals...)
+		defer cancel()
+		ctx = signalCtx
```
roosterfish: After moving it back into the `if` you do need the extra `var` line (remember my earlier comment).

Was it a copilot suggestion to not shadow the `ctx`? I think it's fine as it's what we want in this case, and we use the same pattern also in other places.

roosterfish (changes requested):

microcluster/app.go
tugbataluy: Yep, so what is your suggestion?
should I declare var cancel context.CancelFunc and use it or do something else? Because if I don't 
go gives the error "ctx was declared but never used"

roosterfish (approved):
LGTM!

microcluster/app.go
roosterfish: Thanks, current version looks good to me.

markylaing (approved):
Thanks!

PR #593, opened 2026-01-20
Move dqlite roles to core_cluster_member_roles

kadinsayani:
Introduces a dedicated `core_cluster_member_roles` table and migrates dqlite role storage out of `core_cluster_members`. Updates runtime queries and upgrade helpers to read/write roles from the new table, adds a migration test for update V6, and updates docs to reflect the new storage location.

Associated spec: https://docs.google.com/document/d/1qt5OPBW0NDGENWLKg5T3VYOhct1UCS6FUJ9RZ3kKPxg/edit?usp=sharing.

### DIFF
--- a/internal/cluster/cluster_members.go
+++ b/internal/cluster/cluster_members.go
@@ -87,10 +87,11 @@ WHERE name IN ('api_extensions');
 	}
 
 	// Fetch all cluster members with a smaller schema version than we expect.
-	stmt = `SELECT id, name, address, certificate, schema_internal, schema_external, %s, heartbeat, role
+	stmt = fmt.Sprintf(`SELECT id, name, address, certificate, schema_internal, schema_external, %s, heartbeat, core_cluster_member_roles.dqlite_role
   FROM %s
+  JOIN core_cluster_member_roles ON core_cluster_member_roles.member_id = %s.id
   ORDER BY name
-	`
+	`, "%s", tableName, tableName)
 
 	// If API extensions are supported, ensure the list for each cluster member also matches what we expect,
 	// and only return cluster members for whom it does not.
@@ -99,7 +100,7 @@ WHERE name IN ('api_extensions');
 		apiField = "api_extensions"
 	}
 
-	stmt = fmt.Sprintf(stmt, apiField, tableName)
+	stmt = fmt.Sprintf(stmt, apiField)
 	allMembers, err = getCoreClusterMembersRaw(ctx, tx, stmt)
 	if err != nil {
 		return nil, nil, err
--- a/internal/cluster/cluster_members.mapper.go
+++ b/internal/cluster/cluster_members.mapper.go
@@ -16,21 +16,24 @@ import (
 var _ = api.ServerEnvironment{}
 
 var coreClusterMemberObjects = clusterDB.RegisterStmt(`
-SELECT core_cluster_members.id, core_cluster_members.name, core_cluster_members.address, core_cluster_members.certificate, core_cluster_members.schema_internal, core_cluster_members.schema_external, core_cluster_members.api_extensions, core_cluster_members.heartbeat, core_cluster_members.role
+SELECT core_cluster_members.id, core_cluster_members.name, core_cluster_members.address, core_cluster_members.certificate, core_cluster_members.schema_internal, core_cluster_members.schema_external, core_cluster_members.api_extensions, core_cluster_members.heartbeat, core_cluster_member_roles.dqlite_role
   FROM core_cluster_members
+  JOIN core_cluster_member_roles ON core_cluster_member_roles.member_id = core_cluster_members.id
   ORDER BY core_cluster_members.name
 `)
 
 var coreClusterMemberObjectsByAddress = clusterDB.RegisterStmt(`
-SELECT core_cluster_members.id, core_cluster_members.name, core_cluster_members.address, core_cluster_members.certificate, core_cluster_members.schema_internal, core_cluster_members.schema_external, core_cluster_members.api_extensions, core_cluster_members.heartbeat, core_cluster_members.role
+SELECT core_cluster_members.id, core_cluster_members.name, core_cluster_members.address, core_cluster_members.certificate, core_cluster_members.schema_internal, core_cluster_members.schema_external, core_cluster_members.api_extensions, core_cluster_members.heartbeat, core_cluster_member_roles.dqlite_role
   FROM core_cluster_members
+  JOIN core_cluster_member_roles ON core_cluster_member_roles.member_id = core_cluster_members.id
   WHERE ( core_cluster_members.address = ? )
   ORDER BY core_cluster_members.name
 `)
 
 var coreClusterMemberObjectsByName = clusterDB.RegisterStmt(`
-SELECT core_cluster_members.id, core_cluster_members.name, core_cluster_members.address, core_cluster_members.certificate, core_cluster_members.schema_internal, core_cluster_members.schema_external, core_cluster_members.api_extensions, core_cluster_members.heartbeat, core_cluster_members.role
+SELECT core_cluster_members.id, core_cluster_members.name, core_cluster_members.address, core_cluster_members.certificate, core_cluster_members.schema_internal, core_cluster_members.schema_external, core_cluster_members.api_extensions, core_cluster_members.heartbeat, core_cluster_member_roles.dqlite_role
   FROM core_cluster_members
+  JOIN core_cluster_member_roles ON core_cluster_member_roles.member_id = core_cluster_members.id
   WHERE ( core_cluster_members.name = ? )
   ORDER BY core_cluster_members.name
 `)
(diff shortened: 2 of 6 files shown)


tomponline:
@kadinsayani @roosterfish why do we even need to store the dqlite role in the DB?

internal/db/update/update.go
```diff
@@ -95,6 +96,55 @@ ALTER TABLE core_token_records_new RENAME TO core_token_records;
 	return err
 }
 
+// updateFromV6 moves cluster member roles into a dedicated table.
+func updateFromV6(ctx context.Context, tx *sql.Tx) error {
+	stmt := `
+CREATE TABLE core_cluster_member_roles_new (
+  member_id     INTEGER  NOT NULL,
+  control_plane INTEGER  NOT NULL DEFAULT 0,
```
roosterfish, line 104: Does this refer to the failure domain dqlite setting? If yes should we name it failure domain? Maybe this is covered in the spec but I cannot view it right now :)

internal/db/update/update.go
```diff
@@ -95,6 +96,55 @@ ALTER TABLE core_token_records_new RENAME TO core_token_records;
 	return err
 }
 
+// updateFromV6 moves cluster member roles into a dedicated table.
+func updateFromV6(ctx context.Context, tx *sql.Tx) error {
+	stmt := `
+CREATE TABLE core_cluster_member_roles_new (
+  member_id     INTEGER  NOT NULL,
```
roosterfish, line 103: Worth adding a foreign key to `core_cluster_members`?

internal/db/update/update.go
```diff
@@ -95,6 +96,55 @@ ALTER TABLE core_token_records_new RENAME TO core_token_records;
...
+  UNIQUE(certificate)
+);
+
+INSERT INTO core_cluster_members_new (id, name, address, certificate, schema_internal, schema_external, api_extensions, heartbeat)
+SELECT id, name, address, certificate, schema_internal, schema_external, api_extensions, heartbeat FROM core_cluster_members;
+
+DROP TABLE core_cluster_members;
+ALTER TABLE core_cluster_members_new RENAME TO core_cluster_members;
+
+CREATE TABLE core_cluster_member_roles (
```
roosterfish, line 131: Shall we keep the LXD naming scheme and do `core_cluster_members_roles`? I am fine with both but maybe that aligns both better together.

internal/db/update/update.go
```diff
@@ -95,6 +96,55 @@ ALTER TABLE core_token_records_new RENAME TO core_token_records;
...
+  member_id     INTEGER  NOT NULL REFERENCES core_cluster_members(id) ON DELETE CASCADE,
+  control_plane INTEGER  NOT NULL DEFAULT 0,
+  dqlite_role   TEXT     NOT NULL,
+  PRIMARY KEY (member_id)
+);
+
+INSERT INTO core_cluster_member_roles (member_id, control_plane, dqlite_role)
+SELECT member_id, control_plane, dqlite_role FROM core_cluster_member_roles_new;
```
roosterfish, line 139: Do we need this `INSERT .. SELECT`? As this is a new table there should be nothing we can select from.

roosterfish: Same for the following `DROP`.

internal/db/update/update.go
roosterfish: You can already create this new table with it's final name in the first line of this patch. No need to rename as there isn't yet and roles table. Then you also don't need the `SELECT ... INSERT` below.

At this stage the table already exists with the right content, just with the wrong name as it has the `_new` suffix.

internal/db/update/cluster.go
```diff
@@ -117,7 +117,12 @@ func GetClusterMemberSchemaVersions(ctx context.Context, tx *sql.Tx) (internalSc
 		return nil, nil, err
 	}
 
-	sql := fmt.Sprintf("SELECT %s.schema_internal,%s.schema_external FROM %s JOIN core_cluster_member_roles ON core_cluster_member_roles.member_id = %s.id WHERE NOT core_cluster_member_roles.dqlite_role='pending'", tableName, tableName, tableName, tableName)
+	roleClause, roleField, err := roleSelectClause(ctx, tx, tableName)
```
roosterfish, line 120: I see, you have to support both the old and new schema versions as `GetClusterMemberSchemaVersions` gets executed before upgrading the schema to the final version using the new roles table?

roosterfish: If this is the case please add a comment providing some context on this.

internal/cluster/cluster_members.go
```diff
@@ -87,10 +87,11 @@ WHERE name IN ('api_extensions');
 	}
 
 	// Fetch all cluster members with a smaller schema version than we expect.
-	stmt = `SELECT id, name, address, certificate, schema_internal, schema_external, %s, heartbeat, role
+	stmt = fmt.Sprintf(`SELECT id, name, address, certificate, schema_internal, schema_external, %s, heartbeat, core_cluster_member_roles.dqlite_role
```
roosterfish, line 90: Do we have to perform a `LEFT JOIN` on `core_cluster_members` to not return duplicate members in case a member has more than one role?

internal/cluster/cluster_members.mapper.go
```diff
@@ -16,21 +16,24 @@ import (
 var _ = api.ServerEnvironment{}
 
 var coreClusterMemberObjects = clusterDB.RegisterStmt(`
-SELECT core_cluster_members.id, core_cluster_members.name, core_cluster_members.address, core_cluster_members.certificate, core_cluster_members.schema_internal, core_cluster_members.schema_external, core_cluster_members.api_extensions, core_cluster_members.heartbeat, core_cluster_members.role
+SELECT core_cluster_members.id, core_cluster_members.name, core_cluster_members.address, core_cluster_members.certificate, core_cluster_members.schema_internal, core_cluster_members.schema_external, core_cluster_members.api_extensions, core_cluster_members.heartbeat, core_cluster_member_roles.dqlite_role
   FROM core_cluster_members
+  JOIN core_cluster_member_roles ON core_cluster_member_roles.member_id = core_cluster_members.id
```
roosterfish, line 21: Again `LEFT JOIN` necessary?

roosterfish (changes requested):
Thanks for pushing this already, looking forward to check out the spec.

roosterfish:
> @kadinsayani @roosterfish why do we even need to store the dqlite role in the DB?

That's what is documented about what is stored in the `role` column:

>  Dqlite roles (`voter`, `standby`, `spare`) with the addition of `PENDING` for systems that have recently joined
the cluster, but have not yet received a heartbeat through the Dqlite roles adjustment hook. At least 1
non-`PENDING` cluster member must be present to remove other members. Once a heartbeat detects that a `PENDING`
member exists in Dqlite, it will assign it the correct role.

The `role column is also consulted when asking for the state of a cluster member.

internal/db/update/update.go
kadinsayani: No, I haven't considered failure domain at this stage.

internal/db/update/update.go
kadinsayani: Yes, I'll update this :)

internal/db/update/cluster.go
kadinsayani: > I see, you have to support both the old and new schema versions as `GetClusterMemberSchemaVersions` gets executed before upgrading the schema to the final version using the new roles table?

Yes, I'll add a comment explaining this.

kadinsayani:
> > @kadinsayani @roosterfish why do we even need to store the dqlite role in the DB?
> 
> That's what is documented about what is stored in the `role` column:
> 
> > Dqlite roles (`voter`, `standby`, `spare`) with the addition of `PENDING` for systems that have recently joined
> > the cluster, but have not yet received a heartbeat through the Dqlite roles adjustment hook. At least 1
> > non-`PENDING` cluster member must be present to remove other members. Once a heartbeat detects that a `PENDING`
> > member exists in Dqlite, it will assign it the correct role.
> 
> The `role column is also consulted when asking for the state of a cluster member.

In LXD, we just report the dqlite role stored in dqlite from `ToAPI()`. @tomponline raised a good point on stale data, perhaps the role column should just store manual roles and we should consult with dqlite when determining cluster member state.

internal/db/update/update.go
kadinsayani: > Shall we keep the LXD naming scheme and do `core_cluster_members_roles`? I am fine with both but maybe that aligns both better together.

Sure, let's go with `core_cluster_members_roles` for consistency.

internal/db/update/update.go
kadinsayani: We can drop this.

kadinsayani:
@roosterfish appreciate your review and feedback on this. Closing this PR as I'll open a new one that aligns with the daemon config approach and avoids this schema change.

PR #594, opened 2026-01-21
Removes Cluster() returning error on single node cluster.

crypticC0der:
With a single node cluster the heartbeat code fails, which causes the heartbeat round to fail. This can be solved by reverting the changes done to Cluster() to allow it to return an empty list on a single node.

### DIFF
--- a/internal/state/state.go
+++ b/internal/state/state.go
@@ -261,7 +261,7 @@ func (s *InternalState) RandomMember(isNotification bool) (types.Client, error) 
 	switch clusterClientNum {
 	case 0:
 		// Returns an error if the cluster is uninitialized (not bootstrapped, not joined).
-		return nil, fmt.Errorf("Cluster is uninitialized or has no members")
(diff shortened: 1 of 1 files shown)


roosterfish (changes requested):
Can you please update this PR and instead drop the additional check on `Cluster()` as discussed to restore the old behavior for better usability? Thanks.

internal/state/state.go
```diff
@@ -196,11 +196,6 @@ func (s *InternalState) Cluster(isNotification bool) (types.Clients, error) {
 		}
```
roosterfish, line 196: Can you please use our regular commit message scheme, meaning start the first line with the package name:

`internal/state: Revert ...`

roosterfish: Also on the other commit.

roosterfish (changes requested):

internal/state/state.go
crypticC0der: 🫡

roosterfish (approved):
Thanks!

PR #595, opened 2026-01-22
feat: Add address flag for Dqlite force removal

louiseschmidtgen:
# Add address flag for Dqlite force removal

## Problem

During testing of the membership check, we observed an issue when the cluster enters an inconsistent state where a member exists only in Dqlite. In this scenario, the existing force removal flag fails to remove the lingering Dqlite membership.

The root cause is that we rely on the Truststore to look up a node’s address in order to determine which Dqlite member to remove. However, Dqlite has no concept of node names—it only tracks members by internal IDs and their network addresses.

As a result, we cannot remove a Dqlite member by name. In certain failure scenarios, we also cannot query the Truststore or core_cluster_members to determine the address or Dqlite ID associated with the stale member.

## Solution

Introduce an optional --address flag that explicitly identifies the Dqlite member by its address. This allows us to reliably locate and forcefully remove the member from Dqlite, even if it has already been removed or “nuked” elsewhere in the system.

## Side-note

I intentionally avoided introducing a --dqlite-id flag. Dqlite IDs are harder for users to discover (they must be extracted from the Dqlite cluster.yaml), whereas the node’s address is more intuitive and readily available.

### DIFF
--- a/example/test/main.sh
+++ b/example/test/main.sh
@@ -488,7 +488,108 @@ test_membership_consistency() {
     cat /tmp/token_error
   fi
 
+  # Test member force flag overrides during removal with inconsistent state works
+  echo "  -> Testing member force removal with inconsistent state"
+  if microctl --state-dir "${test_dir}/c1" cluster remove c2 --force --address 127.0.0.1:9002; then
+    echo "  -> Member c2 force removal succeeded as expected"
+  else
+    echo "ERROR: Member force removal should have succeeded despite inconsistent state"
+    exit 1
+  fi
+
+  # Generating a new token should now succeed
+  echo "  -> Testing token generation after resolving inconsistency"
+  if microctl --state-dir "${test_dir}/c1" tokens add c5 2>/tmp/token_resp; then
+    echo "  -> Token generation succeeded as expected after resolving inconsistency"
+    cat /tmp/token_resp
+  else
+    echo "ERROR: Token generation should have succeeded after resolving inconsistency"
+    cat /tmp/token_resp
+    exit 1
+  fi
   echo "  -> Membership consistency checks working as expected"
+  
+  shutdown_systems
+}
+
+test_truststore_force_removal() {
+  echo "Testing force removal"
+  
+  new_systems 3 --heartbeat 2s
+  
+  # Bootstrap first member
+  microctl --state-dir "${test_dir}/c1" init "c1" 127.0.0.1:9001 --bootstrap
+  
+  # Join second and third members
+  token_c2=$(microctl --state-dir "${test_dir}/c1" tokens add "c2")
+  microctl --state-dir "${test_dir}/c2" init "c2" 127.0.0.1:9002 --token "${token_c2}"
+  
+  token_c3=$(microctl --state-dir "${test_dir}/c1" tokens add "c3")
+  microctl --state-dir "${test_dir}/c3" init "c3" 127.0.0.1:9003 --token "${token_c3}"
+  
+  # Wait for cluster to stabilize
+  echo "  -> Waiting for cluster to stabilize"
+  retry_count=0
+  max_retries=10
+  while [[ -n "$(microctl --state-dir "${test_dir}/c1" cluster list -f yaml | yq '.[] | select(.role == "PENDING")')" ]] && [[ ${retry_count} -lt ${max_retries} ]]; do
+    sleep 2
+    retry_count=$((retry_count + 1))
+  done
+  
+  echo "  -> Cluster established with 3 members"
+  microctl --state-dir "${test_dir}/c1" cluster list
+
+  # Test force removal of non-existing member and random address
+  echo "  -> Testing force removal of non-existing member with random address"
+  if microctl --state-dir "${test_dir}/c1" cluster remove nonexist --force --address 127.0.0.1:9999 2>/tmp/remove_error; then
+    echo "ERROR: Force removal of non-existing member should have failed"
+    cat /tmp/remove_error
+    exit 1
+  else
+    echo "  -> Force removal of non-existing member failed as expected"
+    cat /tmp/remove_error
+  fi
+  
+  # Simulate truststore corruption: remove c3 from truststore while keeping DB and dqlite entries
+  # Need to remove from all nodes' truststores to prevent repopulation
+  echo "  -> Simulating truststore deletion of c3 from all nodes (keeping DB and dqlite entries)"
+  rm -f "${test_dir}/c1/truststore/c3.yaml"
+  rm -f "${test_dir}/c2/truststore/c3.yaml"
+  rm -f "${test_dir}/c3/truststore/c3.yaml"
+
+  # Attempt normal removal should fail (membership inconsistency detected)
+  echo "  -> Testing normal removal of c3 (should fail due to missing truststore)"
+  if microctl --state-dir "${test_dir}/c1" cluster remove c3 --address 127.0.0.1:9003 2>/tmp/remove_error; then
+    echo "ERROR: Normal removal should have failed"
+    exit 1
+  else
+    echo "  -> Normal removal blocked as expected"
+    cat /tmp/remove_error
+  fi
+  
+  # Force remove with explicit address should succeed
+  echo "  -> Testing force removal of c3 with address override"
+  if microctl --state-dir "${test_dir}/c1" cluster remove c3 --force --address 127.0.0.1:9003; then
+    echo "  -> Force removal of c3 succeeded"
+  else
+    echo "ERROR: Force removal should have succeeded"
+    exit 1
+  fi
+
+  # Now generate a new token - this should succeed because membership is now consistent
+  echo "  -> Testing token generation after force removal (should succeed)"
+  if microctl --state-dir "${test_dir}/c1" tokens add "c4" 2>/tmp/token_resp; then
+    echo "  -> Token generation succeeded - membership is now consistent"
+    cat /tmp/token_resp
+  else
+    echo "ERROR: Token generation should have succeeded after force removal"
+    cat /tmp/token_resp
+    exit 1
+  fi
+  
+  echo "SUCCESS: Force removal of non-existing member and random address blocked as expected"
+  echo "SUCCESS: Force removal of truststore-orphaned node successful"
+  echo "SUCCESS: Verified membership consistency restored after force removal"
   
   shutdown_systems
 }
@@ -580,6 +681,7 @@ if [ "${1:-"all"}" = "all" ] || [ "${1}" = "" ]; then
   test_join_token_before_cluster_formed
   test_extended_endpoints
   test_membership_consistency
+  test_truststore_force_removal
   test_parallel_joins
 elif [ "${1}" = "recover" ]; then
   test_recover
@@ -595,6 +697,8 @@ elif [ "${1}" = "extended" ]; then
   test_extended_endpoints
 elif [ "${1}" = "membership" ]; then
   test_membership_consistency
+elif [ "${1}" = "force-removal" ]; then
+  test_truststore_force_removal
 elif [ "${1}" = "parallel-join" ]; then
   test_parallel_joins
 else
--- a/internal/rest/client/cluster.go
+++ b/internal/rest/client/cluster.go
@@ -58,13 +58,21 @@ func ResetClusterMember(ctx context.Context, c types.Client, name string, force 
 }
 
 // DeleteClusterMember deletes the cluster member with the given name.
-func DeleteClusterMember(ctx context.Context, c types.Client, name string, force bool) error {
+// If `address` is non-empty it is sent as a query parameter to target dqlite
+// removal by address when the truststore cannot map name to address.
+// Dqlite does not track names, so this is useful when the name is no longer
+// resolvable.
+func DeleteClusterMember(ctx context.Context, c types.Client, name string, address string, force bool) error {
 	queryCtx, cancel := withTimeoutIfUnset(ctx)
 	defer cancel()
 
 	endpoint := api.NewURL().Path("cluster", name)
 	if force {
 		endpoint = endpoint.WithQuery("force", "1")
+	}
+
+	if address != "" {
+		endpoint = endpoint.WithQuery("address", address)
 	}
 
 	return c.Query(queryCtx, "DELETE", types.PublicEndpoint, &endpoint.URL, nil, nil)
--- a/internal/rest/resources/cluster.go
+++ b/internal/rest/resources/cluster.go
@@ -417,18 +417,49 @@ func resetClusterMember(ctx context.Context, s state.State, force bool) (reExec 
 // clusterMemberDelete Removes a cluster member from dqlite and re-execs its daemon.
 func clusterMemberDelete(s state.State, r *http.Request) response.Response {
 	force := r.URL.Query().Get("force") == "1"
+	addr := r.URL.Query().Get("address")
 	name, err := url.PathUnescape(mux.Vars(r)["name"])
 	if err != nil {
 		return response.SmartError(err)
 	}
 
-	allRemotes := s.Truststore().RemotesByName()
-	remote, ok := allRemotes[name]
-	if !ok {
-		return response.SmartError(fmt.Errorf("No remote exists with the given name %q", name))
+	ctx := r.Context()
+
+	logger, err := log.LoggerFromContext(ctx)
+	if err != nil {
+		return response.InternalError(err)
 	}
 
-	ctx := r.Context()
+	allRemotes := s.Truststore().RemotesByName()
+	remote, remotePresent := allRemotes[name]
+
+	// Determine the address to use for dqlite removal:
+	// - If remote exists in truststore and no address provided, use the truststore address.
+	// - If remote missing and no address provided, require explicit address.
+	// - If address provided, it must match the truststore address (if remote exists) or be valid (if not).
+	if remotePresent && addr == "" {
+		addr = remote.Address.String()
+	} else if !remotePresent && addr == "" {
+		// If the remote is not present in the truststore and no address is provided, we cannot proceed.
+		return response.SmartError(fmt.Errorf("Cluster member %q not found in truststore; please provide a node address", name))
+	} else if remotePresent && addr != "" && remote.Address.String() != addr {
+		// Reject if provided address doesn't match the truststore address for this remote name.
+		return response.SmartError(fmt.Errorf("Provided address %q does not match the address %q of the remote with name %q", addr, remote.Address.String(), name))
+	} else if !remotePresent && addr != "" {
+		// Remote missing from truststore; validate the fallback address format.
+		addrPort, err := types.ParseAddrPort(addr)
+		if err != nil {
+			return response.SmartError(fmt.Errorf("Invalid address %q: %w", addr, err))
+		}
+
+		// Ensure the fallback address isn't claimed by another remote in the truststore.
+		existingRemote := s.Truststore().RemoteByAddress(addrPort)
+		if existingRemote != nil {
+			return response.SmartError(fmt.Errorf("Address %q is already used by remote %q (address %q); address is only a fallback for %q when it is missing from the truststore", addr, existingRemote.Name, existingRemote.Address.String(), name))
+		}
+
+		logger.Warn("Cluster member not found in truststore; proceeding with provided fallback address", slog.String("member", name), slog.String("address", addr))
+	}
 
 	// Check cluster membership consistency before allowing removals (unless forced)
 	// This ensures core_cluster_members, truststore, and dqlite are all in sync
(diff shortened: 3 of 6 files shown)


internal/rest/resources/cluster.go
```diff
@@ -417,17 +417,22 @@ func resetClusterMember(ctx context.Context, s state.State, force bool) (reExec
...
 
 	allRemotes := s.Truststore().RemotesByName()
-	remote, ok := allRemotes[name]
-	if !ok {
+	remote, remotePresent := allRemotes[name]
+	if !remotePresent && !force {
 		return response.SmartError(fmt.Errorf("No remote exists with the given name %q", name))
 	}
 
+	if remotePresent && addr == "" {
```
roosterfish, line 440: We should not allow overwriting the address in case it was returned by the truststore as this should be the ultimate source of truth. 
Only if the truststore cannot find the member, use the provided address as a fallback.

In addition let's add a check that the provided address (if `addr != ""`) isn't used by any of the other cluster members? There is a `Truststore().RemoteAddresses` func.

internal/rest/resources/cluster.go
```diff
@@ -461,7 +466,7 @@ func clusterMemberDelete(s state.State, r *http.Request) response.Response {
 
 	// If we are not the leader, just forward the request.
 	if leaderInfo.Address != s.Address().Host {
-		if allRemotes[name].Address.String() == s.Address().Host {
+		if remotePresent && allRemotes[name].Address.String() == s.Address().Host {
```
roosterfish: Why don't we want to acquire this log here? Could you add a comment to make this more clear?

internal/rest/resources/cluster.go
```diff
@@ -511,15 +516,15 @@ func clusterMemberDelete(s state.State, r *http.Request) response.Response {
...
+		if node.Address == addr {
 			index = i
 			break
 		}
 	}
 
 	// If we can't find the node in dqlite, that means it failed to fully initialize. It still might have a record in our database so continue along anyway.
 	if index < 0 {
-		logger.Error(fmt.Sprintf("No dqlite record exists for %q, deleting from internal record instead", remote.Name))
+		logger.Error(fmt.Sprintf("No dqlite record exists for %q.", name))
```
roosterfish: Let's also add a log entry in case we used the provided address instead of the entry returned from the truststore by name similar to this log message.

internal/rest/resources/cluster.go
```diff
@@ -417,17 +417,22 @@ func resetClusterMember(ctx context.Context, s state.State, force bool) (reExec
...
+	addr := r.URL.Query().Get("address")
 	name, err := url.PathUnescape(mux.Vars(r)["name"])
 	if err != nil {
 		return response.SmartError(err)
 	}
 
 	allRemotes := s.Truststore().RemotesByName()
-	remote, ok := allRemotes[name]
-	if !ok {
+	remote, remotePresent := allRemotes[name]
```
roosterfish, line 434: Can we please also have validation that checks if the provided address matches the address of the entry returned from the truststore in case everything is consistent?

internal/rest/resources/cluster.go
```diff
@@ -549,7 +554,7 @@ func clusterMemberDelete(s state.State, r *http.Request) response.Response {
 	}
 
 	// If we are removing the leader of a 2-node cluster, ensure the remaining node is a voter.
-	if len(info) == 2 && allRemotes[name].Address.String() == leaderInfo.Address {
+	if len(info) == 2 && remotePresent && allRemotes[name].Address.String() == leaderInfo.Address {
```
roosterfish: What if the broken member is the current dqlite leader? Can we check for 

`if len(info) == 2 && remotePresent && addr == leaderInfo.Address {`? 

If we performed enough validation before, we can be sure we don't reassign the leader role from any other healthy cluster member.

internal/rest/resources/cluster.go
```diff
@@ -567,7 +572,7 @@ func clusterMemberDelete(s state.State, r *http.Request) response.Response {
 	}
 
 	// If we are the leader and removing ourselves, reassign the leader role and perform the removal from there.
-	if allRemotes[name].Address.String() == leaderInfo.Address {
+	if remotePresent && allRemotes[name].Address.String() == leaderInfo.Address {
```
roosterfish: Same here maybe

`addr == leaderInfo.Address`

roosterfish: Otherwise how could we ever end up calling the ` internalClient.DeleteClusterMember` below?

internal/rest/resources/cluster.go
```diff
@@ -632,22 +637,36 @@ func clusterMemberDelete(s state.State, r *http.Request) response.Response {
...
+	} else {
+		logger.Warn("Remote missing; skipping PreRemove hook", slog.String("member", name))
 	}
 
-	// Remove the cluster member from the database.
+	// Remove the cluster member from the database. If we don't have the remote address, try to
+	// find the member by name and delete by the stored address as a best-effort fallback.
 	err = s.Database().Transaction(ctx, func(ctx context.Context, tx *sql.Tx) error {
-		return cluster.DeleteCoreClusterMember(ctx, tx, remote.Address.String())
+		if addr != "" {
```
roosterfish: Can we always perform the DB delete? Because you might provide both valid member name and address.

internal/rest/resources/cluster.go
```diff
@@ -673,14 +692,22 @@ func clusterMemberDelete(s state.State, r *http.Request) response.Response {
...
-		return response.SmartError(err)
+			logger.Warn("Failed connecting to remote for ResetClusterMember (forcing):", slog.String("error", err.Error()))
+		} else {
+			err = internalClient.ResetClusterMember(ctx, client, name, force)
+			if err != nil && !force {
+				return response.SmartError(err)
+			}
+		}
+	} else {
+		logger.Warn("Remote missing; skipping ResetClusterMember", slog.String("member", name))
```
roosterfish: We usually format log messages without ;

You could do something like `Skipping the reset of cluster member %q ...`.

internal/rest/resources/cluster.go
```diff
@@ -673,14 +692,22 @@ func clusterMemberDelete(s state.State, r *http.Request) response.Response {
 		return response.SmartError(err)
 	}
 
-	client, err := s.Connect().Member(remote.URL(), false, publicKey)
-	if err != nil {
-		return response.SmartError(err)
-	}
+	if remotePresent {
+		client, err := s.Connect().Member(remote.URL(), false, publicKey)
```
roosterfish: Can we connect to `addr` instead? Then we have to check for `remotePresent`.

example/test/main.sh
```diff
@@ -488,6 +488,25 @@ test_membership_consistency() {
     cat /tmp/token_error
   fi
 
+  # Test member force flag overrides during removal with inconsistent state works
```
roosterfish, line 491: Can we please test all three cases:

1) dqlite present, truststore entry and DB entry missing
2) dqlite present, truststore entry missing
3) dqlite present, DB entry missing

From what I understand, (1) and (2) are tackled in this PR.

microcluster/app.go
```diff
@@ -323,13 +323,15 @@ func (m *MicroCluster) RevokeJoinToken(ctx context.Context, name string) error {
 }
 
 // RemoveClusterMember removes a member from the cluster.
-func (m *MicroCluster) RemoveClusterMember(ctx context.Context, name string, force bool) error {
+func (m *MicroCluster) RemoveClusterMember(ctx context.Context, name string, force bool, address string) error {
```
roosterfish: I am hesitant having this on the `app` directly. It might suggest you have to set the `address`.
At least we should have a comment explaining the purpose and that it's optional.

roosterfish (changes requested):
Hi, thanks for following up on the removal parts.

Another thought that came to my mind is if we even have to provide the address:
1) Let's say if we still have all members in the truststore, we can derive the address from the given name to cleanup the dqlite entry.
2) If we still have the member in the DB (not truststore), we can derive the address from the `core_cluster_members` table.
3) If it's in neither of those two, we already have the recovery option, see `microctl cluster recover` when building the `exmaple/` package.

So in this case aren't all cases already covered?

example/test/main.sh
louiseschmidtgen: We've got 1. , 3.
I haven't tested nuking just the truststore entry. 
If we nuke the DB entry the truststore entry get's removed when the heartbeat times out.

internal/rest/resources/cluster.go
louiseschmidtgen: yes, I've adjusted these checks slightly now.

louiseschmidtgen:
> Hi, thanks for following up on the removal parts.
> 
> Another thought that came to my mind is if we even have to provide the address:
> 
> 1. Let's say if we still have all members in the truststore, we can derive the address from the given name to cleanup the dqlite entry.
> 2. If we still have the member in the DB (not truststore), we can derive the address from the `core_cluster_members` table.
> 3. If it's in neither of those two, we already have the recovery option, see `microctl cluster recover` when building the `exmaple/` package.
> 
> So in this case aren't all cases already covered?

On 1/2 yes if those are present we can proceed to remove the entries without a problem.

3. The cluster recover command is not suitable for this case as it is used for recovery from quorum loss. If we are nuking a member we still have quorum. It would not make sense to run cluster recover as this command can result in unnecessary interruption and data loss.

internal/rest/resources/cluster.go
louiseschmidtgen: yes, I've changed this section . It should check if it's the current leader whether remote is found or not

internal/rest/resources/cluster.go
louiseschmidtgen: Good point, if it is not I think it's best to return an error.

internal/rest/resources/cluster.go
louiseschmidtgen: Yes, the check is removed.

internal/rest/resources/cluster.go
louiseschmidtgen: I think it would be best to fail a removal where the address does not match the looked up address of the name and let the administrator pass the correct. I am adding a warning if we didnt get a truststore hit but we proceed with the given address.

internal/rest/resources/cluster.go
louiseschmidtgen: Yes, this check is removed.

example/test/main.sh
louiseschmidtgen: added a truststore test.

internal/rest/resources/cluster.go
louiseschmidtgen: yes let's do that!

internal/rest/resources/cluster.go
louiseschmidtgen: Okay, done.

microcluster/app.go
louiseschmidtgen: I've moved the comment here to explain that the address is optional and a fallback.

internal/rest/resources/cluster.go
louiseschmidtgen: I think it makes sense if both are specified to check that they point to the same address. If not I would propose to error.

microcluster/app.go
```diff
@@ -323,13 +323,15 @@ func (m *MicroCluster) RevokeJoinToken(ctx context.Context, name string) error {
 }
 
 // RemoveClusterMember removes a member from the cluster.
-func (m *MicroCluster) RemoveClusterMember(ctx context.Context, name string, force bool) error {
+// Pass address hint (may be empty) to the internal client so the leader
+// can attempt dqlite removal by address when automatic mapping fails.
```
roosterfish: Can you please elaborate a little bit on automatic mapping? It might not be clear at first glance that this means mapping the provided `name` to existing cluster members.

microcluster/app.go
```diff
@@ -323,13 +323,15 @@ func (m *MicroCluster) RevokeJoinToken(ctx context.Context, name string) error {
...
+func (m *MicroCluster) RemoveClusterMember(ctx context.Context, name string, force bool, address string) error {
```
roosterfish: Better to have `force` as the last func argument?

internal/rest/client/cluster.go
```diff
@@ -58,7 +58,9 @@ func ResetClusterMember(ctx context.Context, c types.Client, name string, force
 }
 
 // DeleteClusterMember deletes the cluster member with the given name.
-func DeleteClusterMember(ctx context.Context, c types.Client, name string, force bool) error {
+// If `address` is non-empty it will be passed as a query parameter to
+// target dqlite removal by address when automatic mapping is not available.
```
roosterfish: Again it refers to automatic mapping which is not very descriptive IMHO.

internal/rest/resources/cluster.go
```diff
@@ -417,18 +417,42 @@ func resetClusterMember(ctx context.Context, s state.State, force bool) (reExec
...
-	ctx := r.Context()
+	allRemotes := s.Truststore().RemotesByName()
+	remote, remotePresent := allRemotes[name]
+
+	if remotePresent && addr == "" {
+		addr = remote.Address.String()
+	} else if remotePresent && addr != "" && remote.Address.String() != addr {
+		return response.SmartError(fmt.Errorf("Provided address %q does not match the address %q of the remote with name %q", addr, remote.Address.String(), name))
+	} else if !remotePresent && addr == "" {
+		return response.SmartError(fmt.Errorf("Cluster member %q not found in truststore. Please provide a node address.", name))
```
roosterfish: We should not expect the "inconsistency" to be the default.
In most cases the provided member probably is just a non-existing name in which case providing a node address is not the right approach as it should always be just a fallback.

internal/rest/resources/cluster.go
```diff
@@ -417,18 +417,42 @@ func resetClusterMember(ctx context.Context, s state.State, force bool) (reExec
...
+	} else if !remotePresent && addr != "" {
+		addrPort, err := types.ParseAddrPort(addr)
+		if err != nil {
+			return response.SmartError(fmt.Errorf("Invalid address %q: %w", addr, err))
+		}
+
+		// Check if address is in use by another remote.
+		existingRemote := s.Truststore().RemoteByAddress(addrPort)
+		if existingRemote != nil {
+			return response.SmartError(fmt.Errorf("address %q is already used by remote %q (address %q); please provide the correct address for %q", addr, existingRemote.Name, existingRemote.Address.String(), name))
```
roosterfish: Please start error with uppercase.

internal/rest/resources/cluster.go
roosterfish: We usually don't end errors with `.`.

internal/rest/resources/cluster.go
roosterfish: `please provide the correct address for %q` suggests you have to provide the address even though it's optional and should only be a fallback.

I suggest removing this bit.

internal/rest/resources/cluster.go
```diff
@@ -417,18 +417,42 @@ func resetClusterMember(ctx context.Context, s state.State, force bool) (reExec
...
+		}
+
+		logger.Warn("Cluster member not found in truststore. Proceeding with provided address.", slog.String("member", name), slog.String("address", addr))
```
roosterfish: Please don't end the error with a dot.

roosterfish: If you have copied this from another log message we should clean it up too.

internal/rest/resources/cluster.go
```diff
@@ -417,18 +417,42 @@ func resetClusterMember(ctx context.Context, s state.State, force bool) (reExec
...
+	logger, err := log.LoggerFromContext(ctx)
+	if err != nil {
+		return response.InternalError(err)
 	}
 
-	ctx := r.Context()
+	allRemotes := s.Truststore().RemotesByName()
+	remote, remotePresent := allRemotes[name]
+
+	if remotePresent && addr == "" {
```
roosterfish, line 440: This `if else` section would look great with some more comments to better understand the various scenarios.

internal/rest/resources/cluster.go
```diff
@@ -511,15 +530,15 @@ func clusterMemberDelete(s state.State, r *http.Request) response.Response {
...
+		if node.Address == addr {
 			index = i
 			break
 		}
 	}
 
 	// If we can't find the node in dqlite, that means it failed to fully initialize. It still might have a record in our database so continue along anyway.
 	if index < 0 {
-		logger.Error(fmt.Sprintf("No dqlite record exists for %q, deleting from internal record instead", remote.Name))
+		logger.Error("No dqlite record exists for the member.", slog.String("member", name))
```
roosterfish: Please don't end with .

internal/rest/resources/cluster.go
```diff
@@ -632,22 +651,32 @@ func clusterMemberDelete(s state.State, r *http.Request) response.Response {
 		return response.SmartError(err)
 	}
 
-	// Set the forwarded flag so that the the system to be removed knows the removal is in progress.
-	c, err := internalClient.New(remote.URL(), s.ServerCert(), publicKey, true)
-	if err != nil {
-		return response.SmartError(err)
-	}
+	// If we have the remote information, tell the cluster member to run its PreRemove hook and return.
+	if remotePresent {
```
roosterfish: Do we want to try this also when there is no entry in the truststore? The member might still be lingering around.

roosterfish: Identical to the reset attempt below. Maybe you can even reuse the bits that craft the URL in case `!remotePresent`.

internal/rest/resources/control.go
```diff
@@ -131,7 +131,7 @@ func controlPost(state state.State, r *http.Request) response.Response {
 			<-r.Context().Done()
 
 			// Use `force=1` to ensure the node is fully removed, in case its listener hasn't been set up.
-			err = internalClient.DeleteClusterMember(context.Background(), client, req.Name, true)
+			err = internalClient.DeleteClusterMember(context.Background(), client, req.Name, true, "")
```
roosterfish: It was already in there before, but I think we can use `r.Context()`?

roosterfish (changes requested):
The deletion bits look much more solid now, thanks for putting in the changes and the additional tests.

Please add some more comments so that it is easier to follow through and check the other commits about some minor nits.

internal/rest/resources/cluster.go
louiseschmidtgen: I'll change the order to reflect the more common cases first.

internal/rest/resources/cluster.go
louiseschmidtgen: I'll revise the error message.

internal/rest/resources/cluster.go
louiseschmidtgen: Good call, recycling the logic to apply here as well.

internal/rest/resources/cluster.go
roosterfish: The error still uses a dot at the end.

In general let's try to be consistent with the structure of error messages. Below you use `;` to separate the two sentences, should we rather do the same here instead of the `.` ?

The ; seems to be used also at other places in the codebase so it seems to be a good fit.

internal/rest/resources/cluster.go
```diff
@@ -673,14 +715,18 @@ func clusterMemberDelete(s state.State, r *http.Request) response.Response {
...
-		return response.SmartError(err)
-	}
+		if !force {
+			return response.SmartError(err)
+		}
 
-	err = internalClient.ResetClusterMember(ctx, client, name, force)
-	if err != nil && !force {
-		return response.SmartError(err)
+		logger.Warn("Failed connecting to remote for ResetClusterMember (forcing)", slog.String("error", err.Error()))
```
roosterfish: Just a nit, let's not use the func name in the warning but rather something descriptive like "Failed connecting to cluster member for reset".

roosterfish: And use another slog value to indicate force true/false rather than embedding it in the message itself.
In this case it will only fire if `force=true` but it looks cleaner.

internal/rest/resources/cluster.go
```diff
@@ -417,18 +417,49 @@ func resetClusterMember(ctx context.Context, s state.State, force bool) (reExec
...
 	}
 
-	ctx := r.Context()
+	allRemotes := s.Truststore().RemotesByName()
+	remote, remotePresent := allRemotes[name]
+
+	// Determine the address to use for dqlite removal:
+	// - If remote exists in truststore and no address provided, use the truststore address.
+	// - If remote missing and no address provided, require explicit address.
+	// - If address provided, it must match the truststore address (if remote exists) or be valid (if not).
```
roosterfish, line 439: Much easier now to differentiate the cases, thanks for adding the comment.

When we try to remove a member with a non existing name and random address together with `--force`, the API endpoint now seems to be a "noop". 
As an example you can run `microctl --state-dir <existing member> cluster remove <non existing member name> --address <random address> --force` and the endpoint will return without any hint that nothing has happened. 
If we cannot find the provided address in neither the DB table nor dqlite, should we still return an error? 

I have the feeling it's a best effort strategy to try and run the pre remove hook, but if the address isn't found anywhere else, maybe the API should indicate that too.

roosterfish (changes requested):

internal/rest/resources/cluster.go
louiseschmidtgen: You're right we can make it more explicit with an error like this:

```
  -> Testing force removal of non-existing member with random address
time=2026-02-04T08:26:49.368+02:00 level=WARN msg="Cluster member not found in truststore; proceeding with provided fallback address" member=nonexist address=127.0.0.1:9999
time=2026-02-04T08:26:49.380+02:00 level=ERROR msg="No dqlite record exists for the member" member=nonexist
  -> Force removal of non-existing member failed as expected
Error: Cluster member "nonexist" with address "127.0.0.1:9999" not found in dqlite or database
```

louiseschmidtgen: I've added the error return and a test case for it.

roosterfish (approved):
Thanks, LGTM.

PR #603, opened 2026-01-29
feat: Add response and log functionalities

HomayoonAlimohammadi:
### Context

While migrating Canonical Kubernetes from Microcluster v2 to v3, I noticed that we're moving away from the `lxd/reponse` and it's now baked into Microcluster itself. These two functions were used in Canonical Kubernetes but are not available in Microcluster. Also, the logging package is internal to Microcluster which makes testing on the k8s snap side impossible. The new `pkg/log` is supposed to surface the logging functionality.

### Overview

This pull request introduces new functionality for error handling in HTTP responses and adds a new logging utility package. The most important changes are the addition of helper functions for standardized error responses and the creation of a `log` package to simplify logger usage with context.

**Error handling improvements:**

* Added `Unauthorized` and `ErrorResponse` helper functions to `response.go` for generating 401 and custom error responses, respectively.

**Logging utility:**

* Introduced a new `microcluster/types/log` package that wraps logger context handling, providing a `WithLogger` function and exposing the logger context key for easier logger management from external repos.

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -27,7 +27,6 @@ import (
 	internalConfig "github.com/canonical/microcluster/v3/internal/config"
 	"github.com/canonical/microcluster/v3/internal/db"
 	"github.com/canonical/microcluster/v3/internal/endpoints"
-	internalLog "github.com/canonical/microcluster/v3/internal/log"
 	"github.com/canonical/microcluster/v3/internal/recover"
 	internalREST "github.com/canonical/microcluster/v3/internal/rest"
 	internalClient "github.com/canonical/microcluster/v3/internal/rest/client"
@@ -153,7 +152,7 @@ func NewDaemon() *Daemon {
 // log is a convenience to retrieve the internal logger from the shutdown context.
 // We always expect the logger to be present.
 func (d *Daemon) log() *slog.Logger {
-	return d.shutdownCtx.Value(internalLog.CtxLogger).(*slog.Logger) //nolint:revive
+	return d.shutdownCtx.Value(types.CtxLogger).(*slog.Logger) //nolint:revive
 }
 
 // Run initializes the Daemon with the given configuration, starts the database,
--- a/internal/log/logger.go
+++ b/internal/log/logger.go
@@ -4,16 +4,13 @@ import (
 	"context"
 	"errors"
 	"log/slog"
+
+	"github.com/canonical/microcluster/v3/microcluster/types"
 )
-
-type ctxKey string
-
-// CtxLogger is the name of the context value for the central logger.
-const CtxLogger ctxKey = "logger"
 
 // LoggerFromContext returns the logger from the given context.
 func LoggerFromContext(ctx context.Context) (*slog.Logger, error) {
-	logger, ok := ctx.Value(CtxLogger).(*slog.Logger)
+	logger, ok := ctx.Value(types.CtxLogger).(*slog.Logger)
 	if !ok {
 		return nil, errors.New("Logger does not exist on context")
 	}
--- a/microcluster/app.go
+++ b/microcluster/app.go
@@ -100,7 +100,7 @@ func (m *MicroCluster) Start(ctx context.Context, daemonArgs DaemonArgs) error {
 	}
 
 	// Attach the logger to the parent context.
-	ctx = context.WithValue(ctx, log.CtxLogger, logger)
+	ctx = types.ContextWithLogger(ctx, logger)
 
 	err := d.Run(ctx, m.FileSystem.StateDir(), daemonArgs)
 	if err != nil {
@@ -280,7 +280,7 @@ func (m *MicroCluster) RecoverFromQuorumLoss(members []types.DqliteMember) (stri
 	// Derive a new context with the central logger attached.
 	// As we don't have a running daemon at this stage, we cannot use its context.
 	// Instead we use the logger populated for the app which uses the custom handler if supplied.
-	ctx := context.WithValue(context.Background(), log.CtxLogger, m.LoggerFromContext(context.Background()))
+	ctx := types.ContextWithLogger(context.Background(), m.LoggerFromContext(context.Background()))
 
 	return recover.RecoverFromQuorumLoss(ctx, m.FileSystem, members)
 }
--- a/microcluster/rest/response/response.go
+++ b/microcluster/rest/response/response.go
@@ -3,6 +3,7 @@ package response
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"log/slog"
@@ -116,6 +117,16 @@ func NotImplemented(err error) Response {
 // Unavailable returns an unavailable response (503) with the given error.
 func Unavailable(err error) Response {
 	return &errorResponse{http.StatusServiceUnavailable, err}
+}
+
+// Unauthorized returns an unauthorized response (401) with the given error.
+func Unauthorized(err error) Response {
+	return &errorResponse{http.StatusUnauthorized, err}
+}
+
+// ErrorResponse returns an error response with the given code and msg.
+func ErrorResponse(code int, msg string) Response {
+	return &errorResponse{code, errors.New(msg)}
 }
 
 func (r *errorResponse) Render(w http.ResponseWriter, req *http.Request) error {
--- a/microcluster/types/log.go
+++ b/microcluster/types/log.go
@@ -1,0 +1,24 @@
+package types
+
+import (
+	"context"
+	"log/slog"
+)
+
+type CtxKey string
+
+// CtxLogger is the name of the context value for the central logger.
+const CtxLogger CtxKey = "logger"
+
+// ContextWithLogger returns a new context with the given logger.
+// If no logger is provided, the default logger is used.
+func ContextWithLogger(ctx context.Context, logger ...*slog.Logger) context.Context {
+	var l *slog.Logger
+	if len(logger) == 0 {
+		l = slog.Default()
+	} else {
+		l = logger[0]
+	}
+
+	return context.WithValue(ctx, CtxLogger, l)
+}
--- a/internal/daemon/daemon_test.go
+++ b/internal/daemon/daemon_test.go
@@ -2,7 +2,6 @@ package daemon
 
 import (
 	"context"
-	"log/slog"
 	"path/filepath"
 	"testing"
 
@@ -13,7 +12,6 @@ import (
 
 	"github.com/canonical/microcluster/v3/internal/config"
 	"github.com/canonical/microcluster/v3/internal/endpoints"
-	"github.com/canonical/microcluster/v3/internal/log"
 	"github.com/canonical/microcluster/v3/internal/rest/client"
 	"github.com/canonical/microcluster/v3/internal/sys"
 	"github.com/canonical/microcluster/v3/internal/trust"
@@ -189,7 +187,7 @@ func (t *daemonsSuite) Test_UpdateServers() {
 	for i, test := range tests {
 		t.T().Logf("%s (case %d)", test.name, i)
 
-		ctx := context.WithValue(context.TODO(), log.CtxLogger, slog.Default())
+		ctx := types.ContextWithLogger(context.TODO())
 		commonDir := t.T().TempDir()
 
 		// Create a temp watcher.
--- a/internal/db/db_test.go
+++ b/internal/db/db_test.go
@@ -5,7 +5,6 @@ import (
 	"database/sql"
 	"encoding/json"
 	"fmt"
-	"log/slog"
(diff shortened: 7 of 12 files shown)


pkg/log/log.go
```diff
@@ -0,0 +1,15 @@
...
+	"context"
+	"log/slog"
+
+	internallog "github.com/canonical/microcluster/v3/internal/log"
+)
+
+const CtxLogger = internallog.CtxLogger
+
+// WithLogger returns a new context with the given logger.
+func WithLogger(ctx context.Context, logger *slog.Logger) context.Context {
```
roosterfish: Where do you intend to call this function? If you are running inside a hook or API handler you can derive the logger from the passed context, see the `example/` package.

There is a new `LoggerFromContext` func on the app construct for this purpose.

roosterfish (changes requested):

pkg/log/log.go
HomayoonAlimohammadi: This is the other way around. I'm injecting a logger to my context. I noticed that we need a logger in context or otherwise some function calls will fail with `Logger does not exist on context`, e.g. here: https://github.com/canonical/k8sd/blob/24d02f59776f619fd12aec9e3ada1d6827ce1ca6/pkg/utils/microcluster/state.go#L46

HomayoonAlimohammadi: In fact I've noticed that this `Logger does not exist on context` is causing too many problems. What if we just return a default logger in case it's not available in the context?

roosterfish: > What if we just return a default logger in case it's not available in the context?

Isn't that what is already happening when running `LoggerFromContext`? It should run `slog.New`.

HomayoonAlimohammadi: So I guess there are two `LoggerFromContext`s. 
- https://github.com/canonical/microcluster/blob/v3/microcluster/app.go#L452-L466
which calls
- https://github.com/canonical/microcluster/blob/v3/internal/log/logger.go#L14-L22

The second one basically throws an error

HomayoonAlimohammadi: I've changed this PR a bit. Now instead of introducing the `pkg/log`, I'm returning a default logger from the internal `LoggerFromContext` instead of returning an error to see if it solves this issue.

internal/log/logger.go
```diff
@@ -15,7 +14,9 @@ const CtxLogger ctxKey = "logger"
 func LoggerFromContext(ctx context.Context) (*slog.Logger, error) {
 	logger, ok := ctx.Value(CtxLogger).(*slog.Logger)
 	if !ok {
-		return nil, errors.New("Logger does not exist on context")
+		logger := slog.Default()
```
roosterfish: But this is already caught by the outer `microcluster.LoggerFromContext` isn't it? If this internal `LoggerFromContext` returns an error, then the outer one is running `slog.New`. 

I don't understand why we have to replicate this here too?

roosterfish (changes requested):

roosterfish (changes requested):
Thanks for adding the two additional response funcs. However I still don't see the need for the proposed logging changes.

HomayoonAlimohammadi:
Thank you for the review @roosterfish. I've reached out on MM with more context around this change.

roosterfish:
> Thank you for the review @roosterfish. I've reached out on MM with more context around this change.

Ty for providing the additional context. 
When you test your API handlers as a standalone, the passed context requires the logger to be set (which happens automatically when the handler is fired inside Microcluster).

So as part of our unit tests we also run `ctx := context.WithValue(context.TODO(), log.CtxLogger, slog.Default())` to ensure we have a context with the required logger set to e.g. run the database funcs which might need to access the logger.

Deriving another default logger on the `internal/log.LoggerFromContext` func isn't trivial as we don't know at this point which log handler was passed when starting Microcluster. 
If you want to test the API handler external from Microcluster, pass a context with the `"context"` key populated as shown above.
If we start returning a default logger, we might miss places where we unintentionally pass the wrong context (not the daemon's shutdown ctx), so it takes the default but instead we want the logs to be routed to the actual logger which isn't available as the context is wrong.

HomayoonAlimohammadi:
Thank you for all the information @roosterfish! I've updated the PR with what we've discussed

microcluster/types/log.go
```diff
@@ -6,10 +6,10 @@ import (
 	"log/slog"
 )
 
-type ctxKey string
+type CtxKey string
 
 // CtxLogger is the name of the context value for the central logger.
-const CtxLogger ctxKey = "logger"
+const CtxLogger CtxKey = "logger"
```
roosterfish, line 11: Can we move the `CtxLogger` itself into `microcluster/types/log`? Then we can reference it both from the `internal/` and public packages.

microcluster/types/log/log.go
```diff
@@ -0,0 +1,23 @@
...
+	"log/slog"
+
+	"github.com/canonical/microcluster/v3/internal/log"
+)
+
+const CtxLogger log.CtxKey = log.CtxLogger
+
+// WithLogger returns a new context with the given logger.
+// If no logger is provided, the default logger is used.
+func WithLogger(ctx context.Context, logger ...*slog.Logger) context.Context {
```
roosterfish: I don't understand the complexity of this func. Why does it have to be variadic and accept multiple instances of the logger if only ever the first instance is used? 

If you wanted to add a helper func that takes care of populating the context value, you can just do something like `return context.WithValue(ctx, CtxLogger, slog.Default())`.
Please also ensure this helper is then used instead, for example here https://github.com/canonical/microcluster/blob/v3/internal/db/db_test.go#L655 (and the other places).

microcluster/types/log.go
```diff
@@ -0,0 +1,23 @@
+package types
+
+import (
+	"context"
+	"log/slog"
+
+	"github.com/canonical/microcluster/v3/internal/log"
```
roosterfish: From `microcluster/types` we should not start importing from `internal/` to prevent import loops in the future.
Therefore I mentioned to move the `CtxLogger` into `microcluster/types/log` directly.

roosterfish (changes requested):

microcluster/types/log/log.go
HomayoonAlimohammadi: The reason behind this variadic arg is to enable users to do:
```go
ctx := types.WithLogger(context.Background())
// instead of
ctx := types.WithLogger(context.Background(), slog.Default())
```
to set a default logger to the context.
and they're also able to do the following to set a custom logger:
```go
ctx := types.WithLogger(context.Background(), myCustomLogger)
```
I've seen this pattern in multiple places and found it interesting. If you don't like it I can change it.

roosterfish:
Can you please rebase, there are now conflicts after merging in other PRs.

microcluster/types/log.go
roosterfish: My proposal was to only move the `CtxLogger` const, not the entire file. It looks you have now moved everything.

roosterfish: If we just move `CtxLogger` const, we can still reference it from anywhere inside `internal/` but you can also make use of it when you try to derive a new context including a logger.

roosterfish (changes requested):

HomayoonAlimohammadi:
@roosterfish Thank you for the review. I've updated the PR with your instructions. Please let me know if it needs further polishing.

internal/log/logger.go
```diff
@@ -4,12 +4,11 @@ import (
...
 	"errors"
 	"log/slog"
-)
 
-type ctxKey string
+	"github.com/canonical/microcluster/v3/microcluster/types/log"
+)
 
-// CtxLogger is the name of the context value for the central logger.
-const CtxLogger ctxKey = "logger"
```
roosterfish, line 12: Let's move this `const` (and type `ctxKey`) entirely into `microcluster/types`. There is no need to define another `const` here once it's moved.

roosterfish: Maybe my comment [here](https://github.com/canonical/microcluster/pull/603/changes/dde6ba3d8f87b6d12be5058b5e39fb8af7d9c494#r2757872580) was not clear.

microcluster/types/log/log.go
roosterfish: I see thanks. 
Let's call it `ContextWithLogger` to be more descriptive. This would also align with the `LoggerFromContext` funcs.

microcluster/app.go
```diff
@@ -22,6 +22,7 @@ import (
 	internalClient "github.com/canonical/microcluster/v3/internal/rest/client"
 	"github.com/canonical/microcluster/v3/internal/sys"
 	"github.com/canonical/microcluster/v3/microcluster/types"
+	mclog "github.com/canonical/microcluster/v3/microcluster/types/log"
```
roosterfish: We tend to use `micro` prefix if there is a collision. I would suggest `microLog`.

internal/daemon/daemon.go
```diff
@@ -153,7 +153,13 @@ func NewDaemon() *Daemon {
 // log is a convenience to retrieve the internal logger from the shutdown context.
 // We always expect the logger to be present.
 func (d *Daemon) log() *slog.Logger {
-	return d.shutdownCtx.Value(internalLog.CtxLogger).(*slog.Logger) //nolint:revive
+	l, err := internalLog.LoggerFromContext(d.shutdownCtx)
```
roosterfish: I am a bit surprised to see those changes.

At this stage there already is a logger set on the `d.shutdownCtx`. See the comment on the func.

roosterfish (changes requested):
The current logger changes feel too invasive to me.

In one of the [recent comments](https://github.com/canonical/microcluster/pull/603#issuecomment-3835347860) I mentioned the approach we take for logging.

I understand you want to have an easy way to derive a context with the logger set to mimic what is done internally by Microcluster so you can test e.g. API handlers as standalone (as we do in our unit tests).
The easiest (which already works today) is to derive a context on your own using `context.WithValue(context.TODO(), log.CtxLogger, slog.Default())`.
But together we identified that `log.CtxLogger` is in the `internal/` package. To overcome this limitation my suggestion was to move this constant into `microcluster/types/log` instead so you can access it too. It can even be a just a file like `log.go` (no need for an extra `log/` package).

But this should not require this amount of changes. Only the places where we currently use `log.CtxLogger` need to be updated to point to the `micrcluster/types` package which would then contain the constant.

roosterfish:
If you intend to have a little helper func, you can keep the `ContextWithLogger` func and we can then use it in the three places where we run `context.WithValue`.

internal/log/logger.go
HomayoonAlimohammadi: Done

microcluster/app.go
HomayoonAlimohammadi: Done

internal/daemon/daemon.go
HomayoonAlimohammadi: Removed these changes completely.

microcluster/types/log/log.go
```diff
@@ -0,0 +1,24 @@
+package log
```
roosterfish: Thanks, only one more smaller nit, please move this file into `microcluster/types/log.go` as mentioned [here](https://github.com/canonical/microcluster/pull/603#pullrequestreview-3756955539). We don't need an extra `log` package for this const and func.

roosterfish (changes requested):

microcluster/types/log/log.go
HomayoonAlimohammadi: Done

microcluster/types/log.go
```diff
@@ -0,0 +1,24 @@
...
+
+// ContextWithLogger returns a new context with the given logger.
+// If no logger is provided, the default logger is used.
+func ContextWithLogger(ctx context.Context, logger ...*slog.Logger) context.Context {
+	var l *slog.Logger
+	if len(logger) == 0 {
+		l = slog.Default()
+	} else {
+		l = logger[0]
+	}
```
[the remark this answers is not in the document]
roosterfish: This is fine to keep the logger optional.

microcluster/rest/response/response.go
```diff
@@ -3,6 +3,7 @@ package response
 import (
 	"bytes"
```
roosterfish, line 4: Ah I just wanted to approve and merge but saw the commit messages don't comply to our regular format.

Please append the package names as prefix, see https://github.com/canonical/microcluster/pull/595 for an example.

roosterfish (changes requested):

microcluster/rest/response/response.go
HomayoonAlimohammadi: Thanks for reminding me. I've updated the commit messages. Do they comply with the format?

roosterfish (approved):
Thanks!

PR #607, opened 2026-02-03
Restructure (7): Make extensions package public

roosterfish:
This is the seventh PR taken from a single commit in https://github.com/canonical/microcluster/pull/534 (even though in this case the actual changes are very different from the original commit).

In this PR the extensions package is made public so that the actual types can be accessed by importers of Microcluster.

### DIFF
--- a/internal/cluster/cluster_members.go
+++ b/internal/cluster/cluster_members.go
@@ -7,7 +7,6 @@ import (
 	"time"
 
 	"github.com/canonical/microcluster/v3/internal/db/update"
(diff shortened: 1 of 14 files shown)


tugbataluy (approved):
LGTM! 👍
The migration of the Extensions type from internal/ to the public microcluster/types package is clean and follows the same pattern established in the previous restructure PRs. As far as I understand from the previous PRs, this continues making internal types properly accessible to downstream consumers while enabling mocking for unit tests.

PR #608, opened 2026-02-03
Restructure (8): Make DB interface public 

roosterfish:
Can only be merged after (won't build before):
* https://github.com/canonical/microcluster/pull/607

This is the eigth PR taken from a single commit in https://github.com/canonical/microcluster/pull/534 (even though in this case the actual changes are very different from the original commit).

In this PR the DB interface is made public. It's returned by the public `State` interface and should therefore not be located inside the `internal/` package.

### DIFF
--- a/internal/state/state.go
+++ b/internal/state/state.go
@@ -45,7 +45,7 @@ type State interface {
 	ClusterCert() *shared.CertInfo
 
 	// Database.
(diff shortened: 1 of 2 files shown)


tugbataluy (approved):
LGTM, thanks ❇️

PR #609, opened 2026-02-04
internal/state: add ctx deadline exceeded check on membership consistency check

louiseschmidtgen:
# Add ctx deadline exceeded check on membership consistency check

This PR adds a ctx deadline exceeded check in the membership consistency check loop on the v2 branch. 

Nice to know: In v3 we pass the request's context, however, we do not want to make any drastic changes to the LTS branch.

### DIFF
--- a/internal/state/state.go
+++ b/internal/state/state.go
@@ -258,7 +258,7 @@ func (s *InternalState) CheckMembershipConsistency(ctx context.Context) error {
 		if err != nil {
 			select {
 			case <-ctx.Done():
-				return fmt.Errorf("Membership consistency check failed after timeout: %w", err)
+				return fmt.Errorf("Membership consistency check failed: %w", err)
 			case <-time.After(200 * time.Millisecond):
 				continue
 			}


internal/state/state.go
```diff
@@ -249,6 +249,11 @@ func (s *InternalState) CheckMembershipConsistency(ctx context.Context) error {
 	}
 
 	for {
+		ctxErr := ctx.Err()
+		if ctxErr != nil {
+			return fmt.Errorf("Membership consistency check failed after timeout: %w", ctxErr)
```
roosterfish: Just a nit, it might be that the provided context was just cancelled and didn't ran into a timeout.
So the error message should be generic, the `%w` will include the well known "deadline exceeded" as a reason.
You can just do `Membership consistency check failed: %w`.

roosterfish: The same applies for the error below that checks `<-ctx.Done()`.

internal/state/state.go
```diff
@@ -249,6 +249,11 @@ func (s *InternalState) CheckMembershipConsistency(ctx context.Context) error {
 	}
 
 	for {
+		ctxErr := ctx.Err()
```
roosterfish: The more I think about this, I am less confident that [this copilot comment](https://github.com/canonical/microcluster/pull/602#discussion_r2737012406) is actually useful.

If we call `getMembershipData` with an already expired context, we still get `Failed to gather membership data for consistency check: context deadline exceeded` which is explanatory enough IMHO.

So it's not that we loose anything around the fact we are trying to perform the consistency check. It's even written in the error message.

Let's just fix https://github.com/canonical/microcluster/pull/609/changes/e76bcffc95e532daab8f8fbdf390570a5dca3a19#r2762736434 and we should be good to go. Fine for you?

roosterfish (changes requested):

internal/state/state.go
louiseschmidtgen: Thanks for the catch in duplication.

internal/state/state.go
roosterfish: I was proposing to again drop this `ctxErr := ctx.Err() ... ` section as it looks to me the copilot comment wasn't that useful as we would still have a nice and descriptive error message returned by `getMembershipData`.

roosterfish (changes requested):

roosterfish (approved):
Ty!

PR #610, opened 2026-02-05
Restructure (9): Make `State` interface and `Hooks` public

roosterfish:
This is the ninth PR taken from a single commit in https://github.com/canonical/microcluster/pull/534 (even though in this case the actual changes are very different from the original commit).

In this PR the `State` interface can finally be moved to a public package together with the `Hooks` struct which allows removing the import indirection which was previously necessary to be able to access both the `State` and `Hooks`.


tugbataluy (approved):
LGTM, thanks ❇️ . This PR successfully restructures the codebase to expose the State and Hooks types properly through the public types package (like the previous PRs). As far as I understand, all imports and references have been updated correctly.

markylaing (approved):
Very nice 🥳

PR #620, opened 2026-02-18
golangci: Fix issues with latest linter

roosterfish:
Do the same as in LXD and skip the check.
See https://github.com/canonical/lxd/pull/17623


roosterfish:
This also requires a backport to v2.

tugbataluy (approved):
LGTM, thanks ❇️ !

PR #621, opened 2026-02-18
Restructure (10): Merge `rest` package into `types`

roosterfish:
This is the tenth PR taken from a single commit in https://github.com/canonical/microcluster/pull/534 (even though in this case the actual changes are very different from the original commit).

In this PR the `rest` package is merged into `types` which now makes all the public types available through `microcluster/types`. 
A bit of refactor is required as it also affects moving the used package/type for all the internal API endpoint/handler related funcs.


tugbataluy (approved):
LGTM, thanks ❇️. As far as I understand, this PR successfully merges the `rest` package into `types`, making all public types available through a single `microcluster/types` package. The renaming of Server to Status (for cluster member status) is a sensible clarification, and all internal references across `rest/resources`,` daemon`, `recover`, `microcluster`, and the `example` package have been updated correctly 💯 .

PR #623, opened 2026-02-23
Restructure (11): Reorganize endpoints

roosterfish:
Closes https://github.com/canonical/microcluster/pull/534.

This is the **eleventh and last** PR taken from a single commit in https://github.com/canonical/microcluster/pull/534 (even though in this case the actual changes are very different from the original commit).

This PR mainly adds context around the various endpoints about there presence during and before init and moves some of the endpoints to the Unix socket for security reasons (e.g. SQL endpoints).

Furthermore I found a regression which was introduced by picking the wrong context. Fixed in the last commit.

### DIFF
--- a/internal/rest/resources/control.go
+++ b/internal/rest/resources/control.go
@@ -129,7 +129,8 @@ func controlPost(state types.State, r *http.Request) types.Response {
 			<-r.Context().Done()
 
 			// Use `force=1` to ensure the node is fully removed, in case its listener hasn't been set up.
-			err = internalClient.DeleteClusterMember(r.Context(), client, req.Name, "", true)
+			// Use the background context as the original request context is already cancelled at this point.
+			err = internalClient.DeleteClusterMember(context.Background(), client, req.Name, "", true)
 			if err != nil {
 				logger.Error("Failed to clean up cluster state after join failure", slog.String("error", err.Error()))
 			}
--- a/internal/rest/resources/control.go
+++ b/internal/rest/resources/control.go
@@ -23,6 +23,7 @@ import (
 )
 
 var controlCmd = types.Endpoint{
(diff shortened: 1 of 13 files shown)


markylaing:
General LGTM from me when the tests pass :)

roosterfish:
> General LGTM from me when the tests pass :)

Thanks for taking a look already, I'll follow up on the test failure now, just wanted to see if it happens intermittently or always.

roosterfish:
@markylaing issue resolved, I was too ambitious removing endpoints from being available before init (cluster bootstrap). Added another comment on the `truststore/` endpoint for clarification.

roosterfish:
Also tested `v2` to `v3 (this version)` upgrade to check for regressions during cluster upgrade.

internal/rest/resources/control.go
```diff
@@ -129,7 +129,7 @@ func controlPost(state types.State, r *http.Request) types.Response {
 			<-r.Context().Done()
 
 			// Use `force=1` to ensure the node is fully removed, in case its listener hasn't been set up.
-			err = internalClient.DeleteClusterMember(r.Context(), client, req.Name, "", true)
+			err = internalClient.DeleteClusterMember(context.Background(), client, req.Name, "", true)
```
markylaing, line 133: Can all the reasoning in the commit description please be added as a comment? Otherwise at some point we'll see a background context and wonder why the request context is not used.

markylaing (changes requested):
Small nit about comments :)

internal/rest/resources/control.go
roosterfish: Fixed.

markylaing (approved):
Great work 🥇

PR #629, opened 2026-02-23
Add failure domain support and daemon configuration APIs (failure domains part 1/2)

kadinsayani:
This adds failure domain configuration to Microcluster per https://docs.google.com/document/d/1Sy2mbkxQSUgMMyI9lE9oT1U9ln77ApYny0XVxmQ9jXs/edit?usp=sharing. A failure-domain field (`uint64`) is added to `daemon.yaml` and supplied to go-dqlite at startup via `WithFailureDomain`.

A new GET/PUT /1.0/daemon/config endpoint exposes the full daemon configuration. PUT allows updating servers and failure-domain; name and address are validated as immutable. Changes are persisted to `daemon.yaml` and take effect on next daemon start. The legacy /1.0/daemon/servers endpoint remains available.

Includes unit tests (config layer, API handlers, validation), an integration test, and documentation updates.

Part 2 will add support for immediate mode via a `restart` flag on the config update, allowing the failure domain to take effect without a manual daemon restart.

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -76,3 +76,81 @@ dargs := microcluster.DaemonArgs{
     },
 }
 ```
+
+### Daemon configuration API
+
+The local daemon configuration is exposed via:
+
+* `GET /core/1.0/daemon/config`
+* `PUT /core/1.0/daemon/config`
+* `PATCH /core/1.0/daemon/config`
+
+`GET` returns `types.DaemonConfig`:
+
+```json
+{
+  "name": "m1",
+  "address": "127.0.0.1:9001",
+  "servers": {
+    "metrics.example.com": {
+      "address": "127.0.0.1:9443"
+    }
+  },
+  "failure-domain": 1
+}
+```
+
+`PUT` performs a full replacement of all mutable fields:
+
+* `servers` — replaces the servers map entirely. Omitting or setting `null` clears all servers.
+* `failure-domain` — sets a new value. Omitting or setting `null` clears the failure domain.
+
+`name` and `address` are immutable. If provided they must match the current values; if omitted they are ignored.
+
+`PATCH` supports **partial updates** using the `DaemonConfigPatch` request body:
+
+* `servers` — omit the field to leave existing server configuration unchanged; include it to replace the servers map entirely (send `{}` to clear all servers).
+* `failure-domain` — omit the field to leave the existing value unchanged; include it to set a new value (send `0` to clear the failure domain).
+
+`name` and `address` are not part of the `PATCH` request body and are ignored if present in the payload.
+
+To update only `failure-domain` without touching servers (PATCH):
+
+```json
+{
+  "failure-domain": 2
+}
+```
+
+To update only `servers` without touching `failure-domain` (PATCH):
+
+```json
+{
+  "servers": {
+    "metrics.example.com": {
+      "address": "127.0.0.1:9443"
+    }
+  }
+}
+```
+
+To clear all servers explicitly, send an empty object for the field:
+
+```json
+{
+  "servers": {}
+}
+```
+
+To clear `failure-domain`, send `0`:
+
+```json
+{
+  "failure-domain": 0
+}
+```
+
+`servers` updates are applied immediately to listener configuration.
+`failure-domain` changes are persisted immediately but only applied to dqlite on the next daemon start.
+
+The legacy `PUT /core/1.0/daemon/servers` endpoint remains available.
--- a/internal/config/daemon_test.go
+++ b/internal/config/daemon_test.go
@@ -1,0 +1,60 @@
+package config
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/canonical/microcluster/v4/microcluster/types"
+)
+
+func TestDaemonConfigFailureDomainDefaultsToZeroWhenUnset(t *testing.T) {
+	t.Parallel()
+
+	path := filepath.Join(t.TempDir(), "daemon.yaml")
+
+	content := `name: m1
+address: 127.0.0.1:9001
+servers: {}
+`
+
+	err := os.WriteFile(path, []byte(content), 0600)
+	require.NoError(t, err)
+
+	cfg := NewDaemonConfig(path)
+
+	err = cfg.Load()
+	require.NoError(t, err)
+
+	require.Equal(t, uint64(0), cfg.GetFailureDomain())
+}
+
+func TestDaemonConfigFailureDomainRoundTrip(t *testing.T) {
+	t.Parallel()
+
+	path := filepath.Join(t.TempDir(), "daemon.yaml")
+
+	cfg := NewDaemonConfig(path)
+	addr, err := types.ParseAddrPort("127.0.0.1:9001")
+	require.NoError(t, err)
+
+	cfg.SetName("m1")
+	cfg.SetAddress(addr)
+	cfg.SetServers(map[string]types.ServerConfig{})
+	cfg.SetFailureDomain(42)
+
+	err = cfg.Write()
+	require.NoError(t, err)
+
+	content, err := os.ReadFile(path)
+	require.NoError(t, err)
+	require.True(t, strings.Contains(string(content), "failure-domain: 42"))
+
+	reloaded := NewDaemonConfig(path)
+	err = reloaded.Load()
+	require.NoError(t, err)
+	require.Equal(t, uint64(42), reloaded.GetFailureDomain())
+}
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -288,7 +288,7 @@ func (d *Daemon) init(listenAddress string, socketGroup string, heartbeatInterva
 		return fmt.Errorf("Failed to initialize trust store: %w", err)
 	}
 
-	d.db, err = db.NewDB(d.shutdownCtx, d.ServerCert, d.ClusterCert, d.Name, d.os, heartbeatInterval)
+	d.db, err = db.NewDB(d.shutdownCtx, d.ServerCert, d.ClusterCert, d.Name, d.config.GetFailureDomain, d.os, heartbeatInterval)
 	if err != nil {
 		return fmt.Errorf("Failed to initialize database: %w", err)
 	}
--- a/internal/rest/client/daemon.go
+++ b/internal/rest/client/daemon.go
@@ -9,6 +9,39 @@ import (
 	"github.com/canonical/microcluster/v4/microcluster/types"
 )
 
+// GetDaemonConfig retrieves local daemon configuration.
+func GetDaemonConfig(ctx context.Context, c types.Client) (*types.DaemonConfig, error) {
+	queryCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	defer cancel()
+
+	endpoint := api.NewURL().Path("daemon", "config")
+	config := types.DaemonConfig{}
+	err := c.Query(queryCtx, "GET", types.PublicEndpoint, &endpoint.URL, nil, &config)
+	if err != nil {
+		return nil, err
+	}
+
+	return &config, nil
+}
+
+// UpdateDaemonConfig updates local daemon configuration.
+func UpdateDaemonConfig(ctx context.Context, c types.Client, config types.DaemonConfig) error {
+	queryCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	defer cancel()
+
+	endpoint := api.NewURL().Path("daemon", "config")
+	return c.Query(queryCtx, "PUT", types.PublicEndpoint, &endpoint.URL, config, nil)
+}
+
+// PatchDaemonConfig partially updates local daemon configuration.
+func PatchDaemonConfig(ctx context.Context, c types.Client, config types.DaemonConfigPatch) error {
+	queryCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	defer cancel()
+
+	endpoint := api.NewURL().Path("daemon", "config")
+	return c.Query(queryCtx, "PATCH", types.PublicEndpoint, &endpoint.URL, config, nil)
+}
+
 // UpdateServers updates the additional servers config.
 func UpdateServers(ctx context.Context, c types.Client, config map[string]types.ServerConfig) error {
 	queryCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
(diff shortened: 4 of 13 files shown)


microcluster/types/config.go
```diff
@@ -2,8 +2,8 @@ package types
 
 // DaemonConfig is the in memory version of the local daemon.yaml file.
 type DaemonConfig struct {
-	Name          string                  `json:"name" yaml:"name"`
-	Address       AddrPort                `json:"address" yaml:"address"`
-	Servers       map[string]ServerConfig `json:"servers" yaml:"servers"`
-	FailureDomain *uint64                 `json:"failure-domain,omitempty" yaml:"failure-domain,omitempty"`
+	Name          string                   `json:"name" yaml:"name"`
+	Address       AddrPort                 `json:"address" yaml:"address"`
+	Servers       *map[string]ServerConfig `json:"servers,omitempty" yaml:"servers,omitempty"`
```
roosterfish, line 14: Should we just land this change before the LTS release? Then we don't have to rush the failure domains in before Friday.

roosterfish (commented):
Thanks for this. I was quickly browsing the changes to spot things that might break the LTS if we slide it in afterwards.

Will check it more thoroughly later.

microcluster/types/config.go
kadinsayani: Yeah we can land this one before. I'll extract the relevant bits to a separate PR. We can include a few smaller refactorings as well.

kadinsayani: https://github.com/canonical/microcluster/pull/631

roosterfish: Thanks I think we can go without the pointer change if I am not mistaken as `map` can already represent `nil` state. Pls see my comment on the other PR.

roosterfish:
Can be rebased.

kadinsayani:
@roosterfish ready for your review :)
Working on https://github.com/canonical/microcluster/pull/635 now.

internal/config/daemon_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+	err := os.WriteFile(path, []byte(content), 0600)
+	require.NoError(t, err)
+
+	cfg := NewDaemonConfig(path)
+
+	err = cfg.Load()
+	require.NoError(t, err)
+
+	require.Equal(t, uint64(0), cfg.GetFailureDomain())
```
roosterfish, line 32: If you start dqlite without `WithFailureDomain`, does it set the failure domain to 0 or nil? 
I am asking because in the first commit you have added support to track a nil failure domain using `*uint64` so I am wondering if this line here shouldn't rather return nil instead of 0 as failure domain isn't set in the `content`.

Btw is 0 a valide failure domain or does dqlite consider it to be set only if `!= 0`?

internal/rest/resources/daemon.go
```diff
@@ -20,6 +20,30 @@ var daemonServersCmd = types.Endpoint{
 	Put: types.EndpointAction{Handler: daemonServersPut, AccessHandler: access.AllowAuthenticated},
 }
 
+var daemonConfigCmd = types.Endpoint{
+	Path: "daemon/config",
```
roosterfish, line 24: We might want to allow this endpoint to be present before init already. What do you think? 
It could be useful to have it available even before you bootstrap/join a cluster to validate some configuration.
Just a thought.

internal/rest/resources/daemon.go
```diff
@@ -20,6 +20,30 @@ var daemonServersCmd = types.Endpoint{
...
+		return types.SmartError(err)
+	}
+
+	daemonConfig := intState.LocalConfig()
+	fd := daemonConfig.GetFailureDomain()
+	config := types.DaemonConfig{
+		Name:          daemonConfig.GetName(),
+		Address:       daemonConfig.GetAddress(),
+		Servers:       daemonConfig.GetServers(),
+		FailureDomain: &fd,
```
roosterfish: Can we use the `Dump` func?

internal/rest/resources/daemon.go
roosterfish, line 21: > they are rejected only when explicitly
provided with a different value. If omitted entirely they are ignored.

That sounds reasonable, like it.

internal/daemon/daemon.go
```diff
@@ -203,6 +205,11 @@ func (d *Daemon) Run(ctx context.Context, stateDir string, args Args) error {
 		return fmt.Errorf("Database recovery failed: %w", err)
 	}
 
+	err = d.loadConfigIfExists()
```
roosterfish: > setConfig is called on bootstrap and join to write the node's name and
address into daemon.yaml. Previously it did so on a blank in-memory
config, silently discarding any servers or failure-domain that had
already been persisted to disk by a prior PUT /daemon/config call.

The `PUT daemon/servers` endpoint isn't available prior to init (bootstrap/join). Is this still true?

internal/rest/client/daemon.go
```diff
@@ -33,6 +33,15 @@ func UpdateDaemonConfig(ctx context.Context, c types.Client, config types.Daemon
 	return c.Query(queryCtx, "PUT", types.PublicEndpoint, &endpoint.URL, config, nil)
 }
 
+// PatchDaemonConfig partially updates local daemon configuration.
```
roosterfish, line 36: Just a nit, the previous client funcs for GET/PUT where added after the commit that added the actual API endpoint.

internal/rest/resources/daemon_test.go
```diff
@@ -0,0 +1,684 @@
...
+			})
+			cfg.SetFailureDomain(3)
+
+			updateCalls := 0
+			s := &internalState.InternalState{
+				LocalConfig: func() *internalConfig.DaemonConfig {
+					return cfg
+				},
+				UpdateServers: func() error {
+					updateCalls++
```
roosterfish, line 222: Cool idea of instrumenting the `UpdateServers`!

roosterfish (changes requested):
This is looking great already, especially the detailed additions to the test suite @kadinsayani.

Mainly a few remarks around what is a nil failure domain and how does dqlite handle the default case or not setting a failure domain.

internal/config/daemon_test.go
kadinsayani: It would end up as 0. 0 is a valid failure domain and is the default when none is set.

internal/rest/resources/daemon.go
kadinsayani: `Dump` returns the internal `*types.DaemonConfig` directly. We don't want to expose mutable shared state here.

internal/rest/client/daemon.go
kadinsayani: Ack, I'll clean up the commits up a bit for better organization.

internal/daemon/daemon.go
kadinsayani: Yes, still true. The overwrite risk is still valid for cases where daemon.yaml exists from a prior run (restart, rejoin or recover).

roosterfish:
Please change the target branch to `v4` as this is the new default after we have released `v3`.

kadinsayani:
> Please change the target branch to `v4` as this is the new default after we have released `v3`.

Done. I've also cleaned up the commit organization and reworked the types. We now have separate structs for `DaemonConfigPut` and `DaemonConfigPatch`. I think this is cleaner and makes the usage more explicit. Let me know what you think :)

internal/rest/resources/daemon.go
kadinsayani: It may be useful to allow `GET /daemon/config` before init.

internal/rest/resources/daemon.go
roosterfish: Should we modify `Dump` to instead return a copy of the config? Then we can use it consistently across the various `daemon/*` endpoints.

internal/daemon/daemon.go
roosterfish: Maybe I don't follow. Currently `SetConfig` is run inside the `controlPost` handler which gets triggered during bootstrap or join and writes the first version of the config (including name and address).
Prior to this point the extension servers endpoint isn't present so you cannot configure the `servers` config key through the API.
Manual modification of the config file isn't something we ever advertised to be done by a user directly.

So when we start making the `daemon/config` endpoints available before init (bootstrap/join), then I see why we would need this change to not accidentally overwrite changes applied to the config (through the `daemon/config` endpoints in this PR).

roosterfish: There already is a `.Load` func on the `*Daemon`'s config struct. Can we use this one instead? Currently it's called as part of `(d *Daemon) reload()` but maybe we can move it up in the call chain for the purpose you are addressing?

microcluster/types/config.go
```diff
@@ -1,8 +1,23 @@
...
+}
+
 // DaemonConfig is the in memory version of the local daemon.yaml file.
 type DaemonConfig struct {
-	Name    string                  `json:"name" yaml:"name"`
-	Address AddrPort                `json:"address" yaml:"address"`
-	Servers map[string]ServerConfig `json:"servers" yaml:"servers"`
+	Name            string   `json:"name" yaml:"name"`
+	Address         AddrPort `json:"address" yaml:"address"`
+	DaemonConfigPut `yaml:",inline"`
```
roosterfish: We cannot break the API structs to allow the failure domain additions being backported to v3. I liked how the struct was looking before as it was only adding `FailureDomain` which isn't a breaking change which is easily backportable.

roosterfish (changes requested):

microcluster/types/config.go
kadinsayani: Ack, I'll revert back to the previous state.

internal/rest/resources/daemon.go
kadinsayani: Sure, sounds good.

internal/rest/resources/daemon.go
kadinsayani: Would we also enable `PUT/PATCH /daemon/config` before init? This would require applying local changes first, then notifying other members post-init.

roosterfish: Mhh good point regarding the notification!

Let's not do it for now, we can always add this later in case it's needed.

kadinsayani: Sounds good.

internal/daemon/daemon.go
kadinsayani: We can drop this logic since we're not making the `daemon/config` endpoints available before init.

microcluster/types/config.go
kadinsayani: We still need to add a `DaemonConfigPatch` struct for proper partial update support.

kadinsayani: Shall we just exclude PATCH and partial update support for now?

roosterfish: > We still need to add a DaemonConfigPatch struct for proper partial update support.

Is it due to the `Servers` and the pointer we talked about in https://github.com/canonical/microcluster/pull/631#discussion_r2854184598?

Yes let's add the `DaemonConfigPatch` as it's a new type but leave the original `DaemonConfig` as is.

microcluster/app.go
```diff
@@ -469,6 +469,36 @@ func (m *MicroCluster) LoggerFromContext(ctx context.Context) *slog.Logger {
...
+	c, err := m.LocalClient()
+	if err != nil {
+		return err
+	}
+
+	return internalClient.UpdateDaemonConfig(ctx, c, config)
+}
+
+// PatchDaemonConfig partially updates local daemon configuration.
+func (m *MicroCluster) PatchDaemonConfig(ctx context.Context, config types.DaemonConfigPatch) error {
```
roosterfish: We should be opinionated here and offer the user only one option to patch the daemon config through the app to keep it simple.
As the `/core/1.0/*` endpoints are kept stable, users can setup their own client funcs to use another endpoint.

Maybe let's keep the `UpdateDaemonConfig` but use the PATCH under the hood? I have the feeling this is more intuitive and less error prone as you just update what you intend to update.
And then drop `PatchDaemonConfig` from the app.

roosterfish (changes requested):
Thanks Kadin. I think once the `DaemonConfig` is restored to be non-breaking to allow backports the PR should be ready.

microcluster/types/config.go
kadinsayani: Correct, sounds good :)

microcluster/app.go
kadinsayani: Sounds good, done.

kadinsayani:
> Thanks Kadin. I think once the `DaemonConfig` is restored to be non-breaking to allow backports the PR should be ready.

Thanks for your reviews on this one :)
Should be good to go now.

microcluster/app.go
```diff
@@ -469,6 +469,27 @@ func (m *MicroCluster) LoggerFromContext(ctx context.Context) *slog.Logger {
...
+	if err != nil {
+		return nil, err
+	}
+
+	return internalClient.GetDaemonConfig(ctx, c)
+}
+
+// UpdateDaemonConfig partially updates local daemon configuration.
+// Only fields explicitly set (non-nil) in the patch are applied; omitted fields preserve their current values.
+func (m *MicroCluster) UpdateDaemonConfig(ctx context.Context, config types.DaemonConfigPatch) error {
```
roosterfish, line 484: Let's also deprecate `(m *MicroCluster) UpdateServers` by adding a comment on the func as this is now superseded by `UpdateDaemonConfig`.

doc/api.md
```diff
@@ -76,3 +76,81 @@ dargs := microcluster.DaemonArgs{
...
+{
+  "servers": {}
+}
+```
+
+To clear `failure-domain`, send `0`:
+
+```json
+{
+  "failure-domain": 0
```
roosterfish, line 149: Worth mentioning that 0 is the default if not set https://github.com/canonical/microcluster/pull/629#discussion_r2866092040?

roosterfish (approved):
Looks good, I'll merge it now. Please check the two other comments but you can address this in a follow up PR.

I understand https://github.com/canonical/microcluster/pull/635 will have the restart capabilities. 
Please also make sure that we can pass the failure domain already when bringing up the daemon using the [`DaemonArgs`](https://github.com/canonical/microcluster/blob/v4/microcluster/app.go#L82) so you can prevent a restart if you know the failure domain upfront.

doc/api.md
kadinsayani: Yeah good idea thanks

microcluster/app.go
kadinsayani: https://github.com/canonical/microcluster/pull/635/changes/d4717991ac41e240b650e4f85b538e2c47f91706

doc/api.md
kadinsayani: https://github.com/canonical/microcluster/pull/635/changes/a8b3cbae64c3aee893d7c2e00d7dadc445c99086

PR #630, opened 2026-02-24
go: Update deps

roosterfish:


roosterfish:
@MusicDin as part of updating the deps I have removed the static pin on LXD 5.21 which we should anyway not continue to use for `v3` Microcluster.

This should remove the go issue on `shared.RunCommand`.

kadinsayani (approved):

MusicDin (approved):
Thanks :rocket:

PR #631, opened 2026-02-24
Extract functions for use in daemon/config handlers

kadinsayani:


microcluster/types/config.go
```diff
@@ -2,7 +2,7 @@ package types
 
 // DaemonConfig is the in memory version of the local daemon.yaml file.
 type DaemonConfig struct {
-	Name    string                  `json:"name" yaml:"name"`
-	Address AddrPort                `json:"address" yaml:"address"`
-	Servers map[string]ServerConfig `json:"servers" yaml:"servers"`
+	Name    string                   `json:"name" yaml:"name"`
+	Address AddrPort                 `json:"address" yaml:"address"`
+	Servers *map[string]ServerConfig `json:"servers,omitempty" yaml:"servers,omitempty"`
```
roosterfish: Mh, as `map` can be nil, wouldn't that be sufficient already combined with `omitempty`?

If the request payload is `{}`, then `Servers` will be nil. If the payload is `{"servers":{}}`, then `Servers` won't be nil but an empty map.

kadinsayani: I was thinking that `map[]` + `omitempty` would omit both `nil` and empty maps, and `*map[]` + `omitempty` would only omit `nil` so empty map marshals as `servers: {}`. If the only requirement was parsing the request, a plain map would work, but we also want to marshal the constructed `types.DaemonConfig` in the new `applyAndNotifyDaemonConfig()` helper.

internal/rest/resources/daemon.go
```diff
@@ -104,11 +103,29 @@ func daemonServersPut(s types.State, r *http.Request) types.Response {
 			return fmt.Errorf("No remote found at address %q to run the %q hook", c.URL().Host, types.OnDaemonConfigUpdate)
 		}
 
-		return internalClient.RunOnDaemonConfigUpdateHook(ctx, c.UseTarget(remote.Name), daemonConfig.Dump())
```
roosterfish, line 107: Any reason for not anymore using the `Dump`? Do we need to remove it too or is it used somewhere else?

internal/rest/resources/daemon.go
```diff
@@ -104,11 +103,29 @@ func daemonServersPut(s types.State, r *http.Request) types.Response {
...
 	})
-	if err != nil {
-		return types.SmartError(err)
+}
+
+// validateServerConfigs checks that each server name has a matching additional
+// listener and that no two servers share the same address.
+func validateServerConfigs(s *internalState.InternalState, servers map[string]types.ServerConfig) error {
+	for serverName := range servers {
+		if !slices.Contains(s.ExtensionServers(), serverName) {
```
roosterfish: Worth assigning `s.ExtensionServers()` to a var before running the loop to not require running it in every iteration? Good call removing the `found` lookup approach.

microcluster/types/config.go
roosterfish: I see thanks, was only thinking about the parsing not marshalling. 

However in this case I think we can just go with this:

```go
Servers map[string]ServerConfig `json:"servers" yaml:"servers"`
```

1) Parsing `{}` results in a `nil` map for `Servers`. Marshalling it will return `{"servers":null}`
2) Parsing `{"servers":{}}` results in an empty map. Marshalling it will return `{"servers":{}}`
3) Parsing `{"servers":{"foo":{}}}` will result in the same when marshalled.

Do I miss a case? This way when you use the PUT endpoint you can do `{"servers":null}`, `{"servers":{}}` or `{}` to reset the servers config entirely. 

And when requesting the GET we will either return `{"servers":{}}` or `{"servers":{"foo":{}}}` depending on whether or not servers config is set in the daemon config.

roosterfish: I think this is a new for Microcluster API (I am not aware of another endpoint that would allow such operation) so we can be free in defining an approach and sticking to it going forward.

roosterfish (changes requested):

microcluster/types/config.go
kadinsayani: With `Servers map[...]...` both `{}` and `{"servers" null}` unmarshal to `nil` for `Servers`, so we can't differentiate between "field omitted" and "explicit null".

The main driver for using a pointer is so PUT requests to `/daemon/config` can apply partial updates. For example, you could just update the failure domain and not provide `servers` without the risk of clearing existing servers config.

kadinsayani: > I think this is a new for Microcluster API (I am not aware of another endpoint that would allow such operation) so we can be free in defining an approach and sticking to it going forward.

We can either a) allow partial updates via PUT /daemon/config or b) require providing the full daemon config with each request.

internal/rest/resources/daemon.go
kadinsayani: `Dump()` returns the the internal `*types.DaemonConfig` directly without any locking or copying, so using it for outbound requests can leak mutable state.

kadinsayani: We can drop it.

internal/rest/resources/daemon.go
kadinsayani: Yeah good idea, thanks!

microcluster/types/config.go
roosterfish: > The main driver for using a pointer is so PUT requests to /daemon/config can apply partial updates. For example, you could just update the failure domain and not provide servers without the risk of clearing existing servers config.

Shouldn't we use PATCH for partial updates? Because then in the PUT case it doesn't matter if you do `"servers":null` or `"servers:{}"` or `{}` as it will have the same effect and remove any server config.

> We can either a) allow partial updates via PUT /daemon/config or b) require providing the full daemon config with each request.

Right now you can only set the servers config so both PUT/PATCH would provide the same functionality. 

I wonder if we instead create a PATCH /daemon/config, would it make it easier? Because with the PATCH you can then modify specific config items, and the PUT wouldn't require extra handling around whether or not something is explicitly or implicitly set to nil.

kadinsayani: Sounds good, let's include both PATCH and PUT for /daemon/config.

internal/rest/resources/daemon.go
kadinsayani: Actually, we are fine to keep and use `Dump()` here.

kadinsayani:
@roosterfish I've reworked this PR to just extract the two helper functions and fix the error message typo.

roosterfish (approved):
Thanks!

PR #632, opened 2026-02-25
Reset: Fix race in cluster member self deletion

roosterfish:
This PR makes the self reset more bulletproof by ensuring it doesn't race with the actual client request. 
The issue surfaced after implementing #520 which started to derive the API handler client contexts from the daemon's shutdown context.

Thanks @HomayoonAlimohammadi for raising this.

### DIFF
--- a/example/test/main.sh
+++ b/example/test/main.sh
@@ -408,12 +408,12 @@ test_join_token_before_cluster_formed() {
 
 test_membership_consistency() {
   echo "Testing membership consistency checks"
-  
+
   new_systems 4 --heartbeat 2s
-  
+
   # Bootstrap first member (daemon already running from new_systems)
   microctl --state-dir "${test_dir}/c1" init "c1" 127.0.0.1:9001 --bootstrap
-  
+
   # Join second member (daemon already running)
   token_c2=$(microctl --state-dir "${test_dir}/c1" tokens add "c2")
   microctl --state-dir "${test_dir}/c2" init "c2" 127.0.0.1:9002 --token "${token_c2}"
@@ -421,13 +421,13 @@ test_membership_consistency() {
   # Start third member and join cluster
   token_c3=$(microctl --state-dir "${test_dir}/c1" tokens add "c3")
   microctl --state-dir "${test_dir}/c3" init "c3" 127.0.0.1:9003 --token "${token_c3}"
-  
+
   # Fetch join token for c4
   token_c4=$(microctl --state-dir "${test_dir}/c1" tokens add "c4")
 
   # Wait for cluster to stabilize
   echo "  -> Waiting for cluster members to exit PENDING state"
-  
+
   # Wait for all members to be promoted from PENDING
   retry_count=0
   max_retries=10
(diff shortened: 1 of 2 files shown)


internal/rest/resources/cluster.go
```diff
@@ -636,30 +636,22 @@ func clusterMemberDelete(s types.State, r *http.Request) types.Response {
...
-				return fmt.Errorf("ResponseWriter is not type http.Flusher")
-			}
+			logger.Info("Releasing cluster self removal lock", slog.String("member", name))
+			clusterDisableMu.Unlock()
+		}()
 
-			f.Flush()
-			return nil
-		})
+		return types.EmptySyncResponse
```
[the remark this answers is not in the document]
roosterfish: Ok, reverted.

example/test/main.sh
```diff
@@ -672,6 +672,28 @@ test_extended_endpoints() {
...
+  token_c4=$(microctl --state-dir "${test_dir}/c1" tokens add "c4")
+
+  # Join the cluster members.
+  microctl --state-dir "${test_dir}/c2" init "c2" 127.0.0.1:9002 --token "${token_c2}"
+  microctl --state-dir "${test_dir}/c3" init "c3" 127.0.0.1:9003 --token "${token_c3}"
+  microctl --state-dir "${test_dir}/c4" init "c4" 127.0.0.1:9004 --token "${token_c4}"
+
+  # Self delete the member c1.
+  microctl --state-dir "${test_dir}/c1" cluster remove c1
+
```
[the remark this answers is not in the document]
roosterfish: Added additional checks.

example/test/main.sh
```diff
@@ -672,6 +672,44 @@ test_extended_endpoints() {
...
+  echo "  -> Testing c1 got reset"
+  while [[ "$(microctl --state-dir "${test_dir}/c1" cluster list 2>&1)" != "Error: Database is not yet initialized" ]]; do
+    sleep 2
+  done
+
+  echo "  -> Testing c1 isn't anymore present on the remaining cluster"
+  [ "$(microctl --state-dir "${test_dir}/c2" cluster list -f csv | wc -l)" = "3" ]
+  [ "$(microctl --state-dir "${test_dir}/c2" cluster list -f json | yq '.[] | select(.name == "m1")')" = "" ]
```
tugbataluy: I'm not sure but should this be "c1" instead of "m1"? The members are named c1-c4 based on new_systems.

roosterfish: Thanks, good catch! I copied it from my terminal...

roosterfish: Fixed.

example/test/main.sh
```diff
@@ -672,6 +672,44 @@ test_extended_endpoints() {
...
+
+  echo "  -> Testing self deletion of member c1"
+  microctl --state-dir "${test_dir}/c1" cluster remove c1
+
+  echo "  -> Testing c1 got reset"
+  while [[ "$(microctl --state-dir "${test_dir}/c1" cluster list 2>&1)" != "Error: Database is not yet initialized" ]]; do
+    sleep 2
+  done
+
+  echo "  -> Testing c1 isn't anymore present on the remaining cluster"
```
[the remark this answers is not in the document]
roosterfish: fixed

internal/rest/resources/cluster.go
```diff
@@ -365,6 +365,13 @@ func resetClusterMember(ctx context.Context, s types.State, force bool) (reExec
 	reExec = func() {
 		<-ctx.Done() // Wait until request has finished.
 
+		// Wait until we can acquire the lock. This way if another request is holding the lock we won't
+		// replace/stop the LXD daemon until that request has finished.
```
tugbataluy: It's just cosmetic but I'm curious: Should this say "microcluster daemon" instead of "LXD daemon"? Or is LXD intentional here? Since microCluster is used by other project, maybe we can use just daemon.

roosterfish: You are right, it should say Microcluster. I was writing this comment too fast and haven't checked the previous contents which already said LXD.

roosterfish: Fixed.

tugbataluy (approved):
LGTM! 🚀
As far as I understand, the fix acquires the lock immediately after the request completes but before any shutdown logic runs. This ensures the shutdownCtx cancellation doesn't race with the client request. Thanks for the test coverage as well.

HomayoonAlimohammadi (approved):
Thanks a lot for the quick and accurate fix @roosterfish! I've tested on my end and the issue is resolved. LGTM!

PR #635, opened 2026-02-26
Add failure domain live update support and update daemon configuration APIs (failure domains part 2/2)

kadinsayani:
This pull request introduces support for failure domain live updates via a restart flag. A new CLI command is introduced for describing dqlite node metadata (useful for testing).

Depends on https://github.com/canonical/microcluster/pull/629.

### DIFF
--- a/example/cmd/microctl/describe.go
+++ b/example/cmd/microctl/describe.go
@@ -1,0 +1,128 @@
+package main
+
+import (
+	"context"
+	"crypto/tls"
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
+	"net"
+	"os"
+	"path/filepath"
+	"time"
+
+	dqliteClient "github.com/canonical/go-dqlite/v3/client"
+	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v3"
+
+	internalClient "github.com/canonical/microcluster/v4/internal/rest/client"
+)
+
+type cmdDescribe struct {
+	common *CmdControl
+
+	flagTimeout int
+}
+
+func (c *cmdDescribe) command() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "describe <address>",
+		Short: "Describe dqlite node metadata",
+		RunE:  c.run,
+	}
+
+	cmd.Flags().IntVarP(&c.flagTimeout, "timeout", "t", 5, "Number of seconds before timing out")
+	return cmd
+}
+
+func (c *cmdDescribe) run(cmd *cobra.Command, args []string) error {
+	if len(args) != 1 {
+		return cmd.Help()
+	}
+
+	ctx := cmd.Context()
+	var cancel context.CancelFunc = func() {}
+	if c.flagTimeout > 0 {
+		ctx, cancel = context.WithTimeout(ctx, time.Duration(c.flagTimeout)*time.Second)
+	}
+
+	defer cancel()
+
+	meta, err := describeDqliteNodeMetadata(ctx, c.common.FlagStateDir, args[0])
+	if err != nil {
+		return err
+	}
+
+	out, err := yaml.Marshal(struct {
+		FailureDomain uint64 `yaml:"failure-domain"`
+		Weight        uint64 `yaml:"weight"`
+	}{
+		FailureDomain: meta.FailureDomain,
+		Weight:        meta.Weight,
+	})
+	if err != nil {
+		return fmt.Errorf("Failed marshalling dqlite node metadata: %w", err)
+	}
+
+	fmt.Print(string(out))
+	return nil
+}
+
+// describeDqliteNodeMetadata connects to the given dqlite address and returns node metadata.
+// This helper exists to support developer/test verification workflows and is not intended as a
+// stable API surface.
+func describeDqliteNodeMetadata(ctx context.Context, stateDir string, address string) (*dqliteClient.NodeMetadata, error) {
+	serverCert, err := tls.LoadX509KeyPair(filepath.Join(stateDir, "server.crt"), filepath.Join(stateDir, "server.key"))
+	if err != nil {
+		return nil, fmt.Errorf("Failed loading server keypair: %w", err)
+	}
+
+	clusterPEM, err := os.ReadFile(filepath.Join(stateDir, "cluster.crt"))
+	if err != nil {
+		return nil, fmt.Errorf("Failed reading cluster certificate: %w", err)
+	}
+
+	block, _ := pem.Decode(clusterPEM)
+	if block == nil {
+		return nil, fmt.Errorf("Failed decoding cluster certificate")
+	}
+
+	clusterX509, err := x509.ParseCertificate(block.Bytes)
+	if err != nil {
+		return nil, fmt.Errorf("Failed parsing cluster certificate: %w", err)
+	}
+
+	pool := x509.NewCertPool()
+	ok := pool.AppendCertsFromPEM(clusterPEM)
+	if !ok {
+		return nil, fmt.Errorf("Failed appending cluster certificates to pool")
+	}
+
+	dial := func(ctx context.Context, address string) (net.Conn, error) {
+		config := &tls.Config{
+			Certificates: []tls.Certificate{serverCert},
+			RootCAs:      pool,
+			MinVersion:   tls.VersionTLS12,
+		}
+
+		if len(clusterX509.DNSNames) > 0 {
+			config.ServerName = clusterX509.DNSNames[0]
+		}
+
+		return internalClient.DialDqlite(ctx, address, config)
+	}
+
+	client, err := dqliteClient.New(ctx, address, dqliteClient.WithDialFunc(dial))
+	if err != nil {
+		return nil, fmt.Errorf("Failed connecting to dqlite: %w", err)
+	}
+
+	defer client.Close()
+
+	meta, err := client.Describe(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("Failed describing dqlite node metadata: %w", err)
+	}
+
+	return meta, nil
+}
--- a/example/test/main.sh
+++ b/example/test/main.sh
@@ -11,7 +11,21 @@ fi
 
 test_dir="$(realpath -e "$(dirname -- "${BASH_SOURCE[0]}")")/system"
 
-trap shutdown_systems EXIT HUP INT TERM
+# Must be set before cleanup().
+TEST_CURRENT="setup"
+TEST_RESULT="failure"
+
+declare -A test_results
+
+cleanup() {
+  shutdown_systems
+
+  echo ""
+  echo ""
+  echo "==> Test result: ${TEST_RESULT}"
+}
+
+trap cleanup EXIT HUP INT TERM
 
 new_systems() {
   if [ -d "${test_dir}" ]; then
@@ -73,6 +87,25 @@ shutdown_systems() {
       kill -9 "${job_pid}" 2>/dev/null || true
     fi
   done
+}
+
+run_test() {
+  local test_name="${1}"
+
+  TEST_CURRENT="${test_name}"
+  echo "==> TEST BEGIN: ${TEST_CURRENT}"
+
+  if "test_${test_name}"; then
+    test_results["${test_name}"]="PASS"
+  else
+    test_results["${test_name}"]="FAIL"
+  fi
+  echo "==> TEST DONE: ${TEST_CURRENT}"
+
+  if [ "${test_results[${test_name}]}" != "PASS" ]; then
+    TEST_RESULT="failure"
+    return 1
+  fi
 }
 
 test_misc() {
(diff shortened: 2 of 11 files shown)


roosterfish:
Please change the target branch to v4 as this is the new default after we have released v3.

internal/daemon/daemon.go
```diff
@@ -73,6 +73,13 @@ type Args struct {
 	// If nil, default signals (SIGPWR, SIGTERM, SIGINT, SIGQUIT) will be used.
 	// If set to an empty slice, no signal handling will be performed.
 	ShutdownSignals []os.Signal
+
+	// FailureDomain is the dqlite failure-domain for this node. It is used by dqlite
+	// to place voting replicas across fault boundaries. If the node has already been
+	// bootstrapped and daemon.yaml contains a failure-domain value, that value takes
+	// precedence over this field. Set this to avoid a dqlite restart when the failure
+	// domain is known at daemon startup time.
+	FailureDomain uint64
```
roosterfish: I wonder if the name should reflect that it's only used initially. Something like `InitFailureDomain` as we already have a `PreInitListenAddress`.

microcluster/types/config.go
```diff
@@ -6,6 +6,8 @@ type DaemonConfig struct {
 	Address       AddrPort                `json:"address" yaml:"address"`
 	Servers       map[string]ServerConfig `json:"servers" yaml:"servers"`
 	FailureDomain uint64                  `json:"failure-domain" yaml:"failure-domain"`
+	// Restart is a request-only field; it is not persisted to daemon.yaml.
```
roosterfish: Should we rather have another arg on the `(m *MicroCluster) UpdateDaemonConfig` like `restart bool`? This way we don't have to add such request-only fields and keep a cleaner structure which isn't filled with control fields.

We can still change the signature of the func as it's in the edge branch and not yet backported to `v2`.

internal/db/dqlite.go
```diff
@@ -66,9 +67,28 @@ const (
...
+	offline := db.status == types.DatabaseOffline
+	db.statusLock.RUnlock()
+
+	if offline {
+		conn.Close()
+		return
+	}
+
+	select {
+	case db.acceptCh <- conn:
```
roosterfish, line 78: Can we keep this blocking until dqlite is restarted? 

Let's say we shutdown using `db.dqlite.Close()`, from this point onward I assume `db.acceptCh` wouldn't anymore have a consumer so the `db.acceptCh <- conn` blocks as it's an unbuffered chan.

We can add a timeout saying if we weren't able to hand off the `conn` in 10s we error out. But this would allow keeping clients "on hold" whilst we restart dqlite on this member and they won't even witness the restart.

Or do we have to create a fresh `acceptCh`?

internal/rest/resources/daemon.go
```diff
@@ -103,6 +103,10 @@ func daemonConfigPut(s types.State, r *http.Request) types.Response {
 		return types.SmartError(err)
 	}
 
+	if req.Restart {
```
roosterfish: Earlier I mentioned to instead have an arg `restart bool` on the app func. Internally we could map this to another query param we can then evaluate here instead of a field on the struct?

example/cmd/microctl/describe.go
```diff
@@ -0,0 +1,158 @@
...
+		return nil, fmt.Errorf("Failed parsing cluster certificate: %w", err)
+	}
+
+	pool := x509.NewCertPool()
+	ok := pool.AppendCertsFromPEM(clusterPEM)
+	if !ok {
+		return nil, fmt.Errorf("Failed appending cluster certificates to pool")
+	}
+
+	dial := func(ctx context.Context, address string) (net.Conn, error) {
```
roosterfish, line 101: Nice one, definitely worth having for testing purposes!

roosterfish: I wonder if we can reuse any of the dial funcs we already have in the code base?

example/cmd/microctl/describe.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	defer cancel()
+
+	meta, err := describeDqliteNodeMetadata(ctx, c.common.FlagStateDir, args[0])
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("failure-domain=%d weight=%d\n", meta.FailureDomain, meta.Weight)
```
roosterfish: Out of interest, what does `weight` mean in this context and why is it added here?

example/test/main.sh
```diff
@@ -879,6 +879,75 @@ test_daemon_config_api() {
     return 1
   }
 
+  # PATCH restart tests.
+  # Set failure-domain to a known value and verify it takes effect immediately.
+
+  # First verify current failure-domain reported by dqlite is 0.
+  fd_before="$(microctl --state-dir "${test_dir}/c1" describe 127.0.0.1:9001 | grep -oP 'failure-domain=\K[0-9]+')"
```
roosterfish: Just a nit, but what about returning yaml format from the `describe` command? Would allow easy use of `yq`.

example/test/main.sh
```diff
@@ -879,6 +879,75 @@ test_daemon_config_api() {
...
+  # PATCH with restart:true — failure-domain should apply immediately.
+  patch_resp="$(daemon_config_patch '{"failure-domain":42,"restart":true}')"
+  [[ "$(response_code "${patch_resp}")" == "200" ]] || {
+    echo "ERROR: PATCH with restart failed"
+    echo "${patch_resp}"
+    return 1
+  }
+
+  # The db should come back online and the failure-domain should be updated.
+  microctl --state-dir "${test_dir}/c1" waitready
```
roosterfish: I wonder when you set `restart=true`, should we wait until the DB is back online and only then return the respective PUT/PATCH?
Otherwise you always need to remember to run `waitready` until the DB is online again.

roosterfish (changes requested):
Thanks, I very much enjoy the new `describe` command and how you have added it to the tests.

In general I wonder if we can be less destructive with the restart when it comes to new connections. Please see my comments.

microcluster/types/config.go
kadinsayani: Yeah, that sounds good!

internal/rest/resources/daemon.go
kadinsayani: Good idea, thanks.

example/cmd/microctl/describe.go
kadinsayani: We can't reuse one directly because there isn't already a helper with the right abstraction boundary.

`dqliteNetworkDial` is tied to `DqliteDB` state for cert lookup and logging, so the CLI can't use it cleanly. However, we can extract a shared helper that dials a remote member's internal database endpoint and upgrades the connection. We can use this in `dqliteNetworkDial` and in `describe.go/describeDqliteNodeMetadata`.

example/cmd/microctl/describe.go
kadinsayani: `go-dqlite.NodeMetadata.Weight` is used to order members in a failure domain. It's used during the rebalancing logic to sort candidates. While less important for our uses, I've included it for verbosity.

example/test/main.sh
kadinsayani: Great idea, thanks!

example/cmd/microctl/describe.go
roosterfish: I would prefer this. Because then we have the bits that set the right headers etc. only in one place.

internal/db/dqlite.go
kadinsayani: We can block and add a 10s timeout. Another approach could be making `acceptCh` buffered. During restart, incoming connections queue silently in the buffer instead of blocking. When dqlite is started, we drain the buffer. Wdyt?

example/test/main.sh
kadinsayani: The restart is synchronous, the response is only sent after `Open()` succeeds and sets `DatabaseReady`, so callers don't actually need a separate waitready.

internal/db/dqlite.go
roosterfish: The issue with the buffered channel is just which buffer length should we pick? If we let's say use 100, the 101th client will be put on hold. And then it's the same as using an unbuffered channel.
Do you see any other benefit of using a buffered channel? The clients would still have to wait until the server picks up the conn from the queue I believe.

Something else that just came to my mind is that we should transfer the dqlite leadership to another member (if > 1 cluster members) before we perform the restart if the to be restarted member is currently the leader.

example/test/main.sh
roosterfish: Great, so the extra waitready here is not required then?

kadinsayani: Yeah I've dropped it :)

internal/db/dqlite.go
kadinsayani: There are tradeoffs to both approaches. You raise a good point on buffer size. 

With the timeout approach, clients are blocked and if the restart fails, connections are dropped.

> we should transfer the dqlite leadership to another member (if > 1 cluster members) before we perform the restart if the to be restarted member is currently the leader.

Good point. I'm currently working on implementing this in LXD as well. I'll add logic for this here.

kadinsayani: I've gone with the timeout approach and I've added leadership transfer pre-restart.

internal/db/dqlite.go
```diff
@@ -553,3 +568,108 @@ func (db *DqliteDB) Stop() error {
...
+	}
+
+	db.statusLock.Lock()
+	db.status = types.DatabaseOffline
+	db.statusLock.Unlock()
+
+	// go-dqlite keeps an accept-loop goroutine alive after Close() when using external connections.
+	// Rotate acceptCh so the next dqlite instance does not share the old channel with stale readers.
+	db.acceptMu.Lock()
+	db.acceptCh = make(chan net.Conn)
```
roosterfish, line 531: Do we have to replace the chan? Or can we just reuse the existing one?

internal/rest/client/daemon.go
```diff
@@ -50,3 +51,14 @@ func UpdateServers(ctx context.Context, c types.Client, config map[string]types.
 	endpoint := api.NewURL().Path("daemon", "servers")
 	return c.Query(queryCtx, "PUT", types.PublicEndpoint, &endpoint.URL, config, nil)
 }
+
+func daemonConfigURL(restart bool) *url.URL {
+	endpoint := api.NewURL().Path("daemon", "config")
+	if restart {
+		query := endpoint.Query()
```
roosterfish: You can do `.WithQuery("restart", "true")`

internal/rest/client/dqlite.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	request.URL = &url.URL{
+		Scheme: "https",
+		Host:   addrPort.String(),
+		Path:   fmt.Sprintf("/%s/%s", types.InternalEndpoint, "database"),
+	}
+
+	request.Header.Set("Upgrade", "dqlite")
+	request.Header.Set("X-Dqlite-Version", "1")
```
roosterfish, line 46: Good call just using "1" :)

roosterfish (changes requested):
Looking great, only one more smaller questions.

internal/db/dqlite.go
kadinsayani: Yes. `Close()` can leave the old go-dqlite accept-loop goroutine alive when using external connections, so if we reuse the same channel the stale reader can race with the new dqlite instance and consume connections meant for it. Rotating acceptCh gives the restarted instance a fresh channel with no leaked readers attached.

internal/rest/client/daemon.go
kadinsayani: Good point, I'll fix that.

internal/db/dqlite.go
roosterfish: Is this a bug in `go-dqlite` that you see happen only in some occations? Or is this happening every time? I wonder what happens if we close the channel proactively, does this terminate the routine?

Can we account for it and try to drain the channel `conn`s after locking `acceptMu` (so that no new client can send its `conn`)? 
And then wait with the restart until the chan is empty? 

To deal with the wait, maybe we use a `sync.WaitGroup` and `wg.Add(1)` everytime there is a new client `conn`. Because then after running `acceptMu.Lock` we could `wg.Wait()` until all client's `conn`s are handled and then we don't ever drop any client due to the restart.
Also at this point we can close the old `acceptCh` which would hopefully cause the go-dqlite routine to exit too.

roosterfish: If this is not feasible I am happy to get the PR in as is. But then we should document that ongoing client connections to this member might be lost during restart. Which is acceptable if you ask me.

kadinsayani: `go-dqlite.WithExternalConn` stores the channel, and `app.New()` starts a goroutine that reads `for remote := range acceptCh` and proxies those conns into the local unix socket. `App.Close()` does not close the channel or wait for that goroutine to exit, so an old reader can outlive `Close()` if we reuse the same channel.

I've updated the restart path to swap in a fresh channel under `acceptMu` and close the old one. Since `Accept()` is the only sender and it takes `acceptMu` for every send, once the write lock is held there are no remaining senders on the old channel from our side. Closing the old channel lets the stale `for remote := range acceptCh` loop exit cleanly.

roosterfish: > // Accept() holds acceptMu for every send, so once the write lock is held there
  // can be no remaining senders on the old channel.

Thanks for this additional context. 
And we can be sure there are no remaining senders because it's an unbuffered channel and the lock is only returned once the write to `acceptCh` was successful, meaning the go-dqlite routine has read the `conn`.

roosterfish (approved):
Neat. Thanks @kadinsayani

PR #643, opened 2026-03-02
Go: Bump to v4

roosterfish:
Changes all Microcluster package paths to `v4` as `v4` is the new edge default branch.


tugbataluy (approved):
LGTM! Thanks for this, great to see v4 as the new default now that v3 has just been released. You've put in a ton of work on this transition, 🚀 Approved!

PR #644, opened 2026-03-02
make: Pin recommended dqlite version (v3)

roosterfish:
Integrating Microcluster requires the dqlite deps to be built.

Pin the recommended version of dqlite for the 'make deps' call for clarity. Also this ensures the tests on v3 are always executed using deps built from v1.18.x


roosterfish:
Thanks for pointing this out @troyanov.

troyanov:
Thanks for addressing it that quick @roosterfish and kudos for all the great work on `v3`

tugbataluy (approved):
LGTM, thanks ❇️

PR #650, opened 2026-03-05
microcluster: Add pending check for cluster member role

HomayoonAlimohammadi:
This pull request introduces a minor update to the `microcluster/app.go` file, primarily adding a utility function to help identify pending cluster members and updating imports to support this change.

Cluster member state handling:

* Added a new function `IsClusterMemberPending` that checks if a `ClusterMember` has a role of `PENDING`, since `Pending` is an internal const and cannot be accessed.

### DIFF
--- a/microcluster/app.go
+++ b/microcluster/app.go
@@ -16,6 +16,7 @@ import (
 	"github.com/canonical/lxd/shared/api"
 	"golang.org/x/sys/unix"
 
+	"github.com/canonical/microcluster/v3/internal/cluster"
 	"github.com/canonical/microcluster/v3/internal/daemon"
 	"github.com/canonical/microcluster/v3/internal/log"
 	"github.com/canonical/microcluster/v3/internal/recover"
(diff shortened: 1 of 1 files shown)


microcluster/app.go
```diff
@@ -239,6 +240,11 @@ func (m *MicroCluster) GetClusterMembers(ctx context.Context) ([]types.ClusterMe
 	return internalClient.GetClusterMembers(ctx, c)
 }
 
+// IsClusterMemberPending returns true if the ClusterMember's role is PENDING.
+func IsClusterMemberPending(cm types.ClusterMember) bool {
```
roosterfish, line 244: I assume you are suggesting this func so you can access `github.com/canonical/microcluster/v3/internal/cluster`? 

Let's rather add the `Pending` const into `microcluster/types/cluster.go` where we already define various other `MemberStatus` constants. What do you think?

I believe we can even remove the type `Role` of `CoreClusterMember` and use the `MemberStatus` from the public `microcluster/types`.

roosterfish (changes requested):

roosterfish:
Thanks for the suggestion, I think we can move the `Pending` status to the others in the public package. Please see my comment.

roosterfish:
@HomayoonAlimohammadi any updates on this one or should we close the PR?

roosterfish:
Please reopen if you still want to get this changed.

PR #731, opened 2026-05-26
Makefile: Update golangci-lint install URL

markylaing:


markylaing:
@simondeziel @roosterfish do we also need the equivalent of #734 for the `v2` branch?

roosterfish:
> @simondeziel @roosterfish do we also need the equivalent of #734 for the `v2` branch?

Yes I think so. Do you want to cherry-pick the commits into this PR?

simondeziel (approved):
You beat me to it, thanks!

PR #751, opened 2026-06-11
internal/rest/resources/cluster: Return 404 when member not found in truststore

claudiubelu:
`clusterMemberDelete` is returning a `SmartError` when the given member is absent from the truststore. This makes it difficult for callers to distinguish "not found" from a real internal error.

Switch to `NotFound` error, so callers can use `IsNotFoundError` to treat an already-absent member as a success.


roosterfish (approved):

roosterfish:
Ty!

PR #752, opened 2026-06-11
internal/rest/resources/cluster: Return 404 when member is not found

claudiubelu:
`clusterMemberDelete` is returning a `SmartError` when the given member is absent from the truststore. This makes it difficult for callers to distinguish "not found" from a real internal error.

Switch to `NotFound` error, so callers can use `IsNotFoundError` to treat an already-absent member as a success.

### DIFF
--- a/internal/rest/resources/cluster.go
+++ b/internal/rest/resources/cluster.go
@@ -418,7 +418,7 @@ func clusterMemberDelete(s types.State, r *http.Request) types.Response {
 	addr := r.URL.Query().Get("address")
 	name, err := url.PathUnescape(mux.Vars(r)["name"])
 	if err != nil {
-		return types.SmartError(err)
+		return types.BadRequest(err)
 	}
 
 	ctx := r.Context()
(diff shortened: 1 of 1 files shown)


internal/rest/resources/cluster.go
```diff
@@ -439,7 +439,7 @@ func clusterMemberDelete(s types.State, r *http.Request) types.Response {
 		addr = remote.Address.String()
 	} else if !remotePresent && addr == "" {
 		// If the remote is not present in the truststore and no address is provided, we cannot proceed.
-		return types.SmartError(fmt.Errorf("Cluster member %q not found in truststore; please provide a node address", name))
+		return types.NotFound(fmt.Errorf("Cluster member %q not found in truststore; please provide a node address", name))
 	} else if remotePresent && addr != "" && remote.Address.String() != addr {
 		// Reject if provided address doesn't match the truststore address for this remote name.
 		return types.SmartError(fmt.Errorf("Provided address %q does not match the address %q of the remote with name %q", addr, remote.Address.String(), name))
```
roosterfish: Whilst you are at it, can we please also change this to `types.BadRequest`? 
There are also a few more below which depend on user input.

roosterfish (changes requested):
Thanks, good catch. 

I see you have also opened up two more PRs for `v3` and `v2`. Please make sure those commits are picked using `git cherry-pick -x` to keep the history from where the changes are coming from.

internal/rest/resources/cluster.go
claudiubelu: Done

roosterfish (approved):
Thanks!

claudiubelu:
> Thanks, good catch.
> 
> I see you have also opened up two more PRs for `v3` and `v2`. Please make sure those commits are picked using `git cherry-pick -x` to keep the history from where the changes are coming from.

I've updated those branches as well, applying the same fixes to them via `cherry-pick -x`.

PR #761, opened 2026-06-17
internal/rest/resources/control: retry remaining join addresses on client connect err

claudiubelu:
When joining a cluster, `joinWithToken` iterates over all addresses in the token to find a reachable cluster member.

However, if `state.Connect().Member` fails for an address, the function returns immediately instead of trying the next address. This is inconsistent with how other failures in the same loop are handled (`GetRemoteCertificate` and `AddClusterMember` both continue to the next address on error).

The function now iterates over all the join addresses in this scenario.

(cherry picked from commit 39f47c058bed6f1cb65db3334a052ecc52139786)


roosterfish:
Marked as draft until https://github.com/canonical/microcluster/pull/760#discussion_r3435537832 gets merged.

PR #774, opened 2026-06-29
internal/db: Remove node from dqlite cluster on failed join

claudiubelu:
`dqlite.Close()` tears down Raft connections but does not remove the node from the dqlite cluster membership. When `db.Join()` fails after the node has already been added to the cluster (via cli.Add in app.run), the reverter only called `Close()`, leaving a ghost member in dqlite.

This ghost node causes `CheckMembershipConsistency` to fail (the node exists in dqlite but not in `core_cluster_members`), which blocks token generation and all future cluster joins.

Fix the Join reverter to explicitly ask the `dqlite` leader to remove us before closing. Also correct a misleading comment in the Bootstrap reverter.

### DIFF
--- a/internal/db/dqlite.go
+++ b/internal/db/dqlite.go
@@ -174,7 +174,9 @@ func (db *DqliteDB) Bootstrap(extensions types.Extensions, addr *url.URL, cluste
 	}
 
 	// Only close the dqlite app if the entire Bootstrap process fails or exits.
-	// dqlite.Close() immediately tears down all Raft connections and removes this node from the dqlite cluster.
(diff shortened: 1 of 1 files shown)


roosterfish (changes requested):
Hey, can you please share some reproducer steps how you could end up in this state? 
I see we have an exception to continue starting up the database in case we hit a `update.ErrGracefulAbort` error. But I would be very interested to see what else could go wrong.

Also could you please add the reproducer as an additional check to `test_membership_consistency` ?

roosterfish:
Marking as draft until you get a chance to get back to it @claudiubelu.

PR #777, opened 2026-06-29
internal/rest/resources/control: Use independent context for join faiure cleanup

claudiubelu:
The `controlPost` reverter used `r.Context()` for all cleanup operations (`PreRemove` hook, certificate fetch, client creation).

When a join fails due to context deadline exceeded, `r.Context()` may already be expired, causing the entire cleanup to silently fail and leaving a ghost node in the dqlite cluster. In particular, `shared.GetRemoteCertificate` was silently returning due an error returned (due to `r.Context()` being expired), which means that `internalClient.DeleteClusterMember` was never called.

Use an independent context with a 2-minute timeout for cleanup operations so they can complete regardless of the original request context state.

Also add error logging to `GetRemoteCertificate` and client creation failures which previously returned silently.

### DIFF
--- a/internal/rest/resources/control.go
+++ b/internal/rest/resources/control.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path/filepath"
 	"slices"
+	"time"
 
 	"github.com/canonical/lxd/shared"
 	"github.com/canonical/lxd/shared/api"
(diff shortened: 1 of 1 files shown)


internal/rest/resources/control.go
```diff
@@ -91,8 +92,14 @@ func controlPost(state types.State, r *http.Request) types.Response {
 			return
 		}
 
+		// Use an independent context for cleanup so that it is not tied
+		// to the request context, which may already be expired (e.g. the
+		// join timed out).
+		cleanupCtx, cleanupCancel := context.WithTimeout(context.Background(), 2*time.Minute)
```
roosterfish, line 98: Instead of deriving a new context, should we rather use the daemon's `shutdownCtx` here to be independent of the join request but still account for the health of the daemon when running the hooks?

roosterfish (changes requested):
Thanks, could you please also add a test that triggers this?

roosterfish:
Marking as draft until you get a chance to get back to it @claudiubelu.

Issue #784, opened 2026-07-09
Prejoin hook does not fail cleanly.


crypticC0der:
If you error within the PreJoin hook in microcluster and then try to rejoin you get this:
```
Membership consistency check failed after timeout: Microcluster node membership is inconsistent across core_cluster_members ([10.33.87.13:6686 10.33.87.53:6686]), truststore ([10.33.87.13:6686 10.33.87.199:6686 10.33.87.53:6686]) and dqlite ([10.33.87.13:6686 10.33.87.53:6686])
```

This is affecting some new https://github.com/canonical/microovn functionality. It is not urgent but a fix would be nice.

PR #785, opened 2026-07-09
Ensure node is fully removed instead of partially.

crypticC0der:
Previously on removing the node failing, which happens if the node is not in the cluster, the function exits. Which means the trust store isn't cleaned. This stops an issue wherin a prejoin hook could fail and cause the trust store to remain in a broken state which is not good.

Fixes: #784

### DIFF
--- a/internal/rest/resources/cluster.go
+++ b/internal/rest/resources/cluster.go
@@ -709,8 +709,12 @@ func clusterMemberDelete(s types.State, r *http.Request) types.Response {
 	// Remove the node from dqlite, if it has a record there.
 	if index >= 0 {
 		err = leader.Remove(ctx, info[index].ID)
-		if err != nil {
+		if err != nil && !force {
 			return types.SmartError(err)
+		}
+
+		if err != nil {
+			logger.Error("Failed to remove cluster member from dqlite", slog.String("error", err.Error()), slog.Bool("force", force), slog.String("name", name))
 		}
 	}
 
--- a/example/cmd/microd/main.go
+++ b/example/cmd/microd/main.go
@@ -157,6 +157,10 @@ func (c *cmdDaemon) run(cmd *cobra.Command, args []string) error {
 			logger.Info("This is a hook that runs after the daemon is initialized and joins an existing cluster, before OnNewMember runs on all peers")
 			logger.Info("Here are the extra configuration keys that were passed into the init --join command", slog.Any("config", initConfig))
 
+			if os.Getenv("FAIL_PREJOIN") == "1" {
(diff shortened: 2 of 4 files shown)


internal/rest/resources/cluster.go
```diff
@@ -709,9 +709,13 @@ func clusterMemberDelete(s types.State, r *http.Request) types.Response {
 	// Remove the node from dqlite, if it has a record there.
 	if index >= 0 {
 		err = leader.Remove(ctx, info[index].ID)
-		if err != nil {
+		if err != nil && !force {
 			return types.SmartError(err)
 		}
+
+		if err != nil {
+			logger.Error("Failed to remove cluster member from dqlite, continuing with other cleanup", slog.String("error", err.Error()))
```
roosterfish: Can you please omit the part after the `,`? Just logging `Failed to remove cluster member from dqlite` should be enough.
Also can you please add `slog.Bool("force", force)` and maybe also a `slog.String("name", name)` to give a bit more context which member couldn't be deleted?

roosterfish (changes requested):
Thanks for this! One remark from my side.

Also does this actually fix https://github.com/canonical/microcluster/issues/784 or does it rather allow cleanly removing the member from the cluster after you hit the consistency error?

crypticC0der:
DeleteClusterMember is called with force on https://github.com/canonical/microcluster/blob/a2735b4a69e9134f1b3abc64c54c52b88a6abe9d/internal/rest/resources/control.go#L133 so this does fix it when it fails with prejoin, it also means deleting a cluster member will 100% fix the trust store if its somehow still broken

roosterfish:
> DeleteClusterMember is called with force on
> 
> https://github.com/canonical/microcluster/blob/a2735b4a69e9134f1b3abc64c54c52b88a6abe9d/internal/rest/resources/control.go#L133
> so this does fix it when it fails with prejoin, it also means deleting a cluster member will 100% fix the trust store if its somehow still broken

I see thanks. Please also append the package name as prefix to the commit message. For example see https://github.com/canonical/microcluster/commit/065316aa4fa0fe23e5606ec176ef9c934c383654.

Also would it be feasible to extend any of the tests in `example/test` to simulate and test this? We could maybe modify the `PreJoin` hook in there to have a different outcome based on something we can set external to Microcluster like the existence of a file or an env variable being set like `FAIL_PREJOIN=1`.

roosterfish:
@crypticC0der tests are failing.

roosterfish:
I'll move this to draft until the failing test/comments are addressed.

